PR #1, opened 2017-12-16
Advanced error codes

wscsprint3r:


ghislainfourny:
Hi Stefan,

It looks very good. I tried a few queries and got the right error messages.

However, a few queries generate a stacktrace instead of the nice error message, so some exceptions may still need to be caught up (maybe also with a global catch surrounding the main code). For example :

    let $a := 2  return for $i in 1 to 10 group by $a return $i

    {"foo":1}.("foo", "bar")

The latter is a null pointer exception.

Thanks!

ghislainfourny:
Another thing I found is that there seems to be an offset of 1 in the line number:

    jiqs$  { [ 1 ] : 1 }
    >>> 
    >>> 
    [ERROR] Error [err: XPDY0130]LINE:2:COLUMN:0:Object must have string keys!

wscsprint3r:
1) {"foo":1}.("foo", "bar") - Null pointer exception - FIXED

2) let $a := 2  return for $i in 1 to 10 group by $a return $i
Seems like most exceptions that are thrown within closures have their stack traces printed by the driver. The only way to avoid this is by setting the logger level to FATAL (not sure about this) and by using a little hack in the shell environment:
 output(ERROR_MESSAGE_PROMPT + ex.getMessage().split("\n")[0]);

This results in the following output:
[ERROR] Job aborted due to stage failure: Task 0 in stage 1.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1.0 (TID 1, localhost, executor driver): org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 2.0 failed 1 times, most recent failure: Lost task 0.0 in stage 2.0 (TID 2, localhost, executor driver): sparksoniq.exceptions.InvalidGroupVariableException: Error [err: XQST0094]LINE:2:COLUMN:47:Variable $a cannot be used in group clause

3) Line number is one line too great. This only happens in the shell environment, running queries from files outputs the correct line number. I will investigate further...

ghislainfourny:
> {"foo":1}.("foo", "bar") - Null pointer exception - FIXED
> Line number is one line too great. FIXED

Awesome, Stefan, thanks!

ghislainfourny:
> Seems like most exceptions that are thrown within closures have their stack traces printed by the driver. The only way to avoid this is by setting the logger level to FATAL (not sure about this) and by using a little hack in the shell environment

Thanks a lot, I tested it successfully. I think it's good for an alpha version, so we can go for this solution.

In the future (beta, RC, release), the clean approach will be to catch exceptions in the closure, and if there is an exception or error, have the Spark task return some special internal sequence of items or tuple stream that carries (encapsulates) the exception or error. Then you can test for that special value received on the caller side, and if it is an exception you can "unwrap" it and print it nicely, as if it had been executed locally (which is the "ideal" way the engine should feel like to the user).

Of course, fatal errors will still be fatal errors. But everything non-fatal we should be able to catch, encapsulate and print nicely. It's the kind of design that makes a product usable and attractive to users.

PR #27, opened 2018-01-31
Issue13 a few useful functions to implement first

wscsprint3r:


sparksoniq-start.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
```
wscsprint3r: This is only a basic version, I will make it much better.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -103,9 +114,24 @@ protected RuntimeIterator(List<RuntimeIterator> children, IteratorMetadata metad
...
+    protected List<Item> getItemsFromIteratorWithCurrentContext(RuntimeIterator iterator) {
+        List<Item> result = new ArrayList<>();
+        iterator.open(_currentDynamicContext);
+        while (iterator.hasNext())
+            result.add(iterator.next());
+        iterator.close();
+        return result;
+    }
+
+    protected <T extends Item> T getSingleItemOfTypeFromIterator(RuntimeIterator iterator, Class<T> type) {
```
wscsprint3r: Curios what you think about this. It's not the best generic solution but this is basic functionality that is used in A LOT of iterators that can be simplified

wscsprint3r:
I introduced auto-formatting in IntelliJ for consistent formating throughout the project. I am using IntelliJ defaults for now, but we can use any formatting you want using the Eclipse Formatter plug-in for IntelliJ and creating a custom formating XML (for Eclipse the file can be imported directly).

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -17,37 +17,53 @@
  * Author: Stefan Irimescu
  *
  */
- package sparksoniq.jsoniq.runtime.iterator;
+package sparksoniq.jsoniq.runtime.iterator;
```
ghislainfourny: As a side note for future pull requests: it is best to put cosmetic/lint changes in separate pull requests on their own, it makes reviewing much easier as pull requests get smaller and have a distinct focus.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
ghislainfourny: Yes, refactoring for functionality used at many places is a great idea.

Regarding functions: in general, we need functionality that checks function signatures in a generic way, not only for single items of a given type. On a high level, there should be somewhere a catalogue of available functions, with their names, arity and signatures (as many sequence types as the arity), and a pointer to the corresponding runtime iterator class.

When a function is called, it is looked up in this catalogue (by name and arity), the corresponding runtime iterator is picked and added to the expression tree, and parameter types are checked at runtime against the signature. If the function isn't found, an error can be through already statically, which makes the code faster at runtime as there is no need to check for the number of parameters any more.

This signature check can be done with fully factored out code, like children iterators that wrap the "actual" children iterators and check the sequence types and may do some numeric casts (s.t. decimal to double, etc).

This kind of architecture can be reused later as well, when we support user-defined functions.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/StringJoinFunction.java
```diff
@@ -0,0 +1,40 @@
...
+import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
+
+import java.util.List;
+
+public class StringJoinFunction extends LocalFunctionCallIterator {
+    public StringJoinFunction(List<RuntimeIterator> arguments, IteratorMetadata iteratorMetadata) {
+        super(arguments, iteratorMetadata);
+        if (arguments.size() != 2)
+            throw new SparksoniqRuntimeException("Incorrect number of arguments for string-join function; " +
+                    "Exactly 2 arguments are required.");
```
ghislainfourny: string-join() also exists with only one parameter. Then the second parameter is implicitly the empty string. That's why a function catalogue by name and arity will help here -- however of course, you can still reuse the same iterator for both string-join#1 and string-join#2.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/SubstringFunctionIterator.java
```diff
@@ -0,0 +1,52 @@
...
+import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
+import sparksoniq.jsoniq.runtime.iterator.functions.base.LocalFunctionCallIterator;
+import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
+
+import java.util.List;
+
+public class SubstringFunctionIterator extends LocalFunctionCallIterator {
+    public SubstringFunctionIterator(List<RuntimeIterator> arguments, IteratorMetadata iteratorMetadata) {
+        super(arguments, iteratorMetadata);
+        if (arguments.size() < 2 || arguments.size() > 3)
```
ghislainfourny: This is also the kind of checks that will no longer be needed at runtime with a function catalogue.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -63,6 +63,17 @@
      * function that returns the length of an array
```
ghislainfourny: This is an excellent candidate location for the function  catalogue. We have only names here, but you simply need to convert these static constants to a (possibly static as well) map from name+arity to signature + pointers to runtime iterator classes.

ghislainfourny:
Is there a way to have the CI put some red or green flag here automatically, or by the push of a button? I guess Travis should integrate with GitHub in this respect.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -17,52 +17,120 @@
...
+
+        buildInFunctions.put(new SparksoniqFunctionSignature(STRINGJOIN, 1), StringJoinFunction.class);
+        buildInFunctions.put(new SparksoniqFunctionSignature(STRINGJOIN, 2), StringJoinFunction.class);
+
+        buildInFunctions.put(new SparksoniqFunctionSignature(SIZE, 1), ArraySizeFunctionIterator.class);
+        buildInFunctions.put(new SparksoniqFunctionSignature(KEYS, 1), ObjectKeysFunctionIterator.class);
+        buildInFunctions.put(new SparksoniqFunctionSignature(VALUES, 1), ObjectValuesFunctionIterator.class);
+    }
+
+    public static Class<? extends RuntimeIterator> getFunctionIteratorClass(FunctionCall expression, List<RuntimeIterator> arguments) {
```
wscsprint3r: @ghislainfourny , this is the method used in the visitor now.

src/main/java/sparksoniq/semantics/visitor/RuntimeIteratorVisitor.java
```diff
@@ -344,125 +347,101 @@ public RuntimeIterator visitOrExpr(OrExpression expression, RuntimeIterator argu
...
-                        ObjectFunctionIterator.ObjectFunctionOperators.KEYS, iteratorMetadata);
-            case Functions.VALUES:
-                return new ObjectFunctionIterator(arguments,
-                        ObjectFunctionIterator.ObjectFunctionOperators.VALUES, iteratorMetadata);
-            case Functions.SIZE:
-                return new ArrayFunctionIterator(arguments,
-                        ArrayFunctionIterator.ArrayFunctionOperators.SIZE, iteratorMetadata);
-
+        try {
+            Class<? extends RuntimeIterator> functionClass = Functions.getFunctionIteratorClass(expression, arguments);
```
wscsprint3r: @ghislainfourny instance creation refactoring

wscsprint3r:
@ghislainfourny the travis build was integrated automatically in the checks section of the PR

src/main/java/sparksoniq/jsoniq/runtime/iterator/control/SwitchRuntimeIterator.java
```diff
@@ -44,37 +48,28 @@ public boolean hasNext() {
...
-        if(test.hasNext())
-            throw new NonAtomicKeyException("Switch test must be atomic", getMetadata().getExpressionMetadata());
-        test.close();
-        for(RuntimeIterator caseKey : cases.keySet()){
-            caseKey.open(_currentDynamicContext);
-            Item caseValue = caseKey.next();
-            if(caseKey.hasNext())
-                throw new NonAtomicKeyException("Switch case test must be atomic", getMetadata().getExpressionMetadata());
+        Item testValue = getSingleItemOfTypeFromIterator(test, Item.class,
+                new NonAtomicKeyException("Switch test must be atomic", getMetadata().getExpressionMetadata()));
```
ghislainfourny: Passing only the string rather than an exception object would probably make the code faster, because other the exception object is created every time, even when there is no exception to throw.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
ghislainfourny: This already looks much better. I think I may have confused you on what "signature" means. I would call the class SparksoniqFunctionNameAndArity. In JSONiq, functions are looked up by name and arity, so the map is 100% correct. There can only be one function with the same name and arity.

The signature, however, is the list of sequence types (like  integer*, object+, etc), as many sequence types as the arity of the function, and this information should also be stored somewhere, either in the map, or as additional information in the function iterator classes.

This signature information can be used to insert "treat as" iterators below function iterators when you generate the runtime iterators. A treat as expression (I am not sure if you have implemented this yet) returns exactly its input, but throws an error if the type does not match. what is expected. You then get the runtime check "for free" if the treat as iterator is correctly implemented, because anything that goes to the function iterator will have been checked by the "treat as" iterator and you no longer need to do these checks in the function iterators.

If you haven't implemented treat as expressions yet, we can postpone until later.

We can handle corner cases and casts (double to decimal, etc) later as well.

PR #31, opened 2018-05-01
tryout: ISSUE 18

wscsprint3r:
I fixed the original nesting problem, simple nests work now. 

However, the first JSONiq example from the tutorial is still broken. The second issue that causes this is something related to the dynamic context missing the variables above. I try to add the parent dynamic context of the entire FLWOR expression to each closure that gets serialized. 

The problem is that I cannot serialize the Dynamic Context in any way, I originally thought it was because of the self-reference but no matter what I try, I get a 'task not serializable exception'. 

If you have any insights please let me know

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,9 +73,9 @@
              <version>3.2.0</version>
          </dependency>
          <dependency>
-             <groupId>com.esotericsoftware</groupId>
+             <groupId>com.esotericsoftware.kryo</groupId>
              <artifactId>kryo</artifactId>
-             <version>4.0.0</version>
+             <version>2.24.0</version>
          </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


ghislainfourny:
@wscsprint3r As a follow up, this will probably be solved when Can has finished implementing local clause execution. We are in the last mile of this now.

ghislainfourny:
@wscsprint3r can we now close this? I believe that Can's work on local FLWOR execution took care of it. :)

ghislainfourny:
> The problem may have been that you had nested Spark jobs, which is not allowed. With a fully local implementation in addition to the RDD-based implementation, it will be no problem any more to nest FLWOR expressions, as long as parallelization itself is not nested.

Exactly! :-)

PR #39, opened 2018-06-16
Issue33 math func

CanBerker:
All mathematical functions have been implemented.

When implementing "round" and "roundhalftoeven" I defined a rounding function in the "RoundFunctionIterator" which is used by both iterators. If you have suggestions for a better practise or location to define the function please let me know.

With the recently added functions the functions folder is getting quite crowded. Moving them to a "Numeric" folder or multiple folders such as "trigonometric", "exponential" and so on can be beneficial for organization.


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/RoundFunctionIterator.java
```diff
@@ -0,0 +1,95 @@
...
+        } else
+            throw new IteratorFlowException(RuntimeIterator.FLOW_EXCEPTION_MESSAGE + " round function", getMetadata());
+    }
+
+
+    /**
+     * Decimal Formatting Solution
+     * https://stackoverflow.com/questions/153724/how-to-round-a-number-to-n-decimal-places-in-java
+     */
+    public static Double getRoundedResult(Double value, Integer precision, RoundingMode rm) {
```
wscsprint3r: This needs to be moved to some utility class, it doesn't make sense to have it in the Iterator as a static method. 

It would be good if you could find a pre-existing solution since rounding of doubles is not an uncommon problem. 
This may or may not help you: 
https://commons.apache.org/proper/commons-math/javadocs/api-3.3/org/apache/commons/math3/util/Precision.html

src/main/resources/test_files/runtime/FunctionRoundHalfToEven.iq
```diff
@@ -0,0 +1,7 @@
+(:JIQS: ShouldRun; Output="( 0.0, 2.0, 2.0, 3567.81, 0.0, 35600.0)" :)
+roundhalftoeven(0.5),
```
ghislainfourny: the name should be round-half-to-even

src/main/resources/test_files/runtime/FunctionSqrt.iq
```diff
@@ -0,0 +1,6 @@
+(:JIQS: ShouldRun; Output="( 0.0, -0.0, 1000.0)" :)
+sqrt(0.0e0),
```
ghislainfourny: It would be great to add a test for sqrt of a negative number (double-checking which error should be returned in the specification)

ghislainfourny:
This is awesome work, Can. I have added my minor comments inline.

PR #40, opened 2018-07-09
Folder restructuring

CanBerker:
I have been restructuring the functions folder in this branch. If this receives positive feedback, I'm thinking of going through a similar process with the test files to improve organization. In attachment you can find a screenshot giving a better overview of the current structure. 
functions_structure

I have renamed folders in a way that they would match the same terminology from the XQuery specication ie. Title "numerics", subtitle(where I felt was necessary) "exponential", from https://www.w3.org/TR/xpath-functions-31/#numeric-functions 

I have renamed the folder previously named as "arithmetic" to "aggregate". I have a question to @wscsprint3r at this step. I have also added the count function to this category in accordance with the XQuery specification. While doing so I have also changed its base class from "LocalFunctionCallIterator" to "AggregateFunctionIterator". Please provide feedback if you think that there is something wrong/counter-intuitive with this change. I can revert that commit if needed.


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -21,15 +21,17 @@
...
 import sparksoniq.jsoniq.compiler.translator.expr.primary.FunctionCall;
-import sparksoniq.jsoniq.item.IntegerItem;
 import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
-import sparksoniq.jsoniq.runtime.iterator.functions.*;
-import sparksoniq.jsoniq.runtime.iterator.functions.arithmetic.AvgFunctionIterator;
-import sparksoniq.jsoniq.runtime.iterator.functions.arithmetic.MaxFunctionIterator;
-import sparksoniq.jsoniq.runtime.iterator.functions.arithmetic.MinFunctionIterator;
-import sparksoniq.jsoniq.runtime.iterator.functions.arithmetic.SumFunctionIterator;
+import sparksoniq.jsoniq.runtime.iterator.functions.arrays.ArraySizeFunctionIterator;
+import sparksoniq.jsoniq.runtime.iterator.functions.sequences.aggregate.*;
```
wscsprint3r: The * imports are in general, to be avoided but we can think about that later. I also forgot about them.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/numerics/trigonometric/TanFunctionIterator.java
```diff
@@ -34,7 +34,7 @@ public Item next() {
                             ItemMetadata.fromIteratorMetadata(getMetadata()));
                 }
                 else {
-                    throw new UnexpectedTypeException("Tan expression has non numeric args " +
+                    throw new UnexpectedTypeException("Tan expression has non numerics args " +
```
wscsprint3r: Was this change intentional, some of these instances, like this one,  are incorrect.

CanBerker: It was caused by refactoring, I'll fix them right away

PR #41, opened 2018-07-13
Issue 38 comparisons

CanBerker:
Implemented general comparisons.

Updated NullRunTimeIterator to fix Null Pointer Exceptions.


src/main/java/sparksoniq/jsoniq/compiler/parser/JsoniqBaseVisitor.java
```diff
@@ -1,4 +1,4 @@
-// Generated from /home/stefan/Work/sparksoniq/sparksoniq/src/main/java/sparksoniq/jsoniq/compiler/parser/Jsoniq.g4 by ANTLR 4.6
+// Generated from /home/can/Project/sparksoniq/src/main/java/sparksoniq/jsoniq/compiler/parser/Jsoniq.g4 by ANTLR 4.6
```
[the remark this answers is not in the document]
CanBerker: Sure I will do that when requesting a review. I tried a .gitignore solution but if I'm not mistaken I think the continuous integration didn't like removal of auto-generated files. Are you suggesting removing only the lines similar to the solution in the link below or am I missing something ?
http://blog.davydovanton.com/2015/11/14/ignore-file-lines-in-git/

src/main/resources/test_files/runtime/ComparisonGeneral.iq
```diff
@@ -0,0 +1,14 @@
...
+(2,3) != (2,3),
+(2,2) != (2,2),
+(2) != (2,2),
+(3) != (2,2),
+(2,3) < (1,5),
+(2,3) <= (1,5),
+(6,7) <= (1,5),
+(1,5) > (2,3),
+(1,5) >= (2,3),
+(1,5) >= (6,7)
```
ghislainfourny: We also need tests with the empty sequence: on the left, on the right, and on both sides.
We also need tests that produce type errors (string on one side, number on the other side, and various other combinations of incompatible types)

src/main/resources/test_files/runtime/ComparisonValue.iq
```diff
@@ -0,0 +1,13 @@
...
+2 ne 3,
+2 ne 2,
+2 lt 3,
+3 lt 2,
+3 gt 2,
+2 gt 3,
+2 le 2,
+2 le 1,
+2 ge 2,
+1 ge 2
```
ghislainfourny: We also need tests with the empty sequence: on the left, on the right, and on both sides.
We also need tests that produce type errors (string on one side, number on the other side, and various other combinations of incompatible types)

src/main/resources/test_files/runtime/ComparisonGeneral.iq
ghislainfourny: We also need tests for null. Null is only equal to itself, not to anything else (but no error), and is smaller than anything.

src/main/resources/test_files/runtime/ComparisonValue.iq
ghislainfourny: We also need tests for null. Null is only equal to itself, not to anything else (but no error), and is smaller than anything.

src/main/resources/test_files/runtime/ComparisonGeneral.iq
ghislainfourny: We also need tests on strings, booleans, between numbers (decimals, doubles, integers etc) as well as with objects/arrays to check that an error is thrown.

src/main/resources/test_files/runtime/ComparisonValue.iq
ghislainfourny: We also need tests on strings, booleans, between numbers (decimals, doubles, integers etc) as well as with objects/arrays to check that an error is thrown.

PR #43, opened 2018-08-23
Empty sequence support

CanBerker:
Fixed exceptions on Spark FLOWR clauses caused by empty sequences and nulls

First two problems were caused by accessing without empty checking. Once fixed they turned into NullPointerExceptions as well.

NullPointerExceptions fixed by introducing a new "EmptySequenceItem" which prints out "" when serialized and prevents NullPointerExceptions as the code now works with this item instead of the Java null.

Following queries which provided the respective exceptions now work as intended and return empty sequences. Related tests are also added.

let $o := (()) return $o()
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.IndexOutOfBoundsException: 

let $o := (()) return size($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.IndexOutOfBoundsException:

let $o := (1, 2, "a", [], 5e+3) return descendant-objects($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.NullPointerException

let $o := ([]) return flatten($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.NullPointerException

let $o := ({}) return descendant-arrays($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.NullPointerException


ghislainfourny:
This is great, Can. Thanks!

Can you make sure to add all queries mentioned in your comment as new tests to avoid regressions?

Also, note that (()) is redundant. You can simply write () for the empty sequence.

src/main/resources/test_files/runtime/FunctionArray/FunctionArrayDescendant5.iq
```diff
@@ -1,4 +1,4 @@
 (:JIQS: ShouldRun; Output="( )" :)
```
ghislainfourny: What is the meaning of Output="( )"? Is it specific to the tests? A query returning an empty sequence should simply serialize it as an empty string on the command line.

src/main/java/sparksoniq/jsoniq/item/EmptySequenceItem.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import sparksoniq.jsoniq.item.metadata.ItemMetadata;
+
+import javax.naming.OperationNotSupportedException;
+import java.math.BigDecimal;
+
+public class EmptySequenceItem extends AtomicItem {
```
ghislainfourny: We would need to discuss this. Such an item does not exist. The correct way to implement an empty sequence returned by an iterator is to make sure that hasNext() returns false -- there is no need for extra complexity.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArrayFunctionIterator.java
```diff
@@ -29,7 +31,7 @@ public Item getResult() {
         // if no results return empty sequence
```
ghislainfourny: Part of this code should go into an override of open(), in the same class (and in every iterator that has the issue). That way, _hasNext will be set to false before hasNext() is called. That should do the trick and EmptySequenceItem won't actually be returned.

src/main/resources/test_files/runtime/FunctionArray/FunctionArrayDescendant5.iq
ghislainfourny: Normally, EmptySequenceItem should not make it to the outer result of the query. The resulting empty string is the consequence of the iterator returning hasNext()=false. No item is ever returned here.

src/main/java/sparksoniq/jsoniq/item/EmptySequenceItem.java
ghislainfourny: Yes. I understand that this is tricky, but it is simpler than it seems. A null atomic value is really something fundamentally different: a sequence of one null item. An empty sequence is not an item: it is the absence of any item in the first place.

I will clarify tomorrow if needed.

src/main/resources/test_files/runtime/FunctionArray/FunctionArrayDescendant5.iq
ghislainfourny: Multiple results which are empty sequences: this can never happen to any query :-) The result is always exactly one sequence. In such a case, for example the query "(), (), ()", the result is simply the empty sequence.

src/main/java/sparksoniq/jsoniq/item/EmptySequenceItem.java
ghislainfourny: The EmptySequenceIterator should not return EmptySequenceItem. It should simply have hasNext() return false, and next() return an error. I recommend resisting the temptation to "copy" the behavior of null.

src/main/resources/test_files/runtime/FunctionArray/FunctionArrayDescendant5.iq
CanBerker: Your first explanation makes sense to me, however there remains the problem that I've encountered earlier regarding FLOWR clauses sent to spark. In the closure example that I'm sending in the picture, the result.iterator() is passed to the framework. When all items contained by the iterator are null (which is the empty sequence case) the framework throws a NullPointerException. This was the reason why I came up with the EmptySequenceItem which acts as a placeholder for Java null, hence preventing the exception. Can you see another solution for this?
image

Yes that's definitely correct for the jsoniq. I'd assumed the test suite behaved differently and hence results of individual queries would be aggregated together in a single output of the form ( , , , ...) . Yet it applies for the test suite as well :) Thanks for pointing it out.

ghislainfourny: If I understand you correctly, this will also get solved "for free" when hasNext() is fixed to return false. next() will thus never be called (and will thus never return a Java null either). Then, result will simply be the empty list, which is consistent with the empty sequence.

Does it make sense?

CanBerker: Yes that might be the case for some queries but I'm afraid I doesn't fit all cases. For example the following image shows the descendant-array function:
image
Iterator of the function itself is opened at least once, hence triggering the calculations. The contained sequence(sequenceIterator) may be an empty sequence by itself or may result in an empty sequence following the calculations of getDescendantArrays. However, in both cases the function iterator has to return an item to the closure, which we return as null. This null causes the NullPointerException. If the list itself is empty however, which I have implemented in the v2 branch(other PR), then it works as you've described above. In order to get the list empty, I have removed the null element from the closure.

CanBerker:
This issue is being addressed in another PR:
https://github.com/Sparksoniq/sparksoniq/pull/44

Closing this PR the avoid confusion.

PR #44, opened 2018-08-23
V2 empty seq

CanBerker:
**Second take** on the empty sequence problem:

First two commits fix the out of bounds and illegal next calling issues.

Last commit is experimental and fixes the nullpointer exceptions occurring in the spark framework. If result.iterator() contains null elements which means that the result of the query is an empty sequence, an exception is thrown. By removing the said null elements, no exceptions are met. However, I'm worried that this solution might be crude and there might be corner cases that I'm missing.

Following queries which provided the respective exceptions now work as intended and return empty sequences. Related tests are also added.

let $o := (()) return $o()
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.IndexOutOfBoundsException:

let $o := (()) return size($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.IndexOutOfBoundsException:

let $o := (1, 2, "a", [], 5e+3) return descendant-objects($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.NullPointerException

let $o := ([]) return flatten($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.NullPointerException

let $o := ({}) return descendant-arrays($o)
✔ [ERROR] Job aborted due to stage failure: ..... java.lang.NullPointerException

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
@@ -35,7 +35,7 @@ public class CommaExpressionIterator extends LocalRuntimeIterator {
 
     @Override
     public Item next() {
-        if(currentChild == null) {
+        if(this._hasNext) {
             return getResult();
         }
         throw new IteratorFlowException("Invalid next() call in Comma expression", getMetadata());
@@ -50,8 +50,9 @@ public class CommaExpressionIterator extends LocalRuntimeIterator {
         this._currentIndex = 0;
         this.results = new ArrayList<>();
 
-        this._children.forEach(c -> c.reset(this._currentDynamicContext));
+        // this._children.forEach(c -> c.reset(this._currentDynamicContext));
         for (RuntimeIterator child:this._children) {
+            child.open(_currentDynamicContext);
             if (child.hasNext()) {
                 results.add(child.next());
             }
@@ -64,26 +65,12 @@ public class CommaExpressionIterator extends LocalRuntimeIterator {
         }
     }
 
-    @Override
-    public void reset(DynamicContext context){
-        super.reset(context);
-        currentChild = null;
-    }
-
-    @Override
-    public boolean hasNext(){
-        return currentChild == null ||
-                this._children.indexOf(currentChild) < this._children.size() - 1 ||
-                currentChild.hasNext();
-    }
-
     protected Item getResult() {
         if (_currentIndex == results.size() - 1)
             _hasNext = false;
         return results.get(_currentIndex++);
     }
 
-    private RuntimeIterator currentChild = null;
     private int _currentIndex;
     private List<Item> results;
 }
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
@@ -46,10 +46,7 @@ public class CommaExpressionIterator extends LocalRuntimeIterator {
 
     @Override
     public void open(DynamicContext context) {
-        if (this._isOpen)
-            throw new IteratorFlowException("Runtime iterator cannot be opened twice", getMetadata());
-        this._isOpen = true;
-        this._currentDynamicContext = context;
+        super.open(context);
         this._currentIndex = 0;
         this.results = new ArrayList<>();
 
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
@@ -17,7 +17,7 @@
  * Author: Stefan Irimescu
  *
  */
- package sparksoniq.jsoniq.runtime.iterator;
+package sparksoniq.jsoniq.runtime.iterator;
 
 import sparksoniq.jsoniq.item.Item;
 import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
@@ -29,20 +29,23 @@ import java.util.List;
 
 public class CommaExpressionIterator extends LocalRuntimeIterator {
 
+    private int _currentIndex;
+    private List<Item> results;
+
     public CommaExpressionIterator(List<RuntimeIterator> childIterators, IteratorMetadata iteratorMetadata) {
         super(childIterators, iteratorMetadata);
     }
 
     @Override
     public Item next() {
-        if(this._hasNext) {
+        if (this._hasNext) {
             return getResult();
         }
         throw new IteratorFlowException("Invalid next() call in Comma expression", getMetadata());
     }
 
     @Override
-    public void open(DynamicContext context){
+    public void open(DynamicContext context) {
         if (this._isOpen)
             throw new IteratorFlowException("Runtime iterator cannot be opened twice", getMetadata());
         this._isOpen = true;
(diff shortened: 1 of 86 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -132,11 +132,15 @@ public IteratorMetadata getMetadata() {
...
+        Item result = null;
+        if(iterator.hasNext()) {
+            result = iterator.next();
+            if (iterator.hasNext()) {
+                throw nonAtomicException;
+            }
+        }
         iterator.close();
-        if (!(type.isInstance(result)))
+        if (result != null && !(type.isInstance(result)))
```
ghislainfourny: Yes, I think that this is fine this way. This function is purely an internal helper that isn't exposed to the outside world, so you can use the Java null as a special return value that says "there is no such single item".

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArraySizeFunctionIterator.java
```diff
@@ -43,6 +43,9 @@ public Item next() {
             this._hasNext = false;
             RuntimeIterator arrayIterator = this._children.get(0);
```
ghislainfourny: This is another example of when you can/should move the array==null test as well as the two lines above to the end of an overriden open() method (after the super call). That way, if array is tested to be the Java null, you can set this._hasNext to false already when opening the iterator and next() won't even be called by the consumer.

In many iterators, a similar thing needs to be done, i.e., move some code from next() to an overriden open() in order to know right away whether an empty sequence is returned or not. This approach is very standard in pull-based querying engines (Zorba does it as well).

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/object/ObjectDescendantFunctionIterator.java
```diff
@@ -20,7 +20,6 @@ public ObjectDescendantFunctionIterator(List<RuntimeIterator> arguments, Iterato
     public Item next() {
         if (this._hasNext) {
             if (results == null) {
```
ghislainfourny: Likewise -- this initialization should be done at the end of an overriden open() method, as well as setting _hasNext to false if there are no results at all.

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
```diff
@@ -64,9 +61,13 @@ public void open(DynamicContext context){
             throw new IteratorFlowException("Variable reference iterator already open", getMetadata());
         this._currentDynamicContext = context;
         this.currentIndex = 0;
-        this.items = null;
-        this._hasNext = true;
-
+        this.items = this._currentDynamicContext.getVariableValue(this._variableName);
+        if(items.size() == 0) {
+            this._hasNext = false;
```
ghislainfourny: Yes, exactly like this :-) This is the general idea.

src/main/java/sparksoniq/spark/closures/ReturnFlatMapClosure.java
```diff
@@ -43,6 +44,7 @@ public ReturnFlatMapClosure(RuntimeIterator expression) {
         while (_expression.hasNext())
             result.add(_expression.next());
         _expression.close();
+        result.removeIf(Objects::isNull);
```
ghislainfourny: I understand the temptation to do this, but this is risky -- if there are nulls (which there should never be), you will no longer know: it is a bit like putting the dust under the carpet or switching off the fire alarm rather than fighting the fire :-) It is better practice to let the exception be thrown so that you can fix all the places were a Java null lands in here (that way, you will catch all iterators in which a move of code from next() to open() is necessary).

src/main/java/sparksoniq/spark/iterator/SparkRuntimeIterator.java
```diff
@@ -70,6 +70,10 @@ public Item next(){
                 result = _rdd.collect();
         }
 
+        if(result.size() == 0){
```
ghislainfourny: Likewise, this and the above initialization should go to open().

CanBerker:
Calculations in iterators have been moved to the overrides of open() function. Since I've worked with each iterator in a different commit, there is a huge number of them. I can squash them together into a single commit after receiving your approval to keep a cleaner commit history.

src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
```diff
@@ -17,59 +17,57 @@
...
-        currentChild = null;
-    }
+        this._currentIndex = 0;
+        this.results = new ArrayList<>();
 
-    @Override
-    public void reset(DynamicContext context){
-        super.reset(context);
-        currentChild = null;
+        // this._children.forEach(c -> c.reset(this._currentDynamicContext));
```
wscsprint3r: I understand the problem you are trying to solve but I think this is taking it a bit too far. Comma expressions, arrays and any Spark Iterator might be dealing with a very large amount of results and pre-computing all of the results in the open() method will potentially remove all the laziness resulting in significantly lower performance. @ghislainfourny what do you think?

src/main/java/sparksoniq/spark/iterator/SparkRuntimeIterator.java
wscsprint3r: @ghislainfourny  that will cause the RDD computation to be executed before next() is actually called. Do we want this?

src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
wscsprint3r: Ideally we would only peek at the first item in the open() method in order to return false for hasNext, instead of computing all the results, at least in some cases.

ghislainfourny:
As discussed today: this PR will need to be split into many, smaller PRs (typically one per iterator, possibly grouped when semantically similar), and we will have a discussion on each one of them separately and in parallel.

src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
ghislainfourny: Stefan has a good point -- we discussed this yesterday with Can.

The idea is to be as lazy as possible and *only* do the computations in open that are absolutely necessary (such as getting a lookahead of one item for hasNext() to return the correct result).

In some (rare) cases, it is necessary to compute everything in open. This is known in literature, and called "materialization". For example, this is the case with the reverse function: you can only return the first item by consuming the entire input sequence. This should be done only if absolutely necessary because of the semantics of the iterator.

This will partially be the case with local group by/order by execution, as the first tuple can only be returned after consuming the entire input stream.

ghislainfourny: Regarding iterators with an underlying RDD: the design pretty much depends on the Spark API. If the Spark API allows us to "peek" at the RDD to see if it is empty, then it is not necessary to precompute in open() before haxNext() is called. If the Spark API allows getting the items one by one, this is also not necessary. If all Spark allows is "taking" all items at once, then materialization is necessary (but of course, in practice, this will only be done for small sequences, and an error can be thrown for bigger sequences).

PR #45, opened 2018-08-23
fix objectlookup single item bug

CanBerker:
Issues with the following queries have been solved. Only simple test cases added. I have some questions to ask during the meeting, regarding diverse test cases.
({"foo" : "bar"}, {"foo" : "foobar"}).foo -- ( bar, foobar)

().foo -- ()


CanBerker:
Thank you. Yes of course that was the work plan I had in mind. The ObjectLookupIterator's open() method was overridden in this PR since without it, the existing tests would break due to faulty(lacking) implementation. 

I am also aiming to add more tests regarding the not working examples, however I have some questions that I'm planning to ask today on that topic.

PR #46, opened 2018-08-26
Issue23 arithmetic

CanBerker:
The problem was caused by faulty type checking at the DoubleItem class.

Test case added for (4 + 2.3) instance of decimal and it successfully returns false.


wscsprint3r:
lgtm

wscsprint3r:
Nice fix!

PR #47, opened 2018-08-26
Issue24 div

CanBerker:
Div operation between two operations has been changed result in a decimal in order to preserve precision. 
4 div 3 returns 1.3333333333

Test files have been updated accordingly. I have one question though:
---2 + 3 * 4 - 2 * 5 div 2 mod 6 * 2
The expression above previously returned **0** (same result on Zorba), however, currently it returns **0E-10** due to the precision of Decimal type. These two values appear to be equal in Zorba, yet I was wondering if this change in output format is undesirable.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/MultiplicativeOperationIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/MultiplicativeOperationIterator.java
@@ -68,7 +68,10 @@ public class MultiplicativeOperationIterator extends BinaryOperationBaseIterator
                         this.result = new IntegerItem(l * r, ItemMetadata.fromIteratorMetadata(getMetadata()));
                         break;
                     case DIV:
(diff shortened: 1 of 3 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/MultiplicativeOperationIterator.java
```diff
@@ -68,7 +68,10 @@ public void open(DynamicContext context) {
                         this.result = new IntegerItem(l * r, ItemMetadata.fromIteratorMetadata(getMetadata()));
                         break;
                     case DIV:
-                        this.result = new IntegerItem(l / r, ItemMetadata.fromIteratorMetadata(getMetadata()));
+                        BigDecimal decLeft = Item.<BigDecimal>getNumericValue(left, BigDecimal.class);
```
wscsprint3r: This will change the return type to decimal even for normal integer division. 4/2 will return 2.0 which is counter intuitive. Maybe we should check if the integer division returns a 0 remain and if it does, don't convert it to a big decimal.

PR #48, opened 2018-08-27
Issue9 adv obj lookup

CanBerker:
Parenthesized expression and variable reference support has been implemented for object lookup. Test cases added.

Predicates are improved. They now support indexing for the left hand side for expressions that evaluate to integers eg: (1 to 10)[2] --> 2  (results tested by comparing to zorba, test cases added).

I have one question: "What's the result of applying a context expression($$) to an object?", for example in the following scenerio:
let $o := {"foobar": 2,  "1" : "bar"}
return $o.$$

Currently this functionality is partially implemented and returns empty sequence. I'll implement the rest depending on your input on the topic.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
@@ -575,6 +575,9 @@ public class JsoniqExpressionTreeVisitor extends sparksoniq.jsoniq.compiler.pars
         } else if(ctx.vr != null) {
             this.visitVarRef(ctx.vr);
             expr = this.currentPrimaryExpression;
+        } else if(ctx.ci != null) {
+            this.visitContextItemExpr(ctx.ci);
+            expr = this.currentPrimaryExpression;
         }
 
         node = new ObjectLookupExtension(expr, createMetadataFromContext(ctx));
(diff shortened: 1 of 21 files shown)


src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
```diff
@@ -559,15 +559,28 @@ else if(child instanceof JsoniqParser.ArrayUnboxingContext) {
...
-        StringLiteral literal = null;
-        if(ctx.lt != null)
-            literal = new StringLiteral(ValueTypeHandler.
-                getStringValue(ctx.lt), createMetadataFromContext(ctx));
-        else if(ctx.nc != null)
-            literal = new StringLiteral(ctx.nc.getText(), createMetadataFromContext(ctx));
-        else if(ctx.kw != null)
-            literal = new StringLiteral(ctx.kw.getText(), createMetadataFromContext(ctx));
-        node = new ObjectLookupExtension(literal, createMetadataFromContext(ctx));
+        Expression expr = null;
```
wscsprint3r: I hate this class, I will rewrite it someday.

src/main/java/sparksoniq/jsoniq/runtime/iterator/EmptySequenceIterator.java
```diff
@@ -36,5 +37,6 @@ public boolean hasNext() {
     @Override
     public Item next() {
         return null;
+        //throw new IteratorFlowException("Invalid next() call on Empty Sequence", getMetadata());
```
wscsprint3r: Delete this

CanBerker:
This code has been used in another PR https://github.com/Sparksoniq/sparksoniq/pull/59. If approved, this PR will be cancelled and removed

PR #52, opened 2018-09-05
Variable reference iterator

CanBerker:
Salvaged the approved implementation of VariableReferenceIterator along with its test.

Added improvements for handling of empty sequence as the final result for both the application and the test suite.


src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
```diff
@@ -60,13 +57,16 @@ public void reset(DynamicContext context){
...
         this.currentIndex = 0;
-        this.items = null;
-        this._hasNext = true;
-
+        this.items = this._currentDynamicContext.getVariableValue(this._variableName);
+        if(items.size() == 0) {
+            this._hasNext = false;
+        }
+        else {
+            this._hasNext = true;
```
ghislainfourny: Perfect! :-)

PR #53, opened 2018-09-05
implement ObjectLookupItr

CanBerker:
Lazy evaluation implemented for open/next structure of the ObjectLookupIterator.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/ObjectLookupIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/ObjectLookupIterator.java
@@ -43,58 +43,55 @@ public class ObjectLookupIterator extends LocalRuntimeIterator {
 
     @Override
     public void open(DynamicContext context) {
-        if(this.isOpen())
-            throw new IteratorFlowException("Variable reference iterator already open", getMetadata());
-        this._currentDynamicContext = context;
-        this._currentIndex = 0;
-        this.results = null;
-        this._hasNext = true;
+        super.open(context);
+        this._nextIndex = 0;
+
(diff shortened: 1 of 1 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/ObjectLookupIterator.java
```diff
@@ -43,58 +43,55 @@ public ObjectLookupIterator(RuntimeIterator object, StringRuntimeIterator string
...
-        this.results = null;
-        this._hasNext = true;
+        super.open(context);
+        this._nextIndex = 0;
+
+        RuntimeIterator iterator = this._children.get(0);
+        this.items = getItemsFromIteratorWithCurrentContext(iterator);
+
+        this._children.get(1).open(_currentDynamicContext);
+        this._lookupKey = this._children.get(1).next();
```
ghislainfourny: You need to first check this._children.get(1).hasNext(). hasNext() must always be called before next() is called.

Adding a test for {"foo":"bar"}.() should make sure that a clean error is thrown (as exactly one string item is expected on the right).

ghislainfourny:
It looks good, just one comment.

It would be great to add tests to catch all corner cases like

{}.foo -> empty sequence
{}.() -> error
{}.("foo", "bar") -> error
[1, 2, 3].foo -> empty sequence
({"foo" : "bar"}, {"foo" : "foobarbar"}).foo - >"bar", "foobarbar"
({"foo" : "bar"}, "test", [ "array" ], {"foo" : "foobarbar"}).foo -> "bar", "foobarbar"

CanBerker:
Lazy evaluation and the suggested test cases implemented in another PR: https://github.com/Sparksoniq/sparksoniq/pull/59

PR #54, opened 2018-09-08
Comma expression itr

CanBerker:
CommaExpressions are lazily evaluated.

Once test case on switchRuntimeIterator has been removed from this PR to be fixed on the PR related to SwitchRuntimeIterator https://github.com/Sparksoniq/sparksoniq/pull/55


src/main/java/sparksoniq/jsoniq/runtime/iterator/CommaExpressionIterator.java
```diff
@@ -27,49 +28,50 @@
...
-                currentChild.hasNext();
+    public void setNextResult() {
+        _nextResult = null;
+
+        for (int i = _nextIndex; i < this._children.size(); i++) {
+            RuntimeIterator childIterator = this._children.get(i);
+            childIterator.open(_currentDynamicContext);
+            if (childIterator.hasNext()) {
+                _nextResult = childIterator.next();
+                _nextIndex = i + 1;
```
ghislainfourny: I am not sure it is quite correct if one of the input expressions returns more than one item. Can you add tests like:

(("foo", "bar"), (), ("foo", "bar", "foobar"), "foo)

?

You need to iterate at two levels:
- within one child iterator
- and across child iterators once a child iterator was fully consumed

CanBerker:
Two Level iteration has been implemented. Another PR (https://github.com/Sparksoniq/sparksoniq/pull/57/commits) has been merged into this PR. 

ValueComparisonEmptySequence.iq test has been deactivated to be handled in a future PR.

PR #56, opened 2018-09-08
Array size itr

CanBerker:
Lazy evaluation and empty sequence support added to array size iterator

PS: I'm purposefully adding the test cases such as ( (), () ) . It's semantically equivalent to an empty sequence however, it prevents the antipattern of checking for emptySequenceIterator. This test case enforces that we're checking the hasNext to be false rather than simply checking if iterator is an empty sequence iterator. If latter is done (()) would be handled successfully however since ( (), (), ) is a commaExpression at the higher level, emptySequence check would've failed.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArraySizeFunctionIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArraySizeFunctionIterator.java
@@ -24,9 +24,11 @@ import sparksoniq.jsoniq.item.ArrayItem;
 import sparksoniq.jsoniq.item.IntegerItem;
 import sparksoniq.jsoniq.item.Item;
 import sparksoniq.jsoniq.item.metadata.ItemMetadata;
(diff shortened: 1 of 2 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArraySizeFunctionIterator.java
```diff
@@ -37,16 +39,31 @@ public ArraySizeFunctionIterator(List<RuntimeIterator> arguments, IteratorMetada
...
-            RuntimeIterator arrayIterator = this._children.get(0);
+
             ArrayItem array = getSingleItemOfTypeFromIterator(arrayIterator, ArrayItem.class);
             return new IntegerItem(array.getSize(), ItemMetadata.fromIteratorMetadata(getMetadata()));
         }
         throw new IteratorFlowException(RuntimeIterator.FLOW_EXCEPTION_MESSAGE + "SIZE function",
                 getMetadata());
     }
 
+    RuntimeIterator arrayIterator;
```
wscsprint3r: Can be private

PR #57, opened 2018-09-09
move calculation to open() for FlworExpressionSparkRuntimeItr

CanBerker:
Calculation is moved to open() for only the specified SparkRuntimeIterator. I also want to point out the earlier feedback given by Stefan https://github.com/Sparksoniq/sparksoniq/pull/44#discussion_r213345584

The base class' next method is unchanged and will be handled once all the classes that implement it are updated (eg. Parallelize)


src/main/java/sparksoniq/spark/iterator/flowr/FlworExpressionSparkRuntimeIterator.java
```diff
@@ -67,6 +69,39 @@ public FlworExpressionSparkRuntimeIterator(FlowrClauseSparkIterator startClause,
...
+        }
+    }
+
+    @Override
+    public Item next(){
+        if(!this._isOpen) {
+            throw new IteratorFlowException("Runtime iterator is not open", getMetadata());
+        }
+
+        if (this.hasNext()) {
```
ghislainfourny: As a general rule, next() does not need to call hasNext(): it can simply look at the _hasNext field directly (assuming hasNext() consistently returns _hasNext). This avoids an extra call.

CanBerker:
This PR's commit has been moved to CommaExpressionItr PR at https://github.com/Sparksoniq/sparksoniq/pull/54

PR #58, opened 2018-09-09
Descendant functions

CanBerker:
Lazy evaluation implemented for descendant functions. Lazy evaluation is implemented in a different manner due to the added complexity caused by recursive structure of the function calls. Each item in the original sequence is recursively investigated to find descendant items of the desired type which are then stored in a queue. Once results returned from the first item is exhausted through next calls, the following item is recursively investigated in the same manner and its results are stored in the queue and so on. 

Empty sequence test cases are activated following the bugfix applied in the previous PR:  https://github.com/Sparksoniq/sparksoniq/pull/57


ghislainfourny:
Here I will trust you and Stefan and simply look at the tests. I think that you now clearly understood how hastNext() and next() work together and how to take care of the empty sequence.

PR #59, opened 2018-09-16
Advanced obj lookup

CanBerker:
While testing for the cases stated at https://github.com/Sparksoniq/sparksoniq/pull/53 some of the problems that I previously solved in Advanced Object Lookup(unmerged and on hold PR https://github.com/Sparksoniq/sparksoniq/pull/48) were met again. So I have taken all the improvements made in the https://github.com/Sparksoniq/sparksoniq/pull/48, then implemented lazy evaluation for ArrayLookup Objectlookup and PredicateIterator. Test cases are added.

Notes from Issue9 adv obj lookup https://github.com/Sparksoniq/sparksoniq/pull/48 are below:
Parenthesized expression and variable reference support has been implemented for object lookup. Test cases added.

Predicates are improved. They now support indexing for the left hand side for expressions that evaluate to integers eg: (1 to 10)[2] --> 2 (results tested by comparing to zorba, test cases added).

**I have one question:** "What's the result of applying a context expression($$) to an object in object lookup?", for example in the following scenerio:
let $o := {"foobar": 2, "1" : "bar"}
return $o.$$

Also are there more complicated examples using context item expressions?

Currently this functionality is partially implemented and returns empty sequence. I'll implement the rest depending on your input on the topic.


ghislainfourny:
The semantics of $$ is that it is taken from the dynamic context. $$ is only defined in specific cases, not in general.

let $o := {"foobar": 2, "1" : "bar"}
return $o.$$ 

throws an error.

But there are cases like

let $o := {"foobar": 2, "1" : "bar"}
("foo", "bar", "foobar")[$o.$$ eq 2]

where this works, and it returns "foobar".

CanBerker:
$$ defined for **filter expression**, following cases are covered and test cases are added.
let $o := {"foobar": 2, "1" : "bar"}
return $o.$$

let $o := {"foobar": 2, "1" : "bar"}
return ("foo", "bar", "foobar")[$o.$$ eq 2]

However, for the **bang operator** (I believe it's also called **MapOperator** http://www.jsoniq.org/docs/JSONiq/html-single/index.html#idm16200896), the expression definition and related mechanisms seem to be _not_ implemented. (I think the same case also applies to the **CountClause** which is right above in the documentation.

After my initial investigation, I have some guesses on actions that can be taken which I listed below. I believe they're not exhaustive or fully accurate however, I hope they'll facilitate to catch my misunderstandings.
1-) **MapClause**(?) definition under ..../sparksoniq/jsoniq/compiler/translator/flowr (?)
2-) **visitSimpleMapExpr** override in JsoniqExpressionTreeVisitor
3-) **Iterator** implementation (however, I'm confused under which directory such iterator should be placed -- local or sparkruntime)

PR #60, opened 2018-09-17
Array fn iterators

CanBerker:
Remaining array related functions (**flatten** and **members**) are implemented with empty sequence support and lazy evaluation. Test cases added


src/main/resources/test_files/runtime/FunctionArray/FunctionArrayMembers1.iq
```diff
@@ -0,0 +1,2 @@
+(:JIQS: ShouldRun; Output="( mercury, venus, earth, mars, 1, 2, 3)" :)
+let $planets :=  ( "foo", { "foo" : "bar "}, [ "mercury", "venus", "earth", "mars" ], [ 1, 2, 3 ]) return members($planets)
```
ghislainfourny: More tests would be needed such as testing that members([]) is the empty sequence, and that members([[1, 2], [3, 4]) is the sequence of two arrays ([1, 2], [3, 4}).

The same with flatten, except that flatten([[1, 2], [3, 4]) is the sequence of four integers (1, 2, 3, 4).

PR #63, opened 2018-09-17
Operational itr

CanBerker:
Empty sequence handling implemented for string concat and Unary Operation

Lazy evaluation and empty sequence handling implemented for range and additive Operations. 

Test cases added for improvements and logics.

Primes.iq test has been disabled and its relevant bug regarding quantifiedExpressionVar will be fixed in a seperate PR.


src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/AdditiveOperationIterator.java
```diff
@@ -28,68 +28,81 @@
...
         super(left,right,operator, iteratorMetadata);
     }
 
     @Override
     public AtomicItem next() {
-
-        if(!this._hasNext)
-            throw new IteratorFlowException(RuntimeIterator.FLOW_EXCEPTION_MESSAGE, getMetadata());
-        if(_leftIterator instanceof EmptySequenceIterator || _rightIterator instanceof EmptySequenceIterator){
+        if(this.hasNext()){
```
ghislainfourny: _hasNext will be faster.

src/main/resources/test_files/runtime/Operational/AdditiveOp2.iq
```diff
@@ -0,0 +1,10 @@
+(:JIQS: ShouldRun; Output="(2, 2.0, 20.0, 2.0, 11.0, 11.0)" :)
+1 + 1,
```
ghislainfourny: Are there any tests for other operations (*, -, div...)?

CanBerker: Some of these tests existed in various files. These have been reorganized in the following commits. And more are introduced in this PR and other PR's which originate from this branch.

CanBerker:
Following case implemented for and/or operations 
If the input sequence has more than one item, and the first item is not an object or array, an error is raised.

( 1, 2, 3 ) or false

PR #65, opened 2018-09-17
QuantifiedExpressionVarItr

CanBerker:
implement lazy evaluation for quantifiedExpressionVar, remove unnecessary resets, fix testing bug


CanBerker:
(:JIQS: ShouldRun; Output="( 2, 3, 5, 7, 11, 13, 17, 19)" :)
for $i in parallelize(2 to 20)
where every $j in 2 to $i - 1 satisfies $i mod $j gt 0
return $i

This is the test that's currently in place (It was renamed in the commit and contents were not visible).

PR #76, opened 2018-09-24
Aggregate functions

CanBerker:
Lazy evaluation and empty sequence support implemented for aggregate functions (avg, max, min sum). Test cases added


ghislainfourny:
Can you confirm that, in the current state of affairs, if the underlying iterator is on an RDD, it materializes it locally and computes the aggregation locally (which explains the issue Ingo reports)?

It would be nice to open an issue for the future in order to add a switch to these five iterators. If the underlying iterator is on an RDD, then the computation of the aggregated value should be pushed down to Spark, wrapped into a number item, and then returned by these iterators. I think it would be straightforward, but no need to overload this PR.

CanBerker:
I'll go through the full details of my investigation in order not to miss any detail that might be of value to you. Please excuse the long comment.

Firstly, testing the count function on the following local case:
let $a := (1,2,3,4,5)
return count($a)

iterator is assigned to a variable(sequenceIterator) from which we can extract whether it's an RDD or not. Then the next calls are carried out via an internal function (getItemsFromIteratorWithCurrentContext). During the next calls made through this internal function VariableReferenceIterator is used and the entire computation is materialized locally as expected.

1:
image

Secondly for the following RDD case:
count(json-file("../confusion_sample.json"))

The sequenceIterator contains an RDD(as seen in the watches). This time the iterator is a ParseJSONiterator which inherits its next() method from the SparkRunTimeIterator shown in the third screenshot.

2:
image
3:
image

This next() call indeed contains a take function with item limit, causing the 100 item result in the count function as the aggregation is calculated locally.

Should I create a new issue or since this topic is very closely related to Ingo's issue, should I append the details there?

PS: I believe the correct place for the switch is very clear in my mind. However, how the computation should be transferred to Spark after realizing the iterator is an RDD still somewhat puzzles me. I'm hoping to discuss that matter, when this topic takes priority.

wscsprint3r:
Looks good to me, great job!

ghislainfourny:
@Can Yes, just open an issue. We can discuss this.

Only count(json-file("../confusion_sample.json")) and queries in which the child iterator directly has an RDD needs to be mentioned in this issue for now.

The idea, on the RDD side of the switch, is to *not* recursively invoke next() as is done on the non-RDD side (this is what otherwise triggers the take()...), but instead simply to get the actual RDD (potentially dynamically casting the child iterator down to a type that supports getting the actual RDD), invoke a Spark count() transformation on it, wrap the obtained number into a JSONiq integer and return it. It is that simple.

Supporting via variables is more advanced and lower priority, no need to look into it but you can still open a separate, lower priority issue to remember it.

PR #81, opened 2018-10-05
change next() call of EmptySequence Iterator

CanBerker:
Since hasNext is always false in emptySequenceIterator iterator by definition. I believe throwing an exception rather than returning null would make more sense in the next() call.


wscsprint3r:
lgtm-bb8

PR #82, opened 2018-10-05
Cleanup empty seq sup

CanBerker:
Unused function/variable definitions in array and object function iterators removed.

Performed simple cleanups for some iterators on error messages and variable definitions.


wscsprint3r:
Great!

PR #86, opened 2018-10-14
handle CTRL+D (EOF) exception

CanBerker:
Following issue has been solved: https://github.com/Sparksoniq/sparksoniq/issues/69

While the exception message for CTRL+D is getting printed, NullPointerException was thrown. 

CTRL+D has been handled separately to provide the stated functionality. There might to be more comprehensive solutions for the special inputs JLineShell. Such an improvement can be planned for future if necessary.


src/main/java/sparksoniq/io/shell/JiqsJLineShell.java
```diff
@@ -118,8 +120,11 @@ private void initialize() throws IOException {
     }
 
     private void handleException(Exception ex) {
-        if (ex != null && !(ex instanceof UserInterruptException))
+        if (ex != null && ex instanceof EndOfFileException) {
```
wscsprint3r: Use an outside if for the null check and 2 nested ifs or a switch for the instance check.

PR #96, opened 2018-10-24
ISSUE95: Implement Count FLWOR clause

wscsprint3r:


ghislainfourny:
Was it fun to implement? :-)

src/main/java/sparksoniq/spark/closures/CountClauseClosure.java
```diff
@@ -0,0 +1,29 @@
...
+        this.variableName = variableName;
+        this.metadata = metadata;
+    }
+
+    @Override
+    public FlworTuple call(Tuple2<FlworTuple, Long> inputTuple) throws Exception {
+        FlworTuple result = inputTuple._1;
+        result.putValue(
+                variableName,
+                new IntegerItem(inputTuple._2.intValue(), ItemMetadata.fromIteratorMetadata(metadata)),
```
wscsprint3r: @ghislainfourny not sure about the type here, we may need a long type

wscsprint3r:
@ghislainfourny it was faster than expected :)

wscsprint3r:
@ghislainfourny I still need to try some cases but sorting/grouping will break the counting I assume.

ghislainfourny:
@wscsprint3r I am not sure if you mean this with "break the counting", but it is expected that if a count clause is followed by a group or order clause, this will have an impact.

for $i in (10, 9, 8)
count $j
order by $i
return $j

will definitely return (3, 2, 1) and not (1, 2, 3)

for $i in (10, 9, 8)
count $j
let  $k := $i mod 2
group by $k
return [ $k, [ $j ] ]

will definitely return [ 0, 2 ], [ 1, [ 1, 3 ] ].

CanBerker:
> for $i in (10, 9, 8)
> count $j
> let $k := $i mod 2
> group by $k
> return [ $k, [ $j ] ]
> 
> will definitely return [ 0, [ 2 ] ], [ 1, [ 1, 3 ] ].

I believe that this query return [ 0, [ 1, 3], [ 1, [ 2 ] ] 

@ghislainfourny Could you please check?

PR #97, opened 2018-10-24
Tested more efficient parsing.

ghislainfourny:


CanBerker:
This branch was accidentally merged and then cleaned up. The branch is saved but it is not functional at the moment (fails to run)

PR #102, opened 2018-11-15
Cli parameter fix

CanBerker:
In spark master parameters taken within the application override the CLI parameters. To simplify the program startup, master configs within the application have been removed. 

For the shell application, all masterconfigs are removed.

For the batch application (Main.java) with input/output files, most of the masterconfigs have been removed. A check using sparkcontext has been added to detect local/remote execution and proceed accordingly.

Local spark configurations in the SparkRuntimeTests.java file have been moved to RuntimeTests.java to support certain test cases needing Spark execution (Alternatively they can be moved to SparkTests, however this would break the organization where similar tests are co-located in the project)


CanBerker:
This branch is successfully merged to the master

PR #103, opened 2018-11-15
Upgrade to 0.9.3 and update documentation.

ghislainfourny:


ghislainfourny:
This upgrades to 0.9.3 and sets ant to 4.5.3 by default (we shouldn't use 4.6. Either 4.5.3 or 4.7, supported by Spark distributions).

src/main/java/sparksoniq/jsoniq/compiler/parser/JsoniqBaseVisitor.java
```diff
@@ -1,4 +1,4 @@
-// Generated from /home/can/Project/sparksoniq/src/main/java/sparksoniq/jsoniq/compiler/parser/Jsoniq.g4 by ANTLR 4.6
+// Generated from /Users/ghislain/Code/sparksoniq/src/main/java/sparksoniq/jsoniq/compiler/parser/Jsoniq.g4 by ANTLR 4.5.3
```
wscsprint3r: There is probably a way to remove this, we'll have to find it later.

PR #105, opened 2018-11-17
Refactor Item class, remove static methods

CanBerker:
As preparation to the following PR on sequence max-min methods, I have implemented Comparable interface in the Item class. Edit: Mentioned PR is https://github.com/Sparksoniq/sparksoniq/pull/106

Following this, I have went through a refactoring process to remove static methods and replaced them with functions that worked on instances. I believe this improves usability as well as provides better overriding capabilities for the child classes. @wscsprint3r If there is a point I'm missing regarding the original static implementations, please provide feedback.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/item/Item.java
+++ b/src/main/java/sparksoniq/jsoniq/item/Item.java
@@ -34,57 +34,54 @@ import java.util.Collection;
 import java.util.List;
 
 //TODO serialize with indentation
-public abstract class Item implements SerializableItem {
-    public static boolean isAtomic(Item resultItem) {
-        return resultItem instanceof AtomicItem;
-    }
+public abstract class Item implements SerializableItem, Comparable<Item> {
 
-    public static boolean isNumeric(Item item) {
-        return item instanceof IntegerItem || item instanceof DecimalItem || item instanceof DoubleItem;
+    public boolean isNumeric() {
+        return this instanceof IntegerItem || this instanceof DecimalItem || this instanceof DoubleItem;
     }
 
     //performs conversions for binary operations with a numeric return type
     //(int,double) -> double
     //(int,decimal) -> decimal
     //(decimal,double) -> double
-    public static Type getNumericResultType(Item left, Item right) {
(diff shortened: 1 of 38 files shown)


CanBerker:
That's a very good point! I'll take another look at the min/max implementations without altering static methods.

CanBerker:
This PR is on hold until performance investigations. The static vs dynamic implementations may be put to test then.

ghislainfourny:
@Can I thought about it and think these changes make a lot of sense. Feel free to solve the conflicts and merge.

wscsprint3r:
I don't remember having another reason for making the methods static apart from call performance

ghislainfourny:
I think the bottleneck in performance is unlikely to be due to static vs. dynamic calls, so in my understanding Can's proposal was more motivated by readability. It is of course certainly a good idea to double-check that we don't get any loss of performance.

In our current state, I feel that performance bottlenecks are rather dominated by object construction, as the recent improvements in performance were all related to constructing less objects and diminishing their footprint.

wscsprint3r:
@CanBerker what is the status of this PR?

CanBerker:
I haven't taken a look in this PR for some quite time. @ghislainfourny had altered the Item classes quite a bit so I was expecting a lot of conflicts(as seen above). However, the said refactoring removed a lot of bloat so it should be way simpler to factor out static methods. 

I think it might probably be easier to just check what was changed here and perform a similar refactoring in a new branch/PR. 

Lastly, there hasn't been any performance measurements regarding this yet. If we still have doubts there, it might be nice to benchmark after doing the necessary changes again.

ghislainfourny:
I think some changes still make sense -- however, the argument speaking in favor of doing so is probably more about having a nice and clean public Item API rather than performance.

CanBerker:
Indeed, that's a nice goal. I will go over this once again at my earliest convenience.

CanBerker:
I had gone over this issue and solved it in the following PR: https://github.com/RumbleDB/rumble/pull/294

Closing this pull request now.

PR #106, opened 2018-11-17
String max min support

CanBerker:
String support for min and max functions added. Sum and average functions were not implemented for non-numerics in Zorba, so they work only with numerics in Sparksoniq as well at the moment.

Test cases added, old test cases disabled (when PR is approved, they will be completely removed before merge).

I have a question for @ghislainfourny . In the current implementation, min and max operations are  accepting null as a valid input and processing it correctly (as a value smaller than all possible integers and strings). However in Zorba following case is observed:

**input:** max((null, "a"))
**output:** no URI):1,2: dynamic error [err:FORG0006]: promotion not possible

Which of the two is the desired behavior for sparksoniq? Should we support nulls for min-max as well, or should we throw an error?


ghislainfourny:
@CanBerker Many thanks for taking a look.

Regarding max/min involving nulls: very good point. We should stick to Zorba's behavior in order to avoid discrepancies between implementations. I will investigate the theory later.

Regarding pushing down to spark if the input is an RDD: yes, absolutely. You should use the max() and min() actions to push down to Spark and NOT collect locally. For future reference (just open an issue for now), there is also a mean() and sum() action.

Issue #37, opened 2018-06-04
Sequence cardinality functions


ghislainfourny:
Implement
 zero-or-one#1, one-or-more#1, exactly-one#1

PR #107, opened 2018-11-17
Implement seq. cardinality functions , add tests

CanBerker:
Sequence cardinality functions implemented with lazy evaluation. Test cases added.

@ghislainfourny Zorba has the following error codes for these functions.
err:FORG0003, fn:zero-or-one called with a sequence containing more than one item.
    Raised by fn:zero-or-one if the supplied value contains more than one item.
err:FORG0004, fn:one-or-more called with a sequence containing no items.
    Raised by fn:one-or-more if the supplied value is an empty sequence.
err:FORG0005, fn:exactly-one called with a sequence containing zero or more than one item.
    Raised by fn:exactly-one if the supplied value is not a singleton sequence.

These codes are not introduced in sparksoniq and no error code is printed out at the moment. What action would you suggest? Shall we introduce each one, or collect them under a single tag?

This PR deprecates the old attempt at: https://github.com/Sparksoniq/sparksoniq/pull/49

Edit: related issue #37


ghislainfourny:
Thank you for looking into the details.

We need to use the exact same codes (and add them to the Sparksoniq documentation), this is also officially specified this way, e.g.:

https://www.w3.org/TR/xpath-functions-30/#func-one-or-more

By the way: also these functions should be pushed to Spark in some way if the input is an RDD (this can easily be done with a count()).

CanBerker:
I have been taking a look at the earlier discussions for the materializations of RDDs and I came up with the two following questions which relate to the following comment: https://github.com/Sparksoniq/sparksoniq/pull/44#discussion_r21695459

**Question 1:** if a count action might be followed by a collect or similar actions, would **persisting** this rdd improve performance? (count can be used to check whether RDD hasNext or not)

**Question 2:** Instead of materializing the entire RDD with collect and processing the results list, is there a way to **materialize** smaller portions (eg 1 item)? Is this more efficient and desirable for lazy evaluation?
>>> Likely answer to Q2: https://stackoverflow.com/questions/32204227/collect-rdd-output-as-a-stream 
>>>The link mentions that it's not possible

ghislainfourny:
1. Yes, I think it could be a great idea in case you want to call multiple actions on the same RDD.

2. There are actions called first() and take(n) that may be what you are looking for. Yes, it is much better in case you don't want to look at the entire RDD (for example for pushing down json-file("myfile.json")[1].

CanBerker:
I'll look into the implications of the first question.

Regarding the second point I would like to elaborate on my efficiency question. I was curious whether spark efficiently supports this iterator logic of taking items one by one. More specifically, I was curious whether skipping the previously taken items and then taking the next one for each item is really an efficient operation on RDD's compared to collecting the RDD. Do we have a limit on what the size of an itemlist that we're going to materialize as such?

ghislainfourny:
The number of actions should still be limited to the bare minimum. Spark is not mean to be used in an iterator fashion.

I think that for such functions, it suffices to collect the first two items just once (if they exist). Based on this, it is already possible to say which sequence cardinalities (+, *, 1, ?)apply.

However, it is fine to keep pushing these functions down to Spark for another PR, we can simply open issues or make sure they are already there. I think that this PR is good to merge.

Issue #36, opened 2018-06-04
Sequence value functions


ghislainfourny:
Implement:
 distinct-values#1, distinct-values#2, index-of#2, index-of#3, deep-equal#2. deep-equal#3

PR #124, opened 2018-11-23
Implement sequence value functions with lazy eval.

CanBerker:
ObjectFunctionUtilities class replaced by ItemUtil.

distinct-values, index-of, deep-equal has been implemented with lazy evaluation.

Test cases added.

This PR deprecates #50 

Edit: Issue #36 addressed


src/main/resources/test_files/runtime/FunctionSequence/Value/FunctionDeepEqual1.iq
```diff
@@ -0,0 +1,6 @@
+(:JIQS: ShouldRun; Output="(false, false, true)" :)
+deep-equal((10, 20, 30), (10, 30, 20)),
+deep-equal((10, 20), (10, 20, 30)),
+deep-equal((10, 20, "a"), (10, 20, "a"))
```
ghislainfourny: You also need to test () vs. a non-empty sequence, both ways.

PR #126, opened 2018-11-24
Fix full path problem in Antlr files

CanBerker:
Full path problem with the "// Generated for ....." line has been fixed in the antlr build file. The problem cause and fix described in comments. Also antlr 4.5.3 is added to lib folder for easy version switching.

There is one additional issue however. When I compile the files // java header and // package .... lines get an unnecessary leading tab and I don't know the cause of this. This is not the case for the files created by @ghislainfourny . I'm open to any ideas what might be causing this. 

As a solution, I can use git filters and sed as following to remove the said tabs:
**in .gitattributes**
src/main/java/sparksoniq/jsoniq/compiler/parser/*.java filter=ignore-lines

**in .git/config**
\# clean executes on "add"
\# smudge executes on "checkout"
[filter "ignore-lines"]
        clean =  sed 's/^\\s.*Java header$/\\/\\/ Java header/' | "sed 's/^\\s.*package sparksoniq.jsoniq.compiler.parser\\;$/package sparksoniq.jsoniq.compiler.parser\\;/'" #  delete leadings tabs for the said two lines
        smudge = cat

\# removed from clean as problem is fixed: sed 's/^.*Generated from.*$/\\\n/' \# remove line starting with "Generated from"


ghislainfourny:
Very good. The tabs may have to do with the ant version. We need to pick one and stick to it -- 4.6 was a bit "unideal" because it is not packaged with any Spark version. We could either pick 4.5.3 or 4.7.

CanBerker:
I have tried it with 4.5.3 and 4.7.1 too and all were leaving tabs for an unknown reason. I think I'll apply a local fix. Thanks for the feedback.

Issue #129, opened 2018-11-26
Add FORG0006 Exception


CanBerker:
Add the exception type and use it for max, min, sum, avg and boolean(*) aggregate functions.

(*) boolean is not exposed at the moment. It will be handled in conjunction with issue #61

PR #133, opened 2018-11-28
Fix aggregate error codes

CanBerker:
Define FORG0006.  Addresses #129 
Fix exceptions of sum,max,min,avg.
Improve the implementation of sum to support a second parameter: $zero
https://www.w3.org/TR/xpath-functions-30/#func-sum


CanBerker:
Merged to the wrong branch, fixed by #142

Issue #11, opened 2018-01-11
Support non-Spark-enabled FLWOR expressions


ghislainfourny:
Allow for "smaller" FLWOR expressions executed locally.

Now that we have refactored the iterators, we need to:

- Convert the ReturnClauseSparkIterator to a HybridIterator and add the local iteration functions (reading through the local iteration API of the child clause iterator).

- Add support for local iteration (also reading via the local iteration API of the child clause iterator) to all other clauses (suggested order: first with let, then for, then when, then group by, then order by). For them, we may or may not use a hybrid iterator model (mostly for code consistency), but this doesn't change anything because a potential materialization will always happen at the return clause and never in the middle of a FLWOR expression.

- For a let clause, isRDD() returns always false if it is the first clause, otherwise it forwards isRDD() from its child iterator. A let iterator *always* materializes its expression's result sequence (another issue exists to tackle further push-downs).

- For a for clause, isRDD() returns true if the expression's iterator's isRDD() returns true OR if the child iterator exists and its isRDD() call returns true

- for all other clauses (including the return item iterator), isRDD() forwards the results from the child's isRDD() call.

ghislainfourny:
let $stores :=
[
  { "store number" : 1, "state" : "MA" },
  { "store number" : 2, "state" : "MA" },
  { "store number" : 3, "state" : "CA" },
  { "store number" : 4, "state" : "CA" }
]
let $sales := [
   { "product" : "broiler", "store number" : 1, "quantity" : 20  },
   { "product" : "toaster", "store number" : 2, "quantity" : 100 },
   { "product" : "toaster", "store number" : 2, "quantity" : 50 },
   { "product" : "toaster", "store number" : 3, "quantity" : 50 },
   { "product" : "blender", "store number" : 3, "quantity" : 100 },
   { "product" : "blender", "store number" : 3, "quantity" : 150 },
   { "product" : "socks", "store number" : 1, "quantity" : 500 },
   { "product" : "socks", "store number" : 2, "quantity" : 10 },
   { "product" : "shirt", "store number" : 3, "quantity" : 10 }
]
let $join :=
  for $store in $stores[], $sale in $sales[]
  where $store."store number" = $sale."store number"
  return {
    "nb" : $store."store number",
    "state" : $store.state,
    "sold" : $sale.product
  }
return [$join]

CanBerker:
I believe this issue is now fully solved.

PR #138, opened 2018-11-29
Define TupleIterators, move files over

CanBerker:
Tuple class moved to the same level as item under "jsoniq/"
tuple iterator folder added in the same level as iterator package(*) under "jsoniq/runtime/" containing the following:
- RuntimeTupleIterator
- RuntimeTupleIteratorInterface (**)
- SparkRuntimeTupleIterator 

(*) : I think iterator package can be renamed as itemiterator, but I didn't do it yet to not bloat up the PR.
(**) : Function defined from line 126 onwards can probably be removed as Tuples don't have any subtypes.

@ghislainfourny This PR initiates the work on local FLWORs defined at #11.


ghislainfourny:
That's a beautiful PR :-) You can go ahead and merge.

Issue #137, opened 2018-11-29
Issue warning whenever a collect() action is performed, truncating the input.


ghislainfourny:
This means:
1. In the shell
and
2. Within the query (which is problematic because it changes the end results silently)

ghislainfourny:
Workaround: the initial for clause should only bind to the json-file() expression, with no further processing. Further processing should be done in subsequent clauses.

    for $i in json-file(...)
    for $j in $i.foo.bar[].foobar[[1]]
    ...

and NOT (yet)

    json-file(...).foo.bar[].foobar[[1]]

PR #139, opened 2018-11-29
Add truncation warning

CanBerker:
A truncation warning has been placed. Addresses issue #137

### DIFF
--- a/src/main/java/sparksoniq/spark/iterator/SparkRuntimeIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/SparkRuntimeIterator.java
@@ -64,11 +64,10 @@ public abstract class SparkRuntimeIterator extends RuntimeIterator {
 
         if(result == null){
             currentResultIndex = 0;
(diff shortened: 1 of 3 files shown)


src/main/java/sparksoniq/spark/iterator/SparkRuntimeIterator.java
```diff
@@ -65,6 +66,7 @@ public Item next(){
             this._rdd = this.getRDD();
             if(SparkContextManager.LIMIT_COLLECT()) {
                 result = _rdd.take(SparkContextManager.COLLECT_ITEM_LIMIT);
+                ShellStart.terminal.output("\nWarning: Results have been truncated to: " + SparkContextManager.COLLECT_ITEM_LIMIT + " items. This value can be configured with the --result-size parameter at startup.\n");
```
ghislainfourny: The warning should only be printed if the limit is reached, meaning, we should surround this output call with a condition on the size of the result list, and only output if the size of the result list is equal to COLLECT_ITEM_LIMIT.

ghislainfourny:
Looks really good! Just a minor comment.

Issue #140, opened 2018-11-30
Filtering using predicates ([ ]) always only returns the last matching item


rschoening:
Example:
`{ "digits" : (0 to 3)[$$ gt 1] }`
returns 
`{ "digits" : 3 }`

expected: 
`{ "digits" : [ 2, 3 ] }`

regardless of predicate used (also for `[true]`).
Same behavior when applied to file ( `json-file("...")[true]`)

Workaround: rewrite as FLWOR/`where` expression

ghislainfourny:
Many thanks for raising this. Indeed, this is a bug. We will address it.

PR #145, opened 2018-12-12
Fix 'only last element returned' bug

CanBerker:
Issue #140 addressed. Test cases added


src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateIterator.java
```diff
@@ -101,6 +101,7 @@ public void setNextResult() {
                 break;
             } else if (Item.getEffectiveBooleanValue(fil)) {
                 _nextResult = item;
+                break;
```
ghislainfourny: These is a really cool PR.

PR #146, opened 2018-12-12
Testing bug

CanBerker:
Bug detected with testing suite. Fixed by changing how unorderedItemSequenceStringsAreEqual is called in RuntimeTests.java

This enabled us to detect further bugs that were previously going unnoticed. All are fixed not and existing test cases are not checked properly.


wscsprint3r:
Good catch!

PR #147, opened 2018-12-12
Add support for object lookup pushdown.

ghislainfourny:
This is a pushdown of object lookup, to provide an example.


ghislainfourny:
@wscsprint3r This can now also be reviewed.

ghislainfourny:
Thanks a lot, @wscsprint3r ! I am leaving that one open for @CanBerker to see as well, so that the same design can be reused with other iterators that we push down. Then I'll merge and we can make a new release.

Happy new year!

ghislainfourny:
@CanBerker Can we now merge this branch, since you have now already coded a similar class with the return clause? Thanks!

PR #148, opened 2018-12-12
Add dynamic context to getRDD() method.

ghislainfourny:


ghislainfourny:
> I just wanted to point out one thing: in RuntimeTupleIterator and SparkRuntimeTupleIterator getRDD() methods do not have the DynamicContext update. Will we handle these later?

Yes. We are doing this in small steps, one at a time. No need to worry about it just yet. You should let the current way (with setDynamicContext) untouched for now.

PR #149, opened 2018-12-12
Extend RuntimeTupleIter by FlworClauseSparkIter

ghislainfourny:


src/main/java/sparksoniq/semantics/visitor/RuntimeIteratorVisitor.java
```diff
@@ -81,34 +82,36 @@ public RuntimeIterator visitDescendants(ExpressionOrClause expression, RuntimeIt
...
+        List<RuntimeTupleIterator> iterators = new ArrayList<>();
+        RuntimeTupleIterator previous = null;
+        previous = this.visitFlowrClause(startClause, argument, previous);
+        iterators.add(previous);
         for (FlworClause clause : expression.get_contentClauses())
-            iterators.addAll(this.visitFlowrClause(clause, argument));
-        ReturnClauseSparkIterator returnIterator =
-                (ReturnClauseSparkIterator) (this.visitFlowrClause(expression.get_returnClause(), argument).get(0));
+            previous = this.visitFlowrClause(clause, argument, previous);
+            iterators.add(previous);
```
ghislainfourny: Isn't a pair of curly braces missing around these two lines? (I think this list may actually be unused and will be dropped later on, but for the sake of correctness, we may want to be careful).

ghislainfourny:
@CanBerker I can't approved because I started the PR --  but this is awesome work. Please go ahead and merge.

PR #150, opened 2018-12-13
Implement array unboxing pushdown.

ghislainfourny:


ghislainfourny:
This allows applying [] to very big sequences of arrays.

Example of query newly supported (assuming object lookup pushdown as well, in separate PR):

count(json-file("file.json").foo[].bar)

The above will all be done in Spark, and the count directly returned to the client even with billions of objects.

Right now, calling .foo[].bar materializes locally and is limited to --result-size, leading to a wrong output.

src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/ArrayUnboxingIterator.java
```diff
@@ -32,6 +32,9 @@
 import java.util.LinkedList;
 import java.util.Queue;
 
+import org.apache.spark.api.java.JavaRDD;
+import org.apache.spark.api.java.function.FlatMapFunction;
+
 public class ArrayUnboxingIterator extends LocalRuntimeIterator {
```
wscsprint3r: This should not be LocalRuntimeIterator

ghislainfourny:
Very good catch!

PR #151, opened 2018-12-14
Testing improvement

CanBerker:
The expected output vs. actual output comparison is made much more restrictive. It now compares the two strings directly without preprocessing spaces and parenthesis (although this can be added if demanded)

Also the output has been formatted to not contain an initial space character. 

Test cases updated to work accordingly. Some test cases which were found to be broken have been fixed.


wscsprint3r:
Great improvement for tests!

CanBerker:
I have also added the condition for sequences in comparison to be the equal size. If they are, unordered item matches can be checked for queries such as groupBy.

PR #153, opened 2018-12-19
Push down filters.

ghislainfourny:
... the fourth PR of the Christmas release. Now we are pushing down everything related to JSON navigation.

e.g.

count(json-file(".../file.json").foo[][$$.bar gt 5][[1]])


src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateIterator.java
```diff
@@ -19,23 +19,22 @@
...
-import sparksoniq.semantics.types.ItemTypes;
 
 import java.util.ArrayList;
 import java.util.List;
 
-public class PredicateIterator extends LocalRuntimeIterator {
+import org.apache.spark.api.java.JavaRDD;
+import org.apache.spark.api.java.function.Function;
+
+public class PredicateIterator extends RuntimeIterator {
```
wscsprint3r: I may not be up to date with all the changes but shouldn't this be a SparkRuntimeIterator?

wscsprint3r: Or it can work both ways now actually. We should probably introduce a new base class for this case.

ghislainfourny: > Or it can work both ways now actually. We should probably introduce a new base class for this case.

How about SparkOrLocalIterator?

ghislainfourny:
@wscsprint3r Actually I thought more about this and you got an excellent point: it should definitely be a SparkRuntimeIterator.

The reason is that a query like

    [ parallelize(1 to 100000)[$$ eq 10000] ]

    parallelize(1 to 100000)[$$ eq 10000] + 1

... didn't work, because [ ... ] was not making use of the new RDD code, materializing the whole sequence before the filter, taking only the first 100 and returning [] or (), causing the warning even though only one item would survive eventually ([10000] and 100001 respectively).

I have now fixed it, i.e., if it is detected that an iterator is an RDD, then open()/next()/... uses the SparkRuntimeIterator base class code to push it down recursively (with a warning issued if it is too big). Otherwise, it is executed locally. I have added one of the above queries as a test.

wscsprint3r:
@ghislainfourny I was about to email you about it, I was confused about the missing next() functionality.

src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateIterator.java
```diff
@@ -50,6 +50,9 @@ public PredicateIterator(RuntimeIterator sequence, RuntimeIterator filterExpress
 
     @Override
     public Item next() {
+        if(isRDD())
```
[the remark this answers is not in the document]
ghislainfourny: Good point. Let me think about it.

ghislainfourny:
@wscsprint3r And there you go. Now there is a clean base class and all the switch machinery is performed by that base class. This makes the derived class much cleaner now. If you approved, I will adapt the other branches as well after the merge.

src/main/java/sparksoniq/jsoniq/runtime/iterator/SparkRuntimeIterator.java
```diff
@@ -58,10 +58,7 @@ public void close(){
     }
 
     @Override
-    public Item next(){
-        if(!this._isOpen)
-            throw new IteratorFlowException("Runtime iterator is not open", getMetadata());
-
+    public boolean hasNext(){
```
ghislainfourny: Note that this code was copied over to a new PR, and tests were added there to make sure there is no regression on this later on.

src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
```diff
@@ -0,0 +1,143 @@
...
+        return _isRDD;
+    }
+
+    @Override
+    public void open(DynamicContext context){
+        super.open(context);
+        if(!_isRDD)
+        {
+            openLocal(context);
+            return;
```
wscsprint3r: Maybe remove this

wscsprint3r:
One small comment, otherwise LGTM

ghislainfourny:
Thanks a lot, Stefan!

PR #159, opened 2019-01-02
Remove all import warnings - v2

CanBerker:
Imports optimized by IntelliJ. Diff is quite big as it changes the ordering of imports in many cases.

This PR makes #154 obsolete. 

Edit: I'm not sure if this covers the case you were aiming for. We can proceed with the Eclipse cleanup if you wish.


ghislainfourny:
Dear Can,

Thanks a lot for taking the initiative! This is much appreciated.

Yes, I think it takes care of it -- this was about superfluous warnings. Since you are using IntelliJ it makes sense to use its own tidying up mechanism.

CanBerker:
To prevent unnecessary conflict handling, import optimization will be done after current PR's are merged.

PR #161, opened 2019-01-03
Remove unneeded wrapper FlworExpressionSparkRuntimeIterator

ghislainfourny:


ghislainfourny:
Now that we have cleaned up the iterator classes and unified them to a single tree of iterators, the wrapper class becomes superfluous as the return iterator now is an item iterator.

ghislainfourny:
Thanks for reviewing the other PR, @CanBerker. This is now the next step.

ghislainfourny:
Thanks :)

PR #162, opened 2019-01-03
Extend HybridIterator in ReturnClause

CanBerker:
Return clause now extends the HybridRuntimeIterator.

@ghislainfourny I've updated the HybridRuntimeIterator line 56 to make a call to isRDD rather than a variable lookup. Before the update, the following query was evaluated fully locally, hence producing errors:
min(for $o in json-file("./src/main/resources/queries/conf-ex.json") return $o.date)
This was caused by the execution never reaching the isRDD method.

I have filled implementations of local calls, however if I'm not mistaken, I believe there are no tests we can perform on this yet (before implementing let clause for example). For this reason I'm not sure if my implementations are sound at this point.

Note: This PR is implemented on top of the existing PR #161  to reduce diff size.

### DIFF
--- a/src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
@@ -19,24 +19,35 @@
  */
  package sparksoniq.spark.iterator.flowr;
 
+import org.apache.spark.api.java.JavaRDD;
+import sparksoniq.exceptions.IteratorFlowException;
 import sparksoniq.jsoniq.item.Item;
+import sparksoniq.jsoniq.runtime.iterator.HybridRuntimeIterator;
 import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
-import sparksoniq.jsoniq.runtime.iterator.SparkRuntimeIterator;
 import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
 import sparksoniq.jsoniq.runtime.tupleiterator.RuntimeTupleIterator;
-import sparksoniq.jsoniq.runtime.tupleiterator.SparkRuntimeTupleIterator;
 import sparksoniq.semantics.DynamicContext;
 import sparksoniq.spark.closures.ReturnFlatMapClosure;
 import sparksoniq.spark.iterator.flowr.base.FlowrClauseSparkIterator;
 
 import java.util.Arrays;
 
-import org.apache.spark.api.java.JavaRDD;
+public class ReturnClauseSparkIterator extends HybridRuntimeIterator {
 
-public class ReturnClauseSparkIterator extends SparkRuntimeIterator {
+    private JavaRDD<Item> itemRDD;
+    private RuntimeTupleIterator _child;
+    private RuntimeIterator _iterator;
+    private Item _nextLocalResult;
+
     public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator expression, IteratorMetadata iteratorMetadata) {
         super(Arrays.asList(expression), iteratorMetadata);
         _child = child;
+        _iterator = expression;
+    }
+
+    @Override
+    protected boolean initIsRDD() {
+        return _child.isRDD();
     }
 
     @Override
@@ -49,8 +60,53 @@ public class ReturnClauseSparkIterator extends SparkRuntimeIterator {
         return itemRDD;
     }
 
-    private JavaRDD<Item> itemRDD;
-    
-    private RuntimeTupleIterator _child;
+    @Override
+    protected boolean hasNextLocal() {
+        return _hasNext;
+    }
+
+    @Override
+    protected Item nextLocal() {
+        if(_hasNext == true){
+            Item result = _nextLocalResult;  // save the result to be returned
+            setNextLocalResult();            // calculate and store the next result
+            return result;
+        }
+        throw new IteratorFlowException("Invalid next() call in Object Lookup", getMetadata());
+    }
+
+    @Override
+    protected void openLocal(DynamicContext context) {
+        this._currentDynamicContext = context;
+
+        _iterator.open(context);
+        setNextLocalResult();
+    }
+
+    private void setNextLocalResult() {
+        if (_iterator.hasNext()) {
+            _nextLocalResult = _iterator.next();
+            this._hasNext = true;
+        } else {
+            _iterator.close();
+            this._hasNext = false;
+        }
+    }
+
+    @Override
+    protected void closeLocal() {
+        _iterator.close();
+    }
+
+    @Override
+    protected void resetLocal(DynamicContext context) {
+        _iterator.reset(_currentDynamicContext);
+        setNextLocalResult();
+    }
+
+
+
+
+
 
 }
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
@@ -53,7 +53,7 @@ public abstract class HybridRuntimeIterator extends RuntimeIterator {
     @Override
     public void open(DynamicContext context){
         super.open(context);
-        if(!_isRDD)
+        if(!isRDD())
         {
             openLocal(context);
         }


src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +60,53 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+            Item result = _nextLocalResult;  // save the result to be returned
+            setNextLocalResult();            // calculate and store the next result
+            return result;
+        }
+        throw new IteratorFlowException("Invalid next() call in Object Lookup", getMetadata());
+    }
+
+    @Override
+    protected void openLocal(DynamicContext context) {
+        this._currentDynamicContext = context;
```
ghislainfourny: I think that this is already done by the superclass that calls openLocal() as part of its open() implementation.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +60,53 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+
+        _iterator.open(context);
+        setNextLocalResult();
+    }
+
+    private void setNextLocalResult() {
+        if (_iterator.hasNext()) {
```
ghislainfourny: This is actually slight more complex (but still relatively simple).

You need to:
1 iterate on the _child iterator (the clause above), getting Tuples
2 for each Tuple obtained above, chain the _currentDynamicContext with the dynamic context obtained from the Tuple (something like that is done in the filter expression implementation, for example).
3 open the _iterator iterator, passing the above chained dynamic context
4 iterate on the _iterator iterator until hasNext() is false, then go back to 2 and continue.

Note that nothing needs to be materialized and this is just a bunch of if-then-elses:

For every call to setNextLocalResult(), you simply check whether _iterator is initialized and has more items.

If it does, you just call next() and assign it to _nextLocalResult.

If it doesn't, then you check whether_child has more tuples (if it is not initialized, you do so).

If it does, you call next(), get the Tuple, initialize a new dynamic context, reset _iterator with that context and assign the first item to _nextLocalResult.

If _child has no more tuples, then this is the end and _hasNext becomes false.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +60,53 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+            return result;
+        }
+        throw new IteratorFlowException("Invalid next() call in Object Lookup", getMetadata());
+    }
+
+    @Override
+    protected void openLocal(DynamicContext context) {
+        this._currentDynamicContext = context;
+
+        _iterator.open(context);
```
ghislainfourny: I think you shouldn't open _iterator here, but in setNextLocalResult. This is because there will be one dynamic context passed to it for every tuple in the child clause :-) (see explanations below)

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
CanBerker: Yes you're absolutely right. However, this left me wondering why we're passing dynamic context as a parameter to openLocal in this case. The context is already initialized and this seems redundant. Shall we remove the parameter from the function signature in HybridIterator?

ghislainfourny: Yes, exactly! I realized this as well. Can you open a separate issue for this? We can take care of it later.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
CanBerker: Thanks a lot for the detailed feedback, it was very helpful. I have just pushed an implementation which I believe covers most of the points of execution you have listed. I'm not sure if I have got the dynamic context chaining right, however.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +61,66 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+    protected void openLocal(DynamicContext context) {
+        this._currentDynamicContext = context;
+
+        _child.open(context);
+        setNextLocalResult();
+    }
+
+    private void setNextLocalResult() {
+        _nextLocalResult = null;
+        if (_expression.hasNext()) {
```
ghislainfourny: Here we have to add a boolean field to know whether _expression was opened yet or not. The first time, the control flow must first go to the else close to make sure _expression() gets opened with the first dynamic context of the tuple stream.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +61,66 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+            _nextLocalResult = _expression.next();
+        } else {
+            _expression.close();
+
+            if (_child.hasNext()) {
+                FlworTuple tuple = _child.next();
+                DynamicContext dynamicContext = new DynamicContext(tuple);
```
ghislainfourny: You need to chain it behind _currentDynamicContext (I think there is another constructor for that somewhere) so that all bindings from the outer context are correctly inherited.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +61,66 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+                _expression.open(dynamicContext);
+                _nextLocalResult = _expression.next();
+            } else {
+                _child.close();
+                this._hasNext = false;
+            }
+        }
+
+        if (_nextLocalResult != null) {
```
ghislainfourny: I think it would be cleaner to set _hasNext to true always at the same time as _nextLocalResult is set, and end each branch with a return; statement to exit the method. This makes the code easier to read.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +61,66 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+            this._hasNext = true;
+        }
+    }
+
+    @Override
+    protected void closeLocal() {
+        _child.close();
```
ghislainfourny: You may need to close _expression as well in case it is open to keep the state consistent (see this new boolean field mentioned above).

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -49,8 +61,66 @@ public ReturnClauseSparkIterator(RuntimeTupleIterator child, RuntimeIterator exp
...
+    }
+
+    @Override
+    protected void resetLocal(DynamicContext context) {
+        _child.reset(_currentDynamicContext);
+        setNextLocalResult();
```
ghislainfourny: Here also you need to close _expression to get back to a consistent state.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
CanBerker: Given that _expression's state is consistent the hasNext should return false for non-opened iterators. So the control flow will indeed always go to the else first for the first iteration. However, if this feels hacky*, I can add the boolean solution as well.

*: in the first iteration close() is called on a not opened iterator (_expression), however this call has no side effects in the current implementation.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
CanBerker: I have found earlier that there is a mechanism to connect dynamic contexts to each other through a parent field relationship. However in the current implementation this is done through a constructor (no setter). 

Additionally, it is also possible to create a dynamic context from a tuple which is also done through a constructor which calls a private method for the operation.

If we want to connect dynamic contexts with the parent field we either need a setter or expose the tuple -> dynamic context method to outside. I can implement your selection, in this or in a different PR based on your feedback.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
CanBerker: I was stuck between that choice and the current implementation :) I'll make the change right away

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
ghislainfourny: I think we simply need to add a constructor to the DynamicContext class:

    public DynamicContext(DynamicContext parent, FlworTuple tuple){
            this._parent = parent;
            this._variableValues = new HashMap<>();
            generateDynamicContextFromTupleRdd(tuple);
        }

So far there was only one version to connect to the parent, and one to convert a tuple, but none that did both at the same time.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
ghislainfourny: Yes -- it is cleaner to not "abuse" hasNext() to test for openness. In the future, we may even want to add an isOpen() method to the iterators that takes care of that boolean, but for now an explicit boolean is fine.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
ghislainfourny: For the record: it is also a viable option, as you suggest, to expose the conversion method to the outside world (maybe calling it setBindingsFromTuple()), because this would avoid constructing a new DynamicContext object for every tuple (performance overhead as object construction is quite expensive): you can created a single DynamicContext object, chained with its parent, and set it in place by calling setBindingsFromTuple for each tuple in turn.

So I am fine either way and leave the decision to you. At some point we will have to profile the code and improve performance anyway.

PR #172, opened 2019-01-10
Local let clause implementation

CanBerker:
I have partially implemented the local let clause support. The following query can be evaluated fully locally.
let $o := {"foobar": 2}
return $o

However, I have ran into an issue that's currently blocking me. Running the following query
let $o := {"foobar": 2}
let $a := ("foo" || "bar")
return $o.$a
causes the following: [ERROR] Error [err: XPDY0130 ] Runtime error retrieving variable $o value

I was wondering whether this was caused by an issue in context chaining or not passing the assignment from the "first let"s tuple(s*) into the "second let". 

Additionally, while looking into the issue of possibly faulty context chaining, I found the "chaining first and opening the child later" much more sensible and made such changes in my last commit. I'd appreciate the feedback on both this change and into the blocking issue very much.

Lastly, I have a very basic question to brush up on my understanding as well as leading up to the issue with the materialization of RDD's for the initial let:
*: Does the let clause always create a single tuple(first let) / preserve the number of tuples (subsequent lets) ?
In the light of this, how should we handle the materialization of an RDD for an initial let clause? Should it be handled as a single tuple or should it be somehow divided into tuples?

### DIFF
--- a/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
@@ -107,9 +107,9 @@ public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
         // evaluate locally
         if (!isRDD()) {
             if (this._child != null) { //if it's not a start clause
-                _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
-                _child.open(_tupleContext);
+                _child.open(_currentDynamicContext);
                 _variableName = _variableReference.getVariableName();
+                _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
 
                 setNextLocalTupleResult();
 
@@ -118,9 +118,8 @@ public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
                     throw new SparksoniqRuntimeException("Initial let clauses don't support RDDs");
                     // materialize the initial RDD
                 else {
-                    _tupleContext = new DynamicContext(_currentDynamicContext);
                     List<Item> contents = new ArrayList<>();
-                    _expression.open(_tupleContext);
+                    _expression.open(this._currentDynamicContext);
                     while (_expression.hasNext())
                         contents.add(_expression.next());
                     _expression.close();
--- a/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
@@ -107,9 +107,9 @@ public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
         // evaluate locally
         if (!isRDD()) {
             if (this._child != null) { //if it's not a start clause
-                _child.open(_currentDynamicContext);
-                _variableName = _variableReference.getVariableName();
                 _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
+                _child.open(_tupleContext);
+                _variableName = _variableReference.getVariableName();
 
                 setNextLocalTupleResult();
 
@@ -118,8 +118,9 @@ public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
                     throw new SparksoniqRuntimeException("Initial let clauses don't support RDDs");
                     // materialize the initial RDD
                 else {
+                    _tupleContext = new DynamicContext(_currentDynamicContext);
                     List<Item> contents = new ArrayList<>();
-                    _expression.open(this._currentDynamicContext);
+                    _expression.open(_tupleContext);
                     while (_expression.hasNext())
                         contents.add(_expression.next());
                     _expression.close();
--- a/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
@@ -103,29 +103,24 @@ public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
     public void open(DynamicContext context) {
         super.open(context);
 
-        // evaluate locally
-        if (!isRDD()) {
-            if (this._child != null) { //if it's not a start clause
-                _child.open(_currentDynamicContext);
-                _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
+        // isRDD checks omitted, as open is used for non-RDD(local) operations
 
-                setNextLocalTupleResult();
+        if (this._child != null) { //if it's not a start clause
+            _child.open(_currentDynamicContext);
+            _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
 
-            } else {    //if it's a start clause, it returns only one tuple
-                if(_expression.isRDD())
-                    throw new SparksoniqRuntimeException("Initial let clauses don't support RDDs");
-                    // materialize the initial RDD
-                else {
-                    List<Item> contents = new ArrayList<>();
-                    _expression.open(this._currentDynamicContext);
-                    while (_expression.hasNext())
-                        contents.add(_expression.next());
-                    _expression.close();
-                    FlworTuple tuple  = new FlworTuple();
-                    tuple.putValue(_variableName, contents, false);
-                    _nextLocalTupleResult = tuple;
-                }
-            }
+            setNextLocalTupleResult();
+
+        } else {    //if it's a start clause, it returns only one tuple
+            // expression is materialized
+            List<Item> contents = new ArrayList<>();
+            _expression.open(this._currentDynamicContext);
+            while (_expression.hasNext())
+                contents.add(_expression.next());
+            _expression.close();
+            FlworTuple tuple  = new FlworTuple();
+            tuple.putValue(_variableName, contents, false);
+            _nextLocalTupleResult = tuple;
         }
     }
 
--- a/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
@@ -91,9 +91,8 @@ public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
                 _expressionResults.add(_expression.next());
             _expression.close();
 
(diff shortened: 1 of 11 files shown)


src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
 import java.util.List;
 
 public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
+
+    private VariableReferenceIterator _variableReference;
```
ghislainfourny: I think we don't even need to store a VariableReferenceIterator anywhere. It is an overkill to pass an iterator just to tell the name of the let variable. You can already resolve this to a String in the constructor. In the future, we can even pass String variableName instead to the constructor (but no need to do it now).

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
...
+            this._hasNext = false;
+        }
+    }
+
+    @Override
+    public void open(DynamicContext context) {
+        super.open(context);
+
+        // evaluate locally
+        if (!isRDD()) {
```
ghislainfourny: I think the check for RDDs is superfluous for local execution of FLWOR clauses. If the caller calls iterator methods, then we just proceed with the local execution. It will anyway never happen that the caller calls iterator methods if it knows it supports RDDs.

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
...
+            if (this._child != null) { //if it's not a start clause
+                _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
+                _child.open(_tupleContext);
+                _variableName = _variableReference.getVariableName();
+
+                setNextLocalTupleResult();
+
+            } else {    //if it's a start clause, it returns only one tuple
+                if(_expression.isRDD())
```
ghislainfourny: I think this check should be removed. Even if the expression is an RDD, the code below (in the else clause) anyway will trigger a materialization with no extra coding efforts, which is exactly what we want (this is how next() is implemented already for hybrid/spark item iterators in case it is an RDD). This is the benefit of all the design work done earlier :-)

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
...
+                    // materialize the initial RDD
+                else {
+                    _tupleContext = new DynamicContext(_currentDynamicContext);
+                    List<Item> contents = new ArrayList<>();
+                    _expression.open(_tupleContext);
+                    while (_expression.hasNext())
+                        contents.add(_expression.next());
+                    _expression.close();
+                    FlworTuple tuple  = new FlworTuple();
+                    tuple.putValue(_variableReference.getVariableName(), contents, false);
```
ghislainfourny: Use _variableName directly?

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
...
+            _tupleContext.removeAllVariables();             // clear the previous variables
+            _tupleContext.setBindingsFromTuple(inputTuple);      // assign new variables from new tuple
+
+            List<Item> result = new ArrayList<>();
+            _expression.open(_tupleContext);
+            while (_expression.hasNext())
+                result.add(_expression.next());
+            _expression.close();
+
+            FlworTuple resultTuple = new FlworTuple();
```
ghislainfourny: The output tuple should also contain all bindings of the input tuple. The let clause merely adds a binding to all input tuples. I am not sure if these input bindings need to be added by hand or if the FlworTuple class has a constructor that copies from an existing FlworTuple (you can look at the closure function used in Spark, which already takes care of the same thing).

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
...
+            this._hasNext = false;
+            return;
+        }
+
+        if (_child.hasNext()) {
+            FlworTuple inputTuple = _child.next();
+            _tupleContext.removeAllVariables();             // clear the previous variables
+            _tupleContext.setBindingsFromTuple(inputTuple);      // assign new variables from new tuple
+
+            List<Item> result = new ArrayList<>();
```
ghislainfourny: For performance, this should be moved to a field in the class. That way, the object is only created once in the LetClauseSparkIterator constructor, and, at this position of the code, you just need to empty it. This saves all the object creations, which are very expensive.

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
...
+    @Override
+    public void open(DynamicContext context) {
+        super.open(context);
+
+        // evaluate locally
+        if (!isRDD()) {
+            if (this._child != null) { //if it's not a start clause
+                _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
+                _child.open(_tupleContext);
+                _variableName = _variableReference.getVariableName();
```
ghislainfourny: This only needs to be done once in the constructor of the class.

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
...
+
+    @Override
+    public void open(DynamicContext context) {
+        super.open(context);
+
+        // evaluate locally
+        if (!isRDD()) {
+            if (this._child != null) { //if it's not a start clause
+                _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
+                _child.open(_tupleContext);
```
ghislainfourny: For consistency (even if it leads to the same result), I suggest

_child.open(_currentDynamicContext).

_tupleContext is meant to be the dynamic context that you update in place, for use on the child expression.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -80,9 +80,9 @@ protected Item nextLocal() {
 
     @Override
     protected void openLocal(DynamicContext context) {
-        _child.open(context);
-        _isExpressionOpen = false;
         _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
+        _child.open(_tupleContext);
```
ghislainfourny: here too, _child.open(context) was fine. _tupleContext is meant to be the context updated in place for use on _expression.

ghislainfourny:
@CanBerker This is a great start.

let $o := {"foobar": 2}
let $a := ("foo" || "bar")
return $o.$a

is failing simply because in the existing code, the output tuple of a let clause (here the second one) does not forward the bindings from the input tuples ($o). See my comment.

ghislainfourny:
> I found the "chaining first and opening the child later" much more sensible and made such changes in my last commit. I'd appreciate the feedback on both this change and into the blocking issue very much.

The child tuple can simply be open with the exact same dynamic context as that passed to the let clause. 
The chaining is only meant for iterating on the child expression, because it is executed in a context where the external context is chained with the input tuples.

So, for the child tuple, no need to overcomplicate things: the context can be passed as is.

ghislainfourny:
>  Does the let clause always create a single tuple(first let) / preserve the number of tuples (subsequent lets) ?

Yes.

ghislainfourny:
> In the light of this, how should we handle the materialization of an RDD for an initial let clause?

Nothing to do here. If the expression is an RDD, it will be materialized automatically by the hasNext()/next() calls. You don't even need to bother checking.

ghislainfourny:
Thanks a lot, @CanBerker. I think I have answered all your open questions. These are only minors things to fix. :-)

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -36,25 +38,113 @@
 import java.util.List;
 
 public class LetClauseSparkIterator extends FlowrClauseSparkIterator {
```
ghislainfourny: I think the FlworClauseSparkIterator may no longer be needed because the overriden code remaining there will never be reached. We can directly inherit from RuntimeTupleIterator.

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -72,10 +162,29 @@ public LetClauseSparkIterator(RuntimeTupleIterator child, VariableReferenceItera
...
             rdd = SparkContextManager.getInstance().getContext().parallelize(tuples);
         }
         return rdd;
     }
+
+
+    @Override
+    public boolean hasNext() {
+        return super.hasNext();
+    }
```
ghislainfourny: This is superflous (the override only calls the super function which only adds a performance overhead)

src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -72,10 +162,29 @@ public LetClauseSparkIterator(RuntimeTupleIterator child, VariableReferenceItera
...
+
+    @Override
+    public void reset(DynamicContext context) {
+        super.reset(context);
```
ghislainfourny: This is superflous (the override only calls the super function which only adds a performance overhead)

CanBerker:
Thanks a lot for the detailed feedback. It has covered all the questions I had in mind and more. I have implemented all the changes you have mentioned.

While changing the parent class of letClause I had to move the setDynamicContext method from FlworClauseSparkIterator to SparkRuntimeTupleIterator. Otherwise this was causing type casting problems when letClause started inheriting directly from SparkRuntimeTupleIterator.

ghislainfourny:
Yay! Let us go ahead and merge this. Now let's continue with the for clause.

PR #173, opened 2019-01-11
Update documentation towards 0.9.5

ghislainfourny:


ghislainfourny:
> This looks very good :)
> 
> Edit: I wanted to share a recent experience about lack of support for Spark in Java 9 at the moment. I made a fresh installation and I tried Sparksoniq with Java9. While running RDD operations I ran into "apache.xbean.asm5.ClassReader IllegalArguementException". Apparently, Spark is not supported at the moment by java 9+ as seen in the following Jira issue: https://jira.apache.org/jira/browse/SPARK-24201
> 
> I thought requirements section in Compile from Github page could specify this strict dependency on java 8. If this issue doesn't get resolved by the developers in the upcoming months, I believe students who don't pay attention to the java 8 requirement might run into the same problem.
> 
> Edit2: I can confirm that the following version seems to be fully functional if you'd like to update that part as well:
> Java(TM) SE Runtime Environment (build 1.8.0_202-b08)

Thanks a lot, Can! I added a hint in the documentation.

PR #174, opened 2019-01-16
local for clauses

CanBerker:
In the first group of commits I have implemented the local execution of the for clause. At this state it fully supported local to local clause transitions. 

In the second group of comments local to RDD clause transition for "for clause" has been implemented. I have put in a few simple test cases where local->local, local->RDD, RDD->RDD transitions have been covered. 

It seems that RDD->RDD transition worked out of the box from Stefan's implementation(I've only simplified it a little bit). The simple test case shows that 25 results (5*5 for 2 RDDS) which would make sense for a Cartesian operation. Do we need a different implementation using the Cartesian function (for performance or accuracy reasons) or does the current implementation look correct @ghislainfourny ?

Note: Changes in LetClause and ReturnClause are just some polishing where code duplication is cleaned up and isOpen() function of the iterator is used instead of local boolean.


src/main/java/sparksoniq/spark/iterator/flowr/LetClauseSparkIterator.java
```diff
@@ -82,14 +80,14 @@ private void setNextLocalTupleResult() {
             _tupleContext.removeAllVariables();             // clear the previous variables
             _tupleContext.setBindingsFromTuple(inputTuple);      // assign new variables from new tuple
 
-            _expressionResults.clear();     // clear the results from previous iteration
+            List<Item> results = new ArrayList<>();
```
ghislainfourny: Aha. Good fix and nice catch! I just understood. You need a copy every time.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -81,35 +80,27 @@ protected Item nextLocal() {
...
         _tupleContext = new DynamicContext(_currentDynamicContext);     // assign current context as parent
         setNextLocalResult();
     }
 
     private void setNextLocalResult() {
-        if (_isExpressionOpen) {
-            if (_expression.hasNext()) {
-                _nextLocalResult = _expression.next();
-                this._hasNext = true;
+        if (_expression.isOpen()) {
```
ghislainfourny: Very good.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -20,45 +20,171 @@
...
+        // execution reaches here when there are no more results
+        _child.close();
+        this._hasNext = false;
+    }
+
+    /**
+     * _expression has to be open prior to call.
+     * @return true if _nextLocalTupleResult is set and _hasNext is true, false otherwise
+     */
+    private boolean setResultFromExpression() {
```
ghislainfourny: You can merge these two functions in just one that takes a boolean. This avoids code duplication.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -20,45 +20,171 @@
...
+        }
     }
 
 
     @Override
     public JavaRDD<FlworTuple> getRDD() {
-        if (this._rdd == null) {
-            String variableReference = ((VariableReferenceIterator)this._children.get(0)).getVariableName();
-            RuntimeIterator assignmentExpression = this._children.get(1);
+        if (this._rdd == null) {        //if it's a start clause
```
ghislainfourny: This if should be removed, because the rdd shouldn't be cached. It may change across different outer dynamic context (at some point we will pass it as a parameter to getRDD() as we do for items).

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -20,45 +20,171 @@
...
+                    // get child tuples
+                    List<FlworTuple> localTuples = new ArrayList<>();
+                    _child.open(_currentDynamicContext);
+                    while (_child.hasNext()) {
+                        localTuples.add(_child.next());
+                    }
+                    _child.close();
+
+                    // turn them into an rdd
+                    _rdd = SparkContextManager.getInstance().getContext().parallelize(localTuples);
```
ghislainfourny: Here you are basically materializing the (potentially big) expression and turning a (usually small because not an RDD) set of tuples into an RDD.

I think it can/should be done the other way round, i.e., local iteration on the input tuples, and for each input tuple, ask for the corresponding output tuple RDD (obtained from a map on the expression RDD. This is a new closure class that maps each item I of the expression to a tuple that is a copy of the input tuple, plus a binding from the variable to I, with the tuple constructor you already created). That gives you one RDD per tuple, and then you can call a union on all these RDD to obtain the final output RDD.

The use case is queries such as
for $i in (1000, 10000, 1000000)
for $j in parallelize(1 to $i)
return...

or

for $f in ("file1.json", "file2.json")
for $i in json-file($f)
return ...

Sorry for having not been precise enough yesterday! I think the case we discussed was a special case in which the expression doesn't depend on the input tuple.

ghislainfourny:
> It seems that RDD->RDD transition worked out of the box from Stefan's implementation(I've only simplified it a little bit). The simple test case shows that 25 results (5*5 for 2 RDDS) which would make sense for a Cartesian operation. Do we need a different implementation using the Cartesian function (for performance or accuracy reasons) or does the current implementation look correct

Exactly - No need to special case RDD->RDD. This works out of the box because the second RDD gets materialized automatically. No need to worry about the cartesian product at this point -- I realized this only makes sense when the expression is independent on the input tuple. We should keep such optimization for later.

ghislainfourny:
A few tests:

for $i in
  for $j in 1 to 10
  return $j + 1
return $i + 1


for $i in
  for $j in
    for $k in 1 to 10
    return $k + 1
  return $j + 1
return $i + 1

for $i in parallelize(1 to 10000)
let $j := i + 1
return $i + $j

let $i := 10000
for $j in parallelize(1 to $i)
return $i + $j

for $i in parallelize(1 to 100)
for $j in parallelize(1 to $i)
return $i + 1

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -20,45 +20,171 @@
...
-            //if it's a start clause
+
             if (this._child == null) {
-                initialRdd = this.getNewRDDFromExpression(assignmentExpression);
-                this._rdd = initialRdd.map(new InitialForClauseClosure(variableReference));
-            } else {
-            //if it's not a start clause
-                this._rdd = this._child.getRDD();
-                this._rdd = this._rdd.flatMap(new ForClauseClosure(assignmentExpression, variableReference));
+                initialRdd = this.getNewRDDFromExpression(_expression);
```
ghislainfourny: I think that can even be simplified, because the else case in getNewRDDFromExpression will never happen (it is now handled with the newly implemented local execution). So you can directly ask for the expression RDD here and remove that function.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Should we remove the RDD caching in the getRDD function of letClause as well?

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Yes that's a very good suggestion. However, we have to postpone this until all isRDD methods are overridden in all flwor clauses. Otherwise it causes Iterator is not an RDD exceptions for LocalIterators in the existing test cases. I'll open an issue for later

ghislainfourny: Yes, an issue is fine. It's good practice to keep PRs small.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
ghislainfourny: yes, it applies there too.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: I have just pushed an implementation try at this. The current forClauseClosure creates a single tuple for each result obtained from the expression. The created tuple is a duplicate of the input tuple containing the new variable binding. I believe this is exactly what you meant in your description so I thought the existing implementation can be used here.

In its current state operation is as follows:
1-) read a single child tuple
2-) Turn it into an RDD with parallelize (not sure if this is the best way)
3-) Apply the forclosure as explained above
4-) union all RDDs(different one for each child tuple) at the end

Is there an implementation detail I'm missing/ have misunderstood?

ghislainfourny: Almost :-)

The map shouldn't map a tuple to tuples. It should map an *item* (from the expression RDD) to tuples. This is a new closure class.

Here is a simple example:

for $i in (2, 3)
for $j in parallelize(1 to $i)
return $i + $j

the two input tuples to the second for clause (locally available) are
(i: 2)
(i:3)

what needs to be done is to iterate through these two tuples locally (without creating any RDD from any of them):

- for the tuple (i:2), take the RDD returned by the expression (internally contains 1, 2)

map that rdd of items to a tuple RDD like so:

1 -> mapped to (i:2, j:1) 
2 -> mapped to (i:2, j:2) 

This is a mapping from an RDD of items to an RDD of tuples. It is a closure that has a private variable containing the input tuple T (here T = (i:2)) passed through the constructor.
It takes x and returns (T, j: x)

The output RDD for that input tuple is thus made of:
(i:2, j:1) 
(i:2, j:2) 

- for the tuple (i:3), take the RDD returned by the expression (internally contains 1, 2, 3)

1 -> mapped to (i:3, j:1) 
2 -> mapped to (i:3, j:2) 
2 -> mapped to (i:3, j:2) 

This is the same closure, but with T = (i:3).

The output RDD for that input tuple is thus made of:
(i:3, j:1) 
(i:3, j:2) 
(i:3, j:3)

- finally, you union all output rdds:
 (i:2, j:1) 
(i:2, j:2) 
(i:3, j:1) 
(i:3, j:2) 
(i:3, j:3)

and output this RDD as the output of the second clause.

Does it help?

CanBerker: Thank you so much for the detailed explanation. I realized that I understood it backwards in a sense in the first place. I have pushed another implementation just now. Unfortunately in its current state I keep facing a "task not serializable" exception and I haven't been able to figure out the cause yet. I'll keep looking for the source of the problem. In the mean time, if you catch something that I'm missing please let me know

ghislainfourny: Very good. The task not serializable is issue is, in all likelihood, a class that doesn't bear the "Serializable" interface, in the transitive closure of all the data stored in the closure class fields.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -20,45 +20,162 @@
...
+                while (_child.hasNext()) {
+                    _inputTuple = _child.next();
+                    _tupleContext.removeAllVariables();             // clear the previous variables
+                    _tupleContext.setBindingsFromTuple(_inputTuple);      // assign new variables from new tuple
+
+                    JavaRDD<Item> expressionRDD = _expression.getRDD(_tupleContext);
+                    if (this._rdd == null){
+                        this._rdd = expressionRDD.map(new ForClauseLocalToRDDClosure(_variableName, _inputTuple));
+                    } else {
+                        this._rdd.union(expressionRDD.map(new ForClauseLocalToRDDClosure(_variableName, _inputTuple)));
```
ghislainfourny: This may requiring an assignment to this._rdd

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -20,45 +20,162 @@
...
+                    }
+                }
+                _child.close();
             }
         }
         return _rdd;
```
ghislainfourny: We may need to create an empty RDD in the case that _rdd = null.

src/main/java/sparksoniq/spark/closures/ForClauseLocalToRDDClosure.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.spark.api.java.function.Function;
+import sparksoniq.jsoniq.item.Item;
+import sparksoniq.jsoniq.tuple.FlworTuple;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ForClauseLocalToRDDClosure implements Function<Item, FlworTuple> {
+    private final String _variableName;
+    private final FlworTuple _inputTuple;
```
ghislainfourny: It may be the FlworTuple that needs to be Serializable (it is KryoSerializable for use in RDDs, but not sure if this is enough for passing to a closure)

ghislainfourny:
We are almost there :-) I hope making FlworTuple Serializable will do the trick (which should be as easy as just adding the interface). Then we do a last pass and it should be good to go.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Through debugging I extracted the following exception cause: It seems FlworTuple is causing the issue. However, the class is KryoSerializable.

java.io.NotSerializableException: sparksoniq.jsoniq.tuple.FlworTuple
Serialization stack:
	- object not serializable (class: sparksoniq.jsoniq.tuple.FlworTuple, value: sparksoniq.jsoniq.tuple.FlworTuple@760f1081)
	- field (class: sparksoniq.spark.closures.ForClauseLocalToRDDClosure, name: _inputTuple, type: class sparksoniq.jsoniq.tuple.FlworTuple)
	- object (class sparksoniq.spark.closures.ForClauseLocalToRDDClosure, sparksoniq.spark.closures.ForClauseLocalToRDDClosure@7a358613)
	- field (class: org.apache.spark.api.java.JavaPairRDD$$anonfun$toScalaFunction$1, name: fun$1, type: interface org.apache.spark.api.java.function.Function)
	- object (class org.apache.spark.api.java.JavaPairRDD$$anonfun$toScalaFunction$1, <function1>)

ghislainfourny: Exactly! :-)

CanBerker: Good news! I have managed to work through the serialization issue and the code is functional now. I have also added the case you've mentioned above as a test case.

CanBerker:
We have finally got the new closure and local->RDD transition working. 

Next up, I have noticed some issues while handling the 6 test cases you've given earlier:
https://github.com/Sparksoniq/sparksoniq/pull/174#issuecomment-455177024

1st and the 2nd ones are perfectly fine and are currently active in the testing system.

3rd and the 4th ones are perfect in functionality, however our testing system is struggling performance wise. In the console application, results are fetched from spark and truncated to print only the first 100. The testing system however, fully reads the results with a series of next() calls and without any truncation. I think this needs to be made more efficient. if these tests are to be put in the system (they take 1-2 minutes each).

Lastly, I'm not sure if 5th and 6th test cases are working at the moment. I think the 5th test case (with 2 parallelize statements) halts for some reason. The 6th test case throws the following exception so we're sure that there is a bug there: "sparksoniq.exceptions.IteratorFlowException: Error [err: XPDY0130]LINE:3:COLUMN:0:Runtime tuple iterator cannot be opened twice"

I believe we can try to address these two issues as the next step.

CanBerker:
We have finally got the new closure and local->RDD transition working. 

Next up, I have noticed some issues while handling the 6 test cases you've given earlier:
https://github.com/Sparksoniq/sparksoniq/pull/174#issuecomment-455177024

1st and the 2nd ones are perfectly fine and are currently active in the testing system.

3rd and the 4th ones are perfect in functionality, however our testing system is struggling performance wise. In the console application, results are fetched from spark and truncated to print only the first 100. The testing system however, fully reads the results with a series of next() calls and without any truncation. I think this needs to be made more efficient. if these tests are to be put in the system (they take 1-2 minutes each).

Lastly, I'm not sure if 5th and 6th test cases are working at the moment. I think the 5th test case (with 2 parallelize statements) halts for some reason. The 6th test case throws the following exception so we're sure that there is a bug there: "sparksoniq.exceptions.IteratorFlowException: Error [err: XPDY0130]LINE:3:COLUMN:0:Runtime tuple iterator cannot be opened twice"

I believe we can try to address these two issues as the next step.

ghislainfourny:
Dear Can.

Let's take them one by one.

For query 5, this is what I feared, because this is a job within a job and Spark documents this as not supported. I suggest we make it an error for now if both the child clause (tuple iterator) and the expression (item iterator) are both RDDs. We can optimize special cases later, especially when the expression does not depend on the input tuples.

ghislainfourny:
For tests 3 and 4, can you replace 10000 with 1000 to make it faster, and open an issue to improve performance there? It will definitely be worth investigating, but we can postpone this (correctness first).

ghislainfourny:
> The serialization issue was fixed when I made the FlworTuple and the SerializableItem class serializable. 

Very nice!

ghislainfourny:
For query 6, I think this should be investigated because the query is expected to work. It may be something along the lines of calling open instead of reset when using an iterator several times.

CanBerker:
> For tests 3 and 4, can you replace 10000 with 1000 to make it faster, and open an issue to improve performance there? It will definitely be worth investigating, but we can postpone this (correctness first).

Tests 3 and 4 are currently in the system with range up to 5. Should I increase this number to 1000?

Opening the issue for testing suite performance improvement now.


>For query 6, I think this should be investigated because the query is expected to work. It may be something along the lines of calling open instead of reset when using an iterator several times.

Query 6 is actually functional now :) The test is in place without any issues.

Lastly, I renamed and disabled a few cases which required implementations of local where and groupby clauses. I will reactivate them when they are ready.

ghislainfourny:
> Tests 3 and 4 are currently in the system with range up to 5. Should I increase this number to 1000?

Or to 200 -- above the default result size to test for local execution.

> Query 6 is actually functional now :) The test is in place without any issues.

Perfect!

> Lastly, I renamed and disabled a few cases which required implementations of local where and groupby clauses. I will reactivate them when they are ready.

Also great.

Feel free to merge after having set the number from 5 to 200 or so.

PR #175, opened 2019-01-17
Where clause local

CanBerker:
Local where clauses implemented. Simple test cases added.

Contains commits from local for implementation. It will require review when local for implementation is merged to master.


ghislainfourny:
After the for branch is done and merged, I suggest taking the time to add the dynamic context parameter to getRDD() for tuples, and remove the then obsolete setDynamicContext() function. This will make things cleaner and easier for the remaining clauses.

src/main/java/sparksoniq/spark/iterator/flowr/WhereClauseSparkIterator.java
```diff
@@ -20,30 +20,103 @@
...
+        // for each incoming tuple, evaluate the expression to a boolean.
+        // forward if true, drop if false
+
+        while (_child.hasNext()) {
+            _inputTuple = _child.next();
+            _tupleContext.removeAllVariables();             // clear the previous variables
+            _tupleContext.setBindingsFromTuple(_inputTuple);      // assign new variables from new tuple
+
+            _expression.open(_tupleContext);
+            Item result = _expression.next();
```
ghislainfourny: This is slightly more complicated than that -- you need to compute the Effective Boolean Value of the sequence, as done at other places.

In particular, you need to test for _expression.hasNext() that may be false (empty sequence, in which the EBV is false).

Then you need to test, if _expression.next() is an atomic item, that it is a singleton sequence: if _expression.hasNext() then returns true after next() was an atomic, then an error must be raised. If next() is an object or array, no need to call hasNext() a second time as the EBV is then always true.

It is very likely that this is already implemented at other places and that you do not need to reinvent any new code, for example you can check in the implementation of the IfThenElse iterator, for the computation of the condition.

src/main/java/sparksoniq/spark/iterator/flowr/WhereClauseSparkIterator.java
```diff
@@ -20,30 +20,103 @@
...
+        }
+
+        // execution reaches here when there are no more results
+        _child.close();
+        this._hasNext = false;
     }
 
     @Override
     public JavaRDD<FlworTuple> getRDD() {
         if (this._rdd == null) {
```
ghislainfourny: As discussed before, the RDD should not be cached. But this can also be handled in your other PR in which you change the signature of this function.

src/main/resources/test_files/runtime-spark/LocalClauses/WhereClauseLocal1.iq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="(5, 1, 2, 3, 4, 5)" :)
```
ghislainfourny: Can we add some more corner cases with the empty sequence?

for $i in ()
return $i

-> ()

for $i in ()
where $j := $i + 1
return $i

-> ()

let $i := ()
return 1

let $i := ()
where exists($i)
return 1

-> 1

for $i in 1 to 10
where ()
return $i

-> ()

for $i in 0 to 3
where $i
return $i

-> (1, 2, 3)

for $i in ("", "foo")
where $i
return $i

-> ("foo")

let $i := 1
where ({}, [], null)
return $i

-> 1

let $i := 1
where ([], {}, null)
return $i

-> 1

ghislainfourny:
Looks very good! Just a few more test suggestions, and the fix to make sure the EBV of the where expression is used.

src/main/java/sparksoniq/spark/iterator/flowr/WhereClauseSparkIterator.java
ghislainfourny: Actually I checked the implementation of IfThenElse and it only checks for one item. You may need to refactor and move over getEffectiveBoolean Value from the item class to the iterator class, and have it take an item iterator as a parameter (rather than an item), and returns a boolean or raises an error.

src/main/resources/test_files/runtime-spark/LocalClauses/WhereClauseLocal1.iq
CanBerker: I have put in most of these test cases but I'm having issues with two of them. Firstly, the query below gives the following error on both Zorba and Sparksoniq. I wasn't sure how the original query is meant to be/if there is a mistake.
>for $i in ()
>where $j := $i + 1
>return $i
(no URI):2,10: static error [err:XPST0003]: invalid expression: syntax error, unexpected ":="

Secondly, 
>let $i := ()
>where exists($i)
>return 1
>-> 1

Since where clause doesn't have a "next" in this case, it appears to be returning an empty sequence on both Zorba and Sparksoniq. However, your example indicates it should return 1. How should we proceed with this?

ghislainfourny:
Dear Can,

My query did have a mistake:

for $i in ()
where $j eq $i + 1
return $i

and you are correct about the other one:

let $i := ()
where exists($i)     (: which is false :)
return 1

-> empty sequence.

At some point the code becomes so robust that it beats our brains :-)

ghislainfourny:
@CanBerker 

One (big milestone) last thing: could you also check whether the first, big query in the tutorial now works? 

https://github.com/ghislainfourny/jsoniq-tutorial

src/main/resources/test_files/runtime-spark/LocalClauses/WhereClauseLocal6
```diff
@@ -0,0 +1,5 @@
+(:JIQS: ShouldRun; Output="" :)
+for $i in ()
+where $j := $i + 1
```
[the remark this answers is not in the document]
CanBerker: The test actually didn't succeed, as a little trick to deactivate the (buggy) tests temporarily, I just remove the ".iq" postfix.

>for $i in ()
>where $j eq $i + 1
>return $i

In the current state the test case given above causes "(no URI):2,7: static error [err:XPST0008]: "j": undeclared variable" both in Zorba and sparksoniq. Shall we add this as a "ShouldCrash" test case?

CanBerker:
Yes the big query indeed works. I had seen that example in the https://github.com/Sparksoniq/sparksoniq/issues/11 yesterday and felt like trying it out. I have added it as a test case to the system as well now.

src/main/resources/test_files/runtime-spark/LocalClauses/WhereClauseLocal6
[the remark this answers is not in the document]
ghislainfourny: I missed that too! Let's replace $j with an integer. I forgot about this. Of course, you are welcome to add one more test case with a ShouldCrash setting. Thanks!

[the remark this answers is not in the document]
CanBerker: I will add the mentioned(integer) test case and then merge the PR. Before that, I have one last question given below:

The following test annotation fails with the assertion given below:
(:JIQS: ShouldCrash; ErrorCode="XPST0008"; ErrorMetadata="LINE:3:COLUMN:6:" :)
for $i in ()
where $j eq $i + 1
return $i

java.lang.AssertionError: Program did not compile when expected to.
Error output: Error [err: XPST0008]LINE:3:COLUMN:6:Uninitialized variable reference: j

It appears that this test case causes a semantic exception which later hits a shouldCompile() check that always returns true, effectively causing the given exception. If semantic exceptions are a part of test cases handled at runtime, a fix may be needed here. If they are related to an earlier stage, this can be ignored. Should I create an issue on this?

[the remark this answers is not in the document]
ghislainfourny: Indeed, this program should parse, but static analysis should then spot the unbound variable. Can you open an issue?

ghislainfourny:
> Yes the big query indeed works. I had seen that example in the #11 yesterday and felt like trying it out. I have added it as a test case to the system as well now.

Wonderful!

PR #178, opened 2019-01-22
Fix performance while parsing.

ghislainfourny:
I tried to run a filtering query on the confusion dataset through a profiler and a lot of the overhead was due to object construction. But a lot of methods that are actually static were executed dynamically, creating a lot of empty objects.

I also rewrote the implementation of the partition map to avoid materialization within the mapPartition function.

A count of the confusion dataset now takes just 18 seconds. 

A filter takes below 15 seconds.

The status quo was... memory overflow on my machine. 😁


ghislainfourny:
Thanks! The tests are also 20 seconds faster.

PR #181, opened 2019-01-23
Reduce memory footprint.:

ghislainfourny:
To further improve performance, I reduced the memory footprint of items by removing item metadata -- instead, the metadata is set by the iterators directly (which is also more correct, as it is the iterator's line number and column number that shows where the error is, not where the item was created). No change to the tests was needed and all line and column numbers are unchanged.

Now items have the bare minimum: only one field for atomic items, just a list for arrays, just two lists for objects.

This improved the quantity of data I am able to group by on a single machine by 50% (Went from 400k to 600k objects).


ghislainfourny:
Thanks! :-)

PR #182, opened 2019-01-26
Refactor getEffectiveBooleanValue

CanBerker:
getEffectiveBoolean value in Item class is copied over to the RuntimeIterator as a static method. 
Where clause implementation at https://github.com/Sparksoniq/sparksoniq/pull/175 is on hold until this PR is merged.

If, And, Or, Not and Boolean iterators are updated with the new getEffectiveBooleanValue.
PredicateIterator is left unchanged and currently uses the implementation in Item class (the integer item is used as an index, rather than evaluating to the effective boolean value)

**Question:**
In your earlier comment in the other PR, you had mentioned that if the next() returns an object or an array the result is always true.
https://github.com/Sparksoniq/sparksoniq/pull/175#discussion_r250166394

However, I've found that the current implementation actually checks whether the object/array item is empty, and returns false if they are. Guessing that this is the correct approach I have added this too in my last commit.

Test cases added.


ghislainfourny:
> However, I've found that the current implementation actually checks whether the object/array item is empty, and returns false if they are. Guessing that this is the correct approach I have added this too in my last commit.

Yes. boolean({}) is true and boolean([]) too. boolean (()) is false.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -144,4 +147,56 @@ public IteratorMetadata getMetadata() {
...
+     * This function calculates the effective boolean value of the sequence given by iterator.
+     * Non-empty objects and arrays always return true.
+     * Empty sequence returns false.
+     * Singleton atomic values are evaluated to their effective boolean value.
+     * Multiple atomic values throw an exception.
+     *
+     * @param iterator has to be opened before calling this function
+     * @return
+     */
+    protected static boolean getEffectiveBooleanValue(RuntimeIterator iterator) {
```
ghislainfourny: This method is beautiful.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -144,4 +147,56 @@ public IteratorMetadata getMetadata() {
...
+                    result = !((DecimalItem) item).getDecimalValue().equals(0);
+                else {
+                    throw new SparksoniqRuntimeException("Unexpected numeric type found while calculating effective boolean value.");
+                }
+            } else if (item instanceof NullItem)
+                result = false;
+            else if (item instanceof StringItem)
+                result = !((StringItem) item).getStringValue().isEmpty();
+            else if (item instanceof ObjectItem)
+                return ((ObjectItem) item).getKeys() != null && !((ObjectItem) item).getKeys().isEmpty();
```
ghislainfourny: &&... should be removed (empty object maps to true).

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -144,4 +147,56 @@ public IteratorMetadata getMetadata() {
...
+            else if (item instanceof ArrayItem)
+                return ((ArrayItem) item).getItems() != null && !((ArrayItem) item).getItems().isEmpty();
```
ghislainfourny: &&... should be removed (empty array maps to true).

src/main/resources/test_files/runtime/ControlIterators/IfExpr10.iq
```diff
@@ -0,0 +1,9 @@
+(:JIQS: ShouldRun; Output="else branch" :)
```
ghislainfourny: first

src/main/resources/test_files/runtime/ControlIterators/IfExpr11.iq
```diff
@@ -0,0 +1,9 @@
+(:JIQS: ShouldRun; Output="else branch" :)
```
ghislainfourny: first

PR #184, opened 2019-02-04
Lower memory footprint.

ghislainfourny:
A few optimizations on the memory allocation behind HashMaps and ArrayLists, as well as interning Strings.

### DIFF
--- a/src/main/java/sparksoniq/io/json/JiqsItemParser.java
+++ b/src/main/java/sparksoniq/io/json/JiqsItemParser.java
@@ -36,7 +36,7 @@ public class JiqsItemParser implements Serializable {
     public static Item getItemFromObject(Object object, IteratorMetadata metadata) {
         if (object != null) {
             if (object instanceof String)
-                return new StringItem(object.toString().intern());
+                return new StringItem(object.toString());
             if (object instanceof Integer)
                 return new IntegerItem((int) object);
             if (object instanceof Double)
--- a/src/main/java/sparksoniq/io/json/JiqsItemParser.java
+++ b/src/main/java/sparksoniq/io/json/JiqsItemParser.java
@@ -36,22 +36,22 @@ public class JiqsItemParser implements Serializable {
     public static Item getItemFromObject(Object object, IteratorMetadata metadata) {
         if (object != null) {
             if (object instanceof String)
-                return new StringItem(object.toString(), ItemMetadata.fromIteratorMetadata(metadata));
+                return new StringItem(object.toString());
             if (object instanceof Integer)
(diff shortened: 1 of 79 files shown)


src/main/java/sparksoniq/io/json/JiqsItemParser.java
```diff
@@ -36,7 +36,7 @@
     public static Item getItemFromObject(Object object, IteratorMetadata metadata) {
         if (object != null) {
             if (object instanceof String)
-                return new StringItem(object.toString());
+                return new StringItem(object.toString().intern());
```
wscsprint3r: StackOverflow wisdom: "be aware that the cache is maintained by JVM in permanent memory pool which is usually limited in size compared to heap so you should not use intern if you don't have too many duplicate values."

I wonder how large this can grow in real life...

src/main/java/sparksoniq/spark/closures/InitialForClauseClosure.java
```diff
@@ -32,7 +32,7 @@ public InitialForClauseClosure(String variableName) {
 
     @Override
     public FlworTuple call(Item v1) throws Exception {
-        FlworTuple result = new FlworTuple();
+        FlworTuple result = new FlworTuple(1);
```
wscsprint3r: Maybe add secondary constructor with no args for the 1 case?

wscsprint3r:
Approved with 1 suggestion and one general question.

src/main/java/sparksoniq/io/json/JiqsItemParser.java
ghislainfourny: Do you have a link to this StackOverflow page, in particular, how old is it? I had this in mind, but then read that newer versions of Java (I think starting with 7) store these Strings in the JVM, no longer in the permanent memory pool.

src/main/java/sparksoniq/spark/closures/InitialForClauseClosure.java
ghislainfourny: Yes, I could do that -- however, I don't think it will ever happen that we do not know the size in advance. Actually, in the end, FLWOR tuples will be done with DataFrames because their structure is known in advance.

ghislainfourny:
Thanks, Stefan! I will keep this unmerged from now until we clarify that interned Strings are indeed on the heap.

wscsprint3r:
https://stackoverflow.com/questions/10578984/what-is-java-string-interning

ghislainfourny:
@wscsprint3r Since I would like to take more time for discussion and due diligence, I reverted the intern() calls. I will merge the rest.

PR #187, opened 2019-02-17
Orderby clause local

CanBerker:
Local evaluation for orderby clause is implemented. 

Handling empty ordering fields with greatest/least keywords are now fully supported.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/item/Item.java
+++ b/src/main/java/sparksoniq/jsoniq/item/Item.java
@@ -154,17 +154,6 @@ public abstract class Item implements SerializableItem {
             else{
                 result = 1;
             }
-        } else if (v1 == null || v2 == null) {
-            // null equals null
-            if (v1 == null && v2 == null) {
-                result = 0;
-            }
-            else if (v1 == null) {
-                result = -1;
-            }
-            else{
-                result = 1;
-            }
         } else {
             result = v1.serialize().compareTo(v2.serialize());
         }
--- a/src/main/java/sparksoniq/jsoniq/item/Item.java
+++ b/src/main/java/sparksoniq/jsoniq/item/Item.java
@@ -154,6 +154,17 @@ public abstract class Item implements SerializableItem {
             else{
                 result = 1;
             }
+        } else if (v1 == null || v2 == null) {
+            // null equals null
+            if (v1 == null && v2 == null) {
+                result = 0;
+            }
+            else if (v1 == null) {
+                result = -1;
+            }
+            else{
+                result = 1;
+            }
         } else {
             result = v1.serialize().compareTo(v2.serialize());
         }
--- a/src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
+++ b/src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
@@ -88,7 +88,21 @@ public class FlworKey implements KryoSerializable {
                     && ((!Item.isNumeric(comparisonItem) || !Item.isNumeric(currentItem)))) {
                 throw new SparksoniqRuntimeException("Invalid sort key: Item types can't be different.");
             } else {
-                result = Item.compareItems(currentItem, comparisonItem);
+                // handle the Java null placeholder used in orderByClauseSparkIterator
+                if (currentItem == null || comparisonItem == null) {
+                    // null equals null
+                    if (currentItem == null && comparisonItem == null) {
+                        result = 0;
+                    }
+                    else if (currentItem == null) {
+                        result = -1;
+                    }
+                    else{
+                        result = 1;
+                    }
+                } else {
+                    result = Item.compareItems(currentItem, comparisonItem);
+                }
             }
             // Simplify comparison result to -1/0/1
             result = (int) Math.signum(result);
--- a/src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
+++ b/src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
@@ -83,37 +83,37 @@ public class FlworKey implements KryoSerializable {
             if (currentItem instanceof ArrayItem || currentItem instanceof ObjectItem ||
                     comparisonItem instanceof ArrayItem || comparisonItem instanceof ObjectItem) {
                 throw new SparksoniqRuntimeException("Non atomic key not allowed");
-            } else if ((currentItem != null && comparisonItem != null)
+            }
+            if ((currentItem != null && comparisonItem != null)
                     && (!currentItem.getClass().getSimpleName().equals(comparisonItem.getClass().getSimpleName()))
                     && ((!Item.isNumeric(comparisonItem) || !Item.isNumeric(currentItem)))) {
                 throw new SparksoniqRuntimeException("Invalid sort key: Item types can't be different.");
-            } else {
-                // handle the Java null placeholder used in orderByClauseSparkIterator
-                if (currentItem == null || comparisonItem == null) {
-                    // null equals null
-                    if (currentItem == null && comparisonItem == null) {
-                        result = 0;
-                    }
-                    else if (currentItem == null) {
-                        result = -1;
-                    }
-                    else{
-                        result = 1;
-                    }
-                } else {
-                    result = Item.compareItems(currentItem, comparisonItem);
-                }
             }
+
+            // handle the Java null placeholder used in orderByClauseSparkIterator
+            if (currentItem == null || comparisonItem == null) {
+                // null equals null
+                if (currentItem == null && comparisonItem == null) {
+                    result = 0;
+                }
+                else if (currentItem == null) {
+                    result = -1;
+                }
+                else{
+                    result = 1;
+                }
+            } else {
+                result = Item.compareItems(currentItem, comparisonItem);
+            }
+
             // Simplify comparison result to -1/0/1
             result = (int) Math.signum(result);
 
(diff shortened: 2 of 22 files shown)


src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
```diff
@@ -17,46 +17,149 @@
...
 import sparksoniq.jsoniq.tuple.FlworKey;
 import sparksoniq.jsoniq.tuple.FlworTuple;
 import sparksoniq.semantics.DynamicContext;
 import sparksoniq.spark.closures.OrderByClauseSortClosure;
 import sparksoniq.spark.closures.OrderByMapToPairClosure;
-import sparksoniq.spark.iterator.flowr.base.FlowrClauseSparkIterator;
 import sparksoniq.spark.iterator.flowr.expression.OrderByClauseSparkIteratorExpression;
 
-import java.util.List;
+import java.util.*;
```
wscsprint3r: Just as a tip, this is in general bad practice.

CanBerker: Thanks a lot for the pointer. I believe this was done by automatic optimization of imports by IntelliJ

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
```diff
@@ -17,46 +17,149 @@
...
+    @Override
+    public void open(DynamicContext context) {
+        super.open(context);
+
+        // isRDD checks omitted, as open is used for non-RDD(local) operations
+
+        if (this._child != null) {
+            _child.open(_currentDynamicContext);
+
+            setAllLocalResults();
```
wscsprint3r: This means that all computation is essentially done on opening. For the local computation case this would not be terrible but it still breaks the iterator pattern.

src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
```diff
@@ -63,48 +63,84 @@ public FlworKey(List<Item> contents) {
...
+            }
+            else {
+                // if comparison results in a difference, return the comparison result and ordering expression index
                 return new ResultIndexKeyTuple(result, index);
+            }
+        }
+        // execution reaches here when "this" Flworkey (can be empty) has matching ordering results  with the other Flworkey
+
+        // if ordering expressions results have equal size, Flworkeys are fully equal
+        if ( sizeOfThisFlworKey == sizeOfOtherFlworKey) {
```
ghislainfourny: I don't think it ever happens that we compare two FLWOR keys with different sizes. You can simply throw a RuntimeException (to avoid a throw clause) if they don't match. This should simplify the code a lot.

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
ghislainfourny: Stefan is correct that, normally, this should be done in the first next() call. I think we do have other places where this will need to be fixed. Can you open an issue for this?

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
```diff
@@ -17,46 +17,149 @@
...
-        if(this._rdd == null) {
-            this._rdd = this._child.getRDD(context);
-            //map to pairs - ArrayItem [sort keys] , tuples
-            JavaPairRDD<FlworKey, FlworTuple> keyTuplePair = this._rdd
-                    .mapToPair(new OrderByMapToPairClosure(this._expressions, _isStable));
-            //sort by key
-            keyTuplePair = keyTuplePair.sortByKey(new OrderByClauseSortClosure(this._expressions, _isStable));
-            //map back to tuple RDD
-            this._rdd = keyTuplePair.map(tuple2 -> tuple2._2());
+        if (this._rdd == null) {
```
ghislainfourny: I think we are no longer caching RDDs. Since they depend on the context, this is now incorrect.

CanBerker: Yes, I missed this, it will be fixed.

src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
CanBerker: In the test cases I have debugged and observed this was not the case actually. Flworkeys of different sizes are compared particularly in the cases where one tuple doesn't have the necessary field. An example would be the following query:
```
jiqs$ for $i in (
   { "product" : "broiler", "store number" : 1, "quantity" : 20  },
   { "product" : "broiler", "store number" : 1, "quantity" : 30  },
   { "product" : "toaster", "store number" : 2  },
   { "product" : "toaster2", "store number2" : 2  })
order by $i."quantity" empty greatest
return $i
```
Flworkey of 1st and 2nd rows contain 1 entry
Flworkey of 3rd and 4th rows contain 0 entries.

Similarly:
```
for $i in (
   { "product" : "toaster", "store number" : 1  },
   { "product" : "broiler", "store number" : 1, "quantity" : 20  })
order by $i."store number", $i."quantity" descending empty greatest
return $i
```
1st row creates Flworkey with a single ordering field
2nd row creates a Flworkey with both ordering fields available.

Detection of empty fields are carried out using the index values returned from the Flworkey comparison in order to handle special empty field handling as greatest/least.

Please point out if you see any misconceptions. We can also discuss this further in this week's meeting.

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
CanBerker: You're right, since we need to complete results over the entire set, I had mistakenly pushed this computation to the open. This has been changed with checking whether child iterator hasNext in open and triggering computation only in the next() now.

let for and where clauses were implemented calculating only a single element in the open() so I believe they comply with the pattern. I had the same misconception for the groupby clause which is also fixed now. Should I open an issue to make a full sweep later?

ghislainfourny: Thanks! Yes, an issue is worth it. I think we have a few item iterators where we have materialization done in open(). Nothing urgent, though. Just for cleanliness.

CanBerker:
Result checking is made more restrictive including the ordering checks for spark tests in a new PR #190.

Review required on whether if any further changes are required for this PR. Testing improvement PR will need review once this PR is merged

src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
ghislainfourny: I see, thanks. Unfortunately, I am afraid that this implementation is not quite correct for two reasons:

1. The missing item may be a different positions in each row, which can "offset" the comparisons.

2. Whether an empty sequence is lowest or greatest depends on each variable, it is not for the entire order by clause.

Example:

for $i in (
   { "product" : "toaster", "store number" : 1  },
   { "product" : "broiler", "store number" : 1, "quantity" : 20  })
   { "product" : "broiler", "quantity" : 20  })
order by $i."store number" descending empty lowest, $i."quantity" descending empty greatest
return $i

I think the easiest way is simply to use Java null values in the FlowrKey map, and to handle them as part of the compareWithFlworKey() logics.

Also, this means that the return value can be an int, which would improve performance a lot comparing to creating a special object with two values.

Does it help?

CanBerker: Yes this was very helpful. I have just pushed a new implementation covering the issues and the test case you've given above.

The indices of the object were used to determine which variable(expression) had the empty item and which ascending/descending selection it had. I was able to preserve the usage of such indices without the use of objects. It is implemented through a small trick with multiplication of integers.

The item comparison results in (or is simplified to, for the case of string comparison) +1/0/1. This value is multiplied by the indice of the expression/variable that the comparison result is determined by. Then, this value is extracted in the orderby closure to handle the empty items as specified in the query for each individual variable.

src/main/java/sparksoniq/jsoniq/item/Item.java
```diff
@@ -154,6 +154,17 @@ else if (v1 instanceof NullItem) {
             else{
                 result = 1;
             }
+        } else if (v1 == null || v2 == null) {
```
ghislainfourny: While this is correct, I would rather have this code to compare Java nulls in compareWithFlowrKey() rather than here.

Indeed, the fact that null represents the empty sequence is an internal implementation detail of FlworKeys that we should "protect" the Item class from, for cleanliness. The Item class only deals with single items. This will avoid surprises in the future.

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
```diff
@@ -17,38 +17,151 @@
...
+            }
+            values.add(inputTuple);
+        }
+        return keyValuePairs;
+    }
+
     @Override
     public JavaRDD<FlworTuple> getRDD(DynamicContext context) {
-        if(this._rdd == null) {
+        if (this._child != null) {
```
ghislainfourny: The code here can be simplified (get rid of _rdd, and directly return the final result).

Issue #158, opened 2019-01-02
Testing Impr. - Ordered Results


CanBerker:
In the long term, we must compare while ensuring the order. JSONiq sequences are ordered lists. While in some cases, the order may be arbitrary (such as for group by clauses), it is usually stable and the tests are still deterministic, meaning that it is compatible with ordered output comparison. 

_Originally posted by @ghislainfourny in https://github.com/Sparksoniq/sparksoniq/pull/151#issuecomment-447828325_

PR #190, opened 2019-02-18
Spark tests ordered

CanBerker:
Spark test case checking is made much more restrictive by using the default Java string comparison.

Issue #158 addressed.


CanBerker:
Yes indeed, all it took was a little bit of formatting.

PR #191, opened 2019-02-21
Count clause local

CanBerker:
Will be reviewed after #188 is merged.

Concludes development of local Flwor clauses at #11


src/main/java/sparksoniq/spark/iterator/flowr/CountClauseSparkIterator.java
```diff
@@ -1,33 +1,86 @@
...
 import sparksoniq.semantics.DynamicContext;
 import sparksoniq.spark.closures.CountClauseClosure;
-import sparksoniq.spark.iterator.flowr.base.FlowrClauseSparkIterator;
 
-import java.util.Collections;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CountClauseSparkIterator extends SparkRuntimeTupleIterator {
+    private VariableReferenceIterator _variableReference;
```
ghislainfourny: I think all you need is a String with the name of the variable. This simplifies the signatures.

PR #194, opened 2019-03-03
Transition to Spark Session instead of only JavaSparkContext

CanBerker:
Previously only JavaSparkContext was used for all spark related operations. Unfortunately, this implementation doesn't support the necessary SparkSQL/DataFrames operations.

SparkSession which is defined as "the new entry point of API for Spark 2.0+" in many sources has been brought in with this PR. Default Spark application configuration is carried out with SparkSession. Then a SparkContext object is fetched which primarily support Scala related operations. A JavaSparkContext is created from the said SparkContext to perform previously used operations such as parallelize, textfile... etc.

SparkSQL artifact added in maven dependencies.
Commits are optimized by IntelliJ IDEA.
Edit: SparkContextManager class is removed and is replaced by SparkSessionManager.


src/main/java/sparksoniq/spark/SparkSessionManager.java
```diff
@@ -32,55 +13,51 @@
...
     }
 
     private void initialize() {
         initializeKryoSerialization();
         Logger.getLogger("org").setLevel(LOG_LEVEL);
         Logger.getLogger("akka").setLevel(LOG_LEVEL);
-        if (context == null)
-            context = new JavaSparkContext(this.configuration);
+        if (session == null)
+            session= SparkSession.builder().config(this.configuration).getOrCreate();
```
ghislainfourny: Can we throw an error if the session already exists? I am afraid that otherwise we can run into inconsistent states (if the public initialization functions are called with several configurations).

PR #199, opened 2019-03-11
Data Frames Primer w/ For and Return Clauses

CanBerker:
Kryo version updated to 5.0.0-RC2

Default constructors added to Item classes to support Kryo serialization.

KryoManager added to efficiently stored Kryo Object
- Kryo doesn't support multithreaded operation by default.
- Version 5.0.0-RC2 adds multithreading support by Pool implementation ( https://github.com/EsotericSoftware/kryo#thread-safety)
- Kryo object is managed by a KryoPool.
- Pool implementation is also implemented for Input/Output objects, however this caused nondeterministic bugs. The code artifacts remain in the manager, however local creation of Input/Output objects seems to be functional at the moment.

getDataFrame is partially implemented in "For" and "return" clauses.
- Serialization/deserialization pipeline works and correctly returns the given input for the example below
```
for $i in parallelize(("test",null,3.5)) 
return $i
```

- No expression evaluation is currently implemented (returns the same input as above w/o object creation). I might need some input on how we are going to tackle the expressions. 
```
for $i in parallelize(("test",null,3.5)) 
return {"i": $i }
```

getDataFrame placeholder functions have been placed in all the other clauses.

Lastly, I have started experimenting with non-initial for clauses. I imagine the process will be:
* Deserialize the data
    * Row object contains columns of byte[]. A byte [] can be deserialized to Object and cast to List<Item> (original data type). Each column can be processed individually and this would yield a List\<List\<Item\>\> for a single row.  A row can potentially be recreated from the list of deserialized columns but at this point, I was worried that I was getting side-tracked and losing the advantages of having Dataframes through too much manual manipulation.
* Apply expression
    * In the RDD structure, we would add Flwortuples to Dynamic context and evaluate the expressions. How are we going to evaluate the expressions in the Dataframe structure? Somehow map them into sql-like expressions? In that case we need to recreate the dataframe with the deserialized data in the previous section* Apply expression.
* Create new schema
* Form the new dataframe

### DIFF
--- a/src/main/java/sparksoniq/spark/closures/ClosureUtils.java
+++ b/src/main/java/sparksoniq/spark/closures/ClosureUtils.java
@@ -1,0 +1,38 @@
+package sparksoniq.spark.closures;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import org.apache.spark.sql.Row;
+import sparksoniq.jsoniq.item.Item;
+import sparksoniq.jsoniq.item.KryoManager;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.util.List;
+
+public class ClosureUtils {
+
+    public static byte[] serializeItemList(List<Item> toSerialize) {
+        Kryo kryo = KryoManager.getInstance().getOrCreateKryo();
+
+        Output output = new Output(new ByteArrayOutputStream());
+        kryo.writeClassAndObject(output, toSerialize);
+        output.close();
+
+        byte[] byteArray = output.getBuffer();
+        return byteArray;
+    }
+
+    public static Object deserializeRow(Row row) {
+        Input input = new Input(new ByteArrayInputStream((byte[])row.get(0)));
+
+        Kryo kryo = KryoManager.getInstance().getOrCreateKryo();
+
+        Object obj = kryo.readClassAndObject(input);
+        input.close();
+
+        return obj;
+    }
+
+}
--- a/src/main/java/sparksoniq/spark/closures/ClosureUtils.java
+++ b/src/main/java/sparksoniq/spark/closures/ClosureUtils.java
@@ -9,30 +9,54 @@ import sparksoniq.jsoniq.item.KryoManager;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.util.ArrayList;
 import java.util.List;
 
 public class ClosureUtils {
 
+    private static KryoManager KM = KryoManager.getInstance();
+
     public static byte[] serializeItemList(List<Item> toSerialize) {
-        Kryo kryo = KryoManager.getInstance().getOrCreateKryo();
+        Kryo kryo = KM.getOrCreateKryo();
 
         Output output = new Output(new ByteArrayOutputStream());
         kryo.writeClassAndObject(output, toSerialize);
         output.close();
 
         byte[] byteArray = output.getBuffer();
+
+        KM.releaseKryoInstance(kryo);
+
         return byteArray;
     }
 
-    public static Object deserializeRow(Row row) {
-        Input input = new Input(new ByteArrayInputStream((byte[])row.get(0)));
+    public static Object deserializeRowColumn(Row row, int columnIndex) {
+        Kryo kryo = KM.getOrCreateKryo();
 
-        Kryo kryo = KryoManager.getInstance().getOrCreateKryo();
+        Input input = new Input(new ByteArrayInputStream((byte[])row.get(columnIndex)));
 
         Object obj = kryo.readClassAndObject(input);
         input.close();
 
+        KM.releaseKryoInstance(kryo);
+
         return obj;
+    }
+
+
+    public static List<Object> deserializeEntireRow(Row row) {
+        Kryo kryo = KM.getOrCreateKryo();
+
+        ArrayList<Object> deserializedColumnObjects = new ArrayList<>();
+        for (int columnIndex = 0; columnIndex < row.length(); columnIndex++) {
+            Input input = new Input(new ByteArrayInputStream((byte[])row.get(columnIndex)));
+            Object deserializedColumnObject = kryo.readClassAndObject(input);
+            deserializedColumnObjects.add(deserializedColumnObject);
+            input.close();
+        }
+
+        KM.releaseKryoInstance(kryo);
+        return deserializedColumnObjects;
     }
 
 }
--- a/src/main/java/sparksoniq/spark/closures/ForClauseSerializeClosure.java
+++ b/src/main/java/sparksoniq/spark/closures/ForClauseSerializeClosure.java
@@ -17,37 +17,30 @@
  * Author: Stefan Irimescu
  *
  */
- package sparksoniq.jsoniq.runtime.iterator;
+package sparksoniq.spark.closures;
 
-import org.apache.spark.api.java.JavaRDD;
-import sparksoniq.exceptions.SparkRuntimeException;
+import org.apache.spark.api.java.function.Function;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.RowFactory;
 import sparksoniq.jsoniq.item.Item;
-import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
-import sparksoniq.semantics.DynamicContext;
 
+import java.util.ArrayList;
 import java.util.List;
 
-public abstract class LocalRuntimeIterator extends RuntimeIterator {
-    protected LocalRuntimeIterator(List<RuntimeIterator> children, IteratorMetadata iteratorMetadata) {
-        super(children, iteratorMetadata);
+public class ForClauseSerializeClosure implements Function<Item, Row> {
+
+    public ForClauseSerializeClosure() {
     }
 
+    /**
+     * @param item
+     * @return Row object, containing byte array of a singleton list containing the given item
+     */
     @Override
-    public JavaRDD<Item> getRDD(DynamicContext dynamicContext)
-    {
-        throw new SparkRuntimeException("Iterator has no RDDs", getMetadata());
-    }
+    public Row call(Item item) throws Exception {
+        List<Item> itemList = new ArrayList<>();
+        itemList.add(item);
 
-    @Override
-    public boolean isRDD(){ return false; }
-
-    @Override
-    public boolean isDataFrame() {
-        return false;
-    }
-
-    @Override
-    public boolean getDataFrame() {
-        throw new SparkRuntimeException("Iterator has no DataFrames", getMetadata());
+        return RowFactory.create(ClosureUtils.serializeItemList(itemList));
     }
 }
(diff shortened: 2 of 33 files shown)


src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -195,4 +199,66 @@ public void close() {
...
+        this._df = null;
+
+        if (this._child == null) {
+            // create initial RDD from expression
+            JavaRDD<Item> initialRdd = this.getNewRDDFromExpression(_expression);
+
+            // define a schema
+            String schemaString = _variableName;
+            List<StructField> fields = new ArrayList<>();
+            for (String fieldName : schemaString.split(" ")) {
```
ghislainfourny: Can you explain this to me (split, for...)? There is only one variable here and its name is the string _variableName.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -195,4 +199,66 @@ public void close() {
...
+                StructField field = DataTypes.createStructField(fieldName, DataTypes.BinaryType, true);
+                fields.add(field);
+            }
+            StructType schema = DataTypes.createStructType(fields);
+
+            // convert initial RDD to row RDD
+            JavaRDD<Row> rowRDD = initialRdd.map(new ForClauseSerializeClosure());
```
ghislainfourny: Perfect.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -195,4 +199,66 @@ public void close() {
...
     }
+
+
+    @Override
+    public Dataset<Row> getDataFrame(DynamicContext context) {
+        this._df = null;
+
+        if (this._child == null) {
+            // create initial RDD from expression
+            JavaRDD<Item> initialRdd = this.getNewRDDFromExpression(_expression);
```
ghislainfourny: Small bug here. getNewRDDFromExpression should take context as the second parameter, it should not read it from _currentDynamicContext. Maybe it can be fixed in another PR or you can open an issue.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -195,4 +199,66 @@ public void close() {
...
+            // apply the schema to row RDD
+            this._df = SparkSessionManager.getInstance().getOrCreateSession().createDataFrame(rowRDD, schema);
+
+        } else {        //if it's not a start clause
+            if (_child.isDataFrame()) {
+                this._df = this._child.getDataFrame(context);
+
+                // TODO: deserialize the byte array dataframe
+                // line below and related closure is currently very experimental
+                Dataset<Row> ds = this._df.map(new ForClauseDeserializeClosure(), Encoders.bean(Row.class));
```
ghislainfourny: I think that all the logics mapping an input Row to one or more output Rows (including the deserialization of the needed columns or reserialization of the newly added column) according to the for expression should be done inside a single new closure (so you probably do not need the deserialization closure). The output Row schema should probably be created here, it is not clear to me yet how the closure is informed of this output Row schema.

src/main/java/sparksoniq/spark/closures/ForClauseSerializeClosure.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * @param item
+     * @return Row object, containing byte array of a singleton list containing the given item
+     */
+    @Override
+    public Row call(Item item) throws Exception {
+        List<Item> itemList = new ArrayList<>();
+        itemList.add(item);
+
+        return RowFactory.create(ClosureUtils.serializeItemList(itemList));
```
ghislainfourny: Perfect. So you picked approach 1 (serializing the list rather than the item) for now.

src/main/java/sparksoniq/spark/iterator/flowr/ReturnClauseSparkIterator.java
```diff
@@ -55,9 +57,36 @@ protected boolean initIsRDD() {
...
+        Dataset<Row> df = this._child.getDataFrame(context);
+        JavaRDD<Item> rdd = df.javaRDD().flatMap(new ReturnClauseDeserializeClosure());
+
+
+        // TODO: apply return expression
+        // itemRDD = rdd.flatMap(new ReturnFlatMapClosure(expression));
+        return rdd;
+    }
+    /*
+    public JavaRDD<Item> getDataFrame(DynamicContext context) {
```
ghislainfourny: The return clause doesn't return any dataframes! It still returns an RDD of Items as before (but will consume the dataframe output by its child clause, if any) :-)

DataFrames only replace RDDs of tuples.

src/main/java/sparksoniq/spark/closures/ForClauseSerializeClosure.java
ghislainfourny: Perfect. So you picked approach 1 (serializing the list rather than the item) for now.

src/main/java/sparksoniq/spark/closures/ReturnClauseDeserializeClosure.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * @param row
+     * @return List<Item>iterator, deserialized from parametrized Row object containing the List<Item>
+     */
+    @Override
+    public Iterator<Item> call(Row row) {
+        Object deserializedRow = ClosureUtils.deserializeRowColumn(row, 0);
+
+        return ((List<Item>) deserializedRow).iterator();
```
ghislainfourny: So I think that all that is missing here is passing the return expression, and invoking it with the deserialized values passed to its context via the variables corresponding to the columns.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
ghislainfourny: For future reference (please don't worry about it just now): I *suspect* that passing Encoders.kryo(Item.class) to a DataFrame map() with a trivial, idempotent closure *may* directly take care of all the serialization for you, which could make a lot of the code dealing with kryo serialization superfluous.

https://spark.apache.org/docs/2.1.2/api/java/org/apache/spark/sql/Encoders.html#kryo(scala.reflect.ClassTag)

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: You're absolutely right that it's an overkill here. It was the code I copied from original Spark documentation and I have been preserving it as is, for future use, when we have multiple columns (it also helps me simulate multiple columns scenerio even now while experimenting with it).

CanBerker:
Great news! Tests are functional again. Currently this branch is based on the testing changes I have made in the #201 PR. Once the issues with that PR is resolved and merged to master, I will rebase this PR accordingly to do one final check on tests.

ghislainfourny:
Wonderful. Congratulations!

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class ClosureUtils {
+
+    private static KryoManager KM = KryoManager.getInstance();
+
+    public static byte[] serializeItemList(List<Item> toSerialize) {
+        Kryo kryo = KM.getOrCreateKryo();
+
+        Output output = new Output(new ByteArrayOutputStream());
+        kryo.writeClassAndObject(output, toSerialize);
```
ghislainfourny: We know the class statistically, so that kryo.writeObject(output, toSerialize) will be more efficient.

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+        return byteArray;
+    }
+
+    public static Object deserializeRowColumn(Row row, int columnIndex) {
+        Kryo kryo = KM.getOrCreateKryo();
+
+        Input input = new Input(new ByteArrayInputStream((byte[])row.get(columnIndex)));
+
+        Object obj = kryo.readClassAndObject(input);
```
ghislainfourny: kryo.readObject(input, List<Item>.class) will be more efficient (see above).

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+public class ClosureUtils {
+
+    private static KryoManager KM = KryoManager.getInstance();
+
+    public static byte[] serializeItemList(List<Item> toSerialize) {
+        Kryo kryo = KM.getOrCreateKryo();
+
+        Output output = new Output(new ByteArrayOutputStream());
```
ghislainfourny: KM.getOrCreateOutput() to use the pool as well (no need to use a ByteArrayOutputStream as the Output class does this already).

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
```diff
@@ -0,0 +1,62 @@
...
+        return obj;
+    }
+
+
+    public static List<Object> deserializeEntireRow(Row row) {
+        Kryo kryo = KM.getOrCreateKryo();
+
+        ArrayList<Object> deserializedColumnObjects = new ArrayList<>();
+        for (int columnIndex = 0; columnIndex < row.length(); columnIndex++) {
+            Input input = new Input(new ByteArrayInputStream((byte[])row.get(columnIndex)));
```
ghislainfourny: KM.getOrCreateInput() and then setBuffer() to use the pool as well (no need to use a ByteArrayInputStream as the Input class does this already).

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public static Object deserializeRowColumn(Row row, int columnIndex) {
+        Kryo kryo = KM.getOrCreateKryo();
+
+        Input input = new Input(new ByteArrayInputStream((byte[])row.get(columnIndex)));
+
+        Object obj = kryo.readClassAndObject(input);
+        input.close();
+
+        KM.releaseKryoInstance(kryo);
```
ghislainfourny: We probably only want to release in the destructor of the class, otherwise this kryo instance keeps getting created and destroyed.

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+        return byteArray;
+    }
+
+    public static Object deserializeRowColumn(Row row, int columnIndex) {
+        Kryo kryo = KM.getOrCreateKryo();
+
+        Input input = new Input(new ByteArrayInputStream((byte[])row.get(columnIndex)));
```
ghislainfourny: KM.getOrCreateInput() and then setBuffer() to use the pool as well (no need to use a ByteArrayInputStream as the Input class does this already).

PR #209, opened 2019-03-24
Fix multiple forClauses.

CanBerker:
In the previous implementation For-return statements were functional, however for-for...-return statements were broken. 

This PR fixes bugs and improves schema transfer between clauses.
  - RowEncoder used in the implementation
  - Schema given as parameter to clauses (rows do not reliably carry schema information)
  - Test cases added

Edit: conditional flow in getDataFrame has been changed to work with return statements instead of else's.


src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -209,56 +209,51 @@ public void close() {
...
-                this._df = this._child.getDataFrame(context);
+        if (_child.isDataFrame()) {
+            Dataset<Row> df = this._child.getDataFrame(context);
 
-                // Add new column name to existing schema
-                String[] columnNamesPrimitive = this._df.schema().fieldNames();
-                ArrayList<String> columnNames = new ArrayList<>(Arrays.asList(columnNamesPrimitive));
-                columnNames.add(_variableName);
+            StructType oldSchema = df.schema();
+            StructType newSchema = oldSchema.add(_variableName, DataTypes.BinaryType);
```
ghislainfourny: Does this code also work with variable hiding?

Test:
for $i in parallelize((1 to 3)) for $i in 4 to 6 return $i
which returns 4 5 6 4 5 6 4 5 6

If the variable is hidden, you may need to replace, rather than add the variable in the schema.

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
```diff
@@ -20,36 +20,34 @@
...
 
-    public ForClauseFlatMapClosure(RuntimeIterator expression) {
+    public ForClauseFlatMapClosure(RuntimeIterator expression, StructType oldSchema) {
         this._expression = expression;
+        this._oldSchema = oldSchema;
     }
 
     @Override
     public Iterator<Row> call(Row row) {
-        StructType schema = row.schema();
```
ghislainfourny: Could you explain what is wrong (or what is a bug or a known bug) with row.schema()? It sounds intriguing indeed that this doesn't work properly, as DataFrames documents afaik that rows carry their schema. Or maybe this has to do with the way they are initialized since we are making our own rows.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Both the alpha and the dataframes (to my surprise) show the identical behavior given below. I will be adding test cases for variable hiding as well.
```
jiqs$ for $i in parallelize((1 to 3)) for $i in 4 to 6 return $i
>>>
>>>
(4, 5, 6, 4, 5, 6, 4, 5, 6)
jiqs$ for $i in parallelize((1 to 5)) for $i in 4 to 6 return $i
>>>
>>>
(4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6)
jiqs$ for $i in parallelize((1 to 5)) for $i in 4 to 5 return $i
>>>
>>>
(4, 5, 4, 5, 4, 5, 4, 5, 4, 5)
```

I believe this works out of the box due to our usage of Dynamic context as shown below:
```
// Taken from ForClauseMapClosure.java Line 55
// Create dynamic context with deserialized data
DynamicContext context = new DynamicContext();
for (int columnIndex = 0; columnIndex < columnNames.length; columnIndex++) {
    context.addVariableValue(columnNames[columnIndex], rowColumns.get(columnIndex));
}
```
The second column with the identical name overwrites the existing value in the Dynamic context's hashmap (addVariableValue calls put function of Hashmap -> Java Doc: https://docs.oracle.com/javase/8/docs/api/java/util/Hashtable.html#put-K-V-)

Should we replace the variable or is the current implementation good?

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
CanBerker: Your guess is very much on point. When dataframes are created through createDataFrame function the rows indeed accurately carry the schema information. I believe the rows are attached the schema information during this function's call. 

In our case, DF manipulation in the consequent operations is carried out in the following form:
```
return df.flatMap(new ForClauseFlatMapClosure(_expression, oldSchema), RowEncoder.apply(newSchema));
``` 
New rows are created within the closure and the RowEncoder.apply(newSchema) is applied on the results. This applies the new schema at the dataframe level, but individual rows don't contain the schema information. Fortunately for our use case, this information is not needed at the row level as we can simply pass the schema information to the closure.

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
```diff
@@ -30,6 +31,21 @@
         return byteArray;
     }
 
+    public static Row reserializeRowWithNewData(Row prevRow, List<Item> newColumn, int duplicateColumnIndex) {
+        List<byte[]> newRowColumns = new ArrayList<>();
```
ghislainfourny: This should also be a field of the closure class, constructed once there, and passed as a parameter to this function. It must be cleared every time.

A few more thoughts:
The input and output buffers should also be fields in the closure class, as well as the input and output objects, using these buffers. The use of array byte streams is documented as superfluous by kryo. The function serializeItemList should then return Output.position() rather than the buffer (there is a "memory leak" here, as a lot of byte arrays are created, which are longer than the bytes actually used). Then, only the relevant part of the buffer (0 to returned position) should be copied over into the row to optimize space.

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
```diff
@@ -20,36 +20,35 @@
...
-import sparksoniq.spark.SparkSessionManager;
 
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
 public class ForClauseFlatMapClosure implements FlatMapFunction<Row, Row> {
-    RuntimeIterator _expression;
+    private RuntimeIterator _expression;
+    private StructType _oldSchema;
```
ghislainfourny: I suggest to use the terminology "inputSchema" and "outputSchema" to be more accurate.

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
ghislainfourny: I understand, thank you for the precision. I think this hack is fine for now, but it is worth opening an issue to investigate it fully later on. It may actually even "solve itself" when we harness Spark SQL.

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
```diff
@@ -20,36 +20,35 @@
...
 
     @Override
     public Iterator<Row> call(Row row) {
-        StructType schema = row.schema();
-        String[] columnNames = schema.fieldNames();
+        String[] columnNames = _oldSchema.fieldNames();
 
         // Deserialize row
         List<Object> deserializedRow = ClosureUtils.deserializeEntireRow(row);
         List<List<Item>> rowColumns = new ArrayList<>();
```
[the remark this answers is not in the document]
CanBerker: I have moved actually decided to make these changes out of curiosity. After moving all these new statements into private variables of the closure, the performance is nearly identical to the non-dataframe implementation. Results can be seen in the performance repo at https://github.com/CanBerker/sparksoniq-dataframe-perf-v1  (the jars are also available)

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
CanBerker: I have briefly experimented with the Output.toBytes(). This works with the output part of the picture but when the data is read through the input, following exception is thrown:
org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 15.0 failed 1 times, most recent failure: Lost task 1.0 in stage 15.0 (TID 61, localhost, executor driver): com.esotericsoftware.kryo.KryoException: Buffer underflow.

I will play around a bit more with the Kryo input/output classes.

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
CanBerker: Perfect suggestion, changes made.

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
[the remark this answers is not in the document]
CanBerker: I realized that I had forgotten the custom partition number of json-file in the initial run. I have also added this to the query and got the medium - 5m test set working. Results are fairly close between DF and alpha (Dataframes are even faster than the alpha by about 7 seconds).

The full dataset didn't work though. I will increase partitioning and try again.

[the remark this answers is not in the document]
ghislainfourny: This is very good news!

In the long-term we may want to introduce a default partitioning number for local execution, because this also confused the students. On HDFS, the partitions are built automatically to match 128MB blocks, but in a local execution only one partition is made no matter how big the file is.

src/main/java/sparksoniq/spark/closures/ClosureUtils.java
ghislainfourny: Thanks! If this causes issues, feel free to merge the PR and leave this aside for later. This is something that I think is not in the way of starting with the change to Spark SQL, i.e., it can be investigated in parallel (this code will be in a user-defined function rather than a closure, but the spirit remains identical).

CanBerker: I have managed to get the Output.toBytes() working.

Additionally, ByteBufferInput/Output caught my eye in the kryo documentation and I have started using these as our input/output classes. They have provided a somewhat noticeable performance boost in our local tests. I'm hoping their effects will show up in the tests with data.

ghislainfourny: Very nice!

Also good to hear about these alternative input/outputs. The performance boost sounds good.

src/main/java/sparksoniq/spark/closures/ForClauseFlatMapClosure.java
ghislainfourny: Thanks!

PR #210, opened 2019-03-24
Implement let clause with dataframes.

CanBerker:
Dataframes with LetClause implemented. 
Test cases added.

### DIFF
--- a/src/main/java/sparksoniq/spark/closures/LetClauseMapClosure.java
+++ b/src/main/java/sparksoniq/spark/closures/LetClauseMapClosure.java
@@ -17,33 +17,70 @@
  * Author: Stefan Irimescu
  *
  */
- package sparksoniq.spark.closures;
+package sparksoniq.spark.closures;
 
-import org.apache.spark.api.java.function.Function;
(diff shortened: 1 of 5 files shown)


src/main/java/sparksoniq/spark/closures/LetClauseMapClosure.java
```diff
@@ -17,33 +17,61 @@
...
+            Item nextItem = _expression.next();
+            newColumn.add(nextItem);
+        }
         _expression.close();
-        v1.putValue(_variableName, result, true);
-        return v1;
+
+        List<byte[]> newRowColumns = new ArrayList<>();
+        for (List<Item> column : rowColumns) {
+            newRowColumns.add(ClosureUtils.serializeItemList(column));
```
ghislainfourny: Can we not just copy over the existing binary values from the former row? Serializing again looks redundant and is a performance overhead. I think only the new column needs to be serialized.

This comment may also apply to the for closure if the same is done here, and that could help improve performance.

ghislainfourny:
Depends on #209

PR #211, opened 2019-03-24
Where clause df

CanBerker:
Whereclause implemented.

Branch resides on top of letClause PR. This can be evaluated after it is merged.

### DIFF
--- a/src/main/java/sparksoniq/spark/iterator/flowr/WhereClauseSparkIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/WhereClauseSparkIterator.java
@@ -22,6 +22,8 @@
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.encoders.RowEncoder;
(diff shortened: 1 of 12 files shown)


src/main/java/sparksoniq/spark/iterator/flowr/WhereClauseSparkIterator.java
```diff
@@ -111,15 +112,19 @@ private void setNextLocalTupleResult() {
...
         if (this._child != null) {
             this._rdd = _child.getRDD(context);
-            this._rdd = this._rdd.filter(new WhereClauseClosure(_expression));
-        } else {
-            throw new SparksoniqRuntimeException("Invalid where clause.");
+            this._rdd = this._rdd.filter(new OLD_WhereClauseClosure(_expression));
+            return _rdd;
         }
-        return _rdd;
+        throw new SparksoniqRuntimeException("Invalid where clause.");
```
ghislainfourny: In this kind of situation, for cosmetics, you can negate the if condition, throw the exception if child==null and then continue with the code with one less level of nesting, which improves readability (this is a general good practice).

ghislainfourny:
Depends on #210 and #209

PR #208, opened 2019-03-23
Local clause cleanup

CanBerker:
Remove unreached getNewRDDFromExpression code block in for and let clauses.
Earlier issue found here:
https://github.com/Sparksoniq/sparksoniq/issues/176

PR #212, opened 2019-03-24
Local query logging

CanBerker:


CanBerker:
Precisely, this PR was based on the previous PR #208

PR #216, opened 2019-04-02
Fix string serialization in arrays.

ghislainfourny:
Strings in arrays were unquoted.


ghislainfourny:
Thank you, Can! Yes, we can discuss these details later on today.

Issue #119, opened 2018-11-21
Issue with escaping


ghislainfourny:
This doesn't work quite right on the shell.

"This is a newline\nand an escaped \\"quote\\"."

CanBerker:
@ghislainfourny 
Should the following line:
> "This is a newline\nand an escaped "quote"."
be of the format:
> "This is a newline\nand an escaped \"quote\"."

If not, could you elaborate ?

ghislainfourny:
You are correct!

ghislainfourny:
After investigating: there is nothing wrong with the lexer.

The issue is with the JLines shell, which interprets \ as escaping sequences before the query reaches Sparksoniq. We should see if we can deactivate this and forward the raw input to the engine.

PR #217, opened 2019-04-02
Fix quote escaping when serializing strings.

ghislainfourny:
Quotes and newlines were not escaped properly.


src/main/resources/test_files/runtime/QuoteEscaping.jq
```diff
@@ -0,0 +1,2 @@
+(:JIQS: ShouldRun; Output="{ "foo\"foo\nbar" : "foo\"foo\nbar" }" :)
+json-file("./src/main/resources/quotes.json")
```
CanBerker: The test file is actually located under ./src/main/resources/**queries**/quotes.json
I'm surprised that the testing suite works with either of these address inputs. The shell application only accepts the latter as a valid parameter. Throws an input path does not exist error otherwise

CanBerker:
I just wanted to present the following cases from the shell application for sanity checking as I'm not very familiar with the escaping mechanism of sparksoniq. The grammar file indicates that ESC is detected with "\\" characters. Are the following behavior as expected? (particularly the output of the second query)

```
jiqs$ { "foo\"foo\nbar" : "foo\"foo\nbar" }
>>>
>>>
[ERROR] Error [err: XPST0003]LINE:1:COLUMN:14:Parser failed. foonbar
jiqs$ { "foo\\"foo\\nbar" : "foo\\"foo\\nbar" }
>>>
>>>
{ "foo\\"foo\nbar" : "foo\\"foo\nbar" }
```

src/main/resources/test_files/runtime/QuoteEscaping.jq
ghislainfourny: Very good catch. The test wasn't executed because the extension must be .iq. I have fixed this.

ghislainfourny:
> I just wanted to present the following cases from the shell application for sanity checking as I'm not very familiar with the escaping mechanism of sparksoniq. The grammar file indicates that ESC is detected with "\" characters. Are the following behavior as expected? (particularly the output of the second query)
> 
> ```
> jiqs$ { "foo\"foo\nbar" : "foo\"foo\nbar" }
> >>>
> >>>
> [ERROR] Error [err: XPST0003]LINE:1:COLUMN:14:Parser failed. foonbar
> jiqs$ { "foo\\"foo\\nbar" : "foo\\"foo\\nbar" }
> >>>
> >>>
> { "foo\\"foo\nbar" : "foo\\"foo\nbar" }
> ```

This is a known bug that is orthogonal to this PR, namely, string literals in a query don't support escaping yet ( see https://github.com/Sparksoniq/sparksoniq/issues/119 ). This is the reason why I used JSON input for testing my change, because there the escaping is done correctly by the JSON parser. When Issue #119 gets solved, the queries you tried will work correctly.

ghislainfourny:
The tests fail because of the way the expected output is parsed by the testing code. I will take a closer look.

ghislainfourny:
I fixed the test -- note that the test output string has always been a "hack" where the newline escape sequence was manually interpreted as a new line. I have fixed it so that \\n is indeed the actual two-character string "\n".

All of this will be fixed when we migrate our tests to a cleaner setup with the output in a separate file, exactly matching the actual output.

ghislainfourny:
Thanks a lot, Can!

PR #218, opened 2019-04-02
Fixing serialized output to follow standard practice

ghislainfourny:
In the interactive shell as well as in local output, the serialization should put one item per line, with no parentheses or commas around -- this was already the case when writing back to HDFS.

(For tests, the behavior is unchanged as this is internal and it can wait until the test refactoring).

In particular, as a consequence of this fix and the other two fixing PRs, now, outputs can be re-read as inputs.

### DIFF
--- a/src/main/java/sparksoniq/JsoniqQueryExecutor.java
+++ b/src/main/java/sparksoniq/JsoniqQueryExecutor.java
@@ -82,7 +82,7 @@ public class JsoniqQueryExecutor {
         generateStaticContext(visitor.getQueryExpression());
         // generate iterators
         RuntimeIterator result = generateRuntimeIterators(visitor.getQueryExpression());
-        String output = runIterators(result, true);
+        String output = runIterators(result);
         long endTime = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


src/main/java/sparksoniq/JsoniqQueryExecutor.java
```diff
@@ -225,36 +225,34 @@ private RuntimeIterator generateRuntimeIterators(Expression expression) {
...
+    private String getIteratorOutput(RuntimeIterator iterator) {
         iterator.open(new DynamicContext());
         Item result = null;
         if (iterator.hasNext()) {
             result = iterator.next();
         }
-        if (result == null) {
-            return "";
-        }
         String singleOutput = result.serialize();
```
CanBerker: This line will call serialize method on the null object and crash the app, (for example just put in the query "()" to return empty sequence).

I suggest keeping result==null check intact or making sure in some other way that the exception isn't thrown.

ghislainfourny:
Merging as discussed in the meeting. Thanks!

PR #222, opened 2019-04-04
Optimized for faster local output.

ghislainfourny:
This should help :)

Query 1 with the entire dataset takes less than 25 seconds.


ghislainfourny:
Thanks, Can! Solving the conflict then merging. This should unblock you.

PR #235, opened 2019-04-12
Fix idiv result.

ghislainfourny:
idiv must always return an integer. Fixes https://github.com/Sparksoniq/sparksoniq/issues/74

https://www.w3.org/TR/xpath-functions-31/#func-numeric-integer-divide


ghislainfourny:
Thank you!

PR #236, opened 2019-04-14
Count clause sql

CanBerker:
Count clause is implemented with dataframes sql.

Method given below is used as template and translated into Java.
https://stackoverflow.com/questions/30304810/dataframe-ified-zipwithindex

After implementing the tests I have noticed that the partition_id was starting from a non-zero value sometimes. ~~In order to solve this problem I normalize the values to start from 0 in order to be able to use them as array indices.~~

Technically, if the partition_id's are assigned non-sequentially, this can also break execution. ~~I don't know if this this possible as documentation is not detailed~~ and the behavior of the function is explained to be undeterministic at the following page:
https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/functions.html#spark_partition_id--

Edit: When partition count is increased and empty partitions appear, exceptions occur b/c of non-sequential partition_id's. ~~Working on this now,~~ This issue is resolved using a map of partitionIds. This has also provided a more elegant solution for the id's not starting at 0 problem. Test case added to prevent regressions.


src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -48,19 +62,6 @@
         return byteArray;
     }
 
-    public static Object deserializeByteArray(byte[] toDeserialize) {
-        Kryo kryo = KM.getOrCreateKryo();
```
CanBerker: Function moved within file, purely cosmetic change

src/main/java/sparksoniq/spark/iterator/flowr/CountClauseSparkIterator.java
```diff
@@ -95,6 +97,24 @@ private void setNextLocalTupleResult() {
 
     @Override
     public Dataset<Row> getDataFrame(DynamicContext context) {
-        return null;
+        if (this._child == null) {
+            throw new SparksoniqRuntimeException("Invalid where clause.");
```
ghislainfourny: Invalid count clause?

src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -146,4 +159,31 @@ public static Object deserializeRowColumn(Row row, int columnIndex) {
...
+    public static Dataset<Row> zipWithIndex(Dataset<Row> df, Long offset, String indexName) {
+        Dataset<Row> dfWithPartitionId = df
+                .withColumn("partition_id", spark_partition_id())
+                .withColumn("inc_id", monotonically_increasing_id());
+
+        Object partitionOffsetsObject = dfWithPartitionId
+                .groupBy("partition_id")
+                .agg(count(lit(1)).alias("cnt"), first("inc_id").alias("inc_id"))
+                .orderBy("partition_id")
+                .select(col("partition_id"), sum("cnt").over(Window.orderBy("partition_id")).minus(col("cnt")).minus(col("inc_id")).plus(lit(offset).alias("cnt")))
```
ghislainfourny: It would be great if this could be indented and arranged over multiple lines to make it more readable.

PR #237, opened 2019-04-16
Escaping

ghislainfourny:
Fix escaping issues (known as #119  )

- get JLine to not resolve backslash escapes, forwarding the string as it is.
- use Apache Commons to unescape while parsing and to escape while serializing.


src/main/java/sparksoniq/io/shell/JiqsJLineShell.java
```diff
@@ -110,8 +111,10 @@ private void initialize() throws IOException {
         Terminal terminal = TerminalBuilder.builder()
                 .system(true)
                 .build();
+        DefaultParser parser = new DefaultParser();
+        parser.setEscapeChars(null);
```
ghislainfourny: Deactivate \ escaping on JLine so Apache Commons can do so instead.
Version needed to be updated to 3.11.0 because of a known bug.

ghislainfourny:
And thank you!

PR #238, opened 2019-04-23
Fix local execution of count clause

### LINKED ISSUE #229: Count clause doesn't reset counter
for $i in 1 to 4
return
  for $j in 1 to 4
  count $c
  return $c

should return 1 2 3 4 1 2 3 4 1 2 3 4 1 2 3 4 

but returns 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16

ghislainfourny:
Fixes #229


ghislainfourny:
Thank you!

PR #243, opened 2019-04-26
Simplification of Item API

ghislainfourny:
The intent of such an API design, with all functions present in the top class, is that casts should not be necessary on the consumer side: the consumer of the Item API should not have to be aware of any of the involved subclasses. Method calls on an Item instance should be sufficient for all purposes for proper encapsulation.

This is fundamental as this API will be used and exposed to Sparksoniq users in language bindings later on, but this also will be directly useful to simplify our own code.

The casts were forced by exceptions thrown on improper API use to avoid propagating "throw" signatures, although this kind of exceptions should be used if they are part of the API itself. They are now replaced with RuntimeExceptions that indicate when the Item API's contract is not fulfilled.

This simplifies the code at many places, and instance of tests as well as casts are now superfluous.

This is a long PR, but the changes are very repetitive. This may also help with performance as we get rid of casting, exception try-catch as well as type testing.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/item/Item.java
+++ b/src/main/java/sparksoniq/jsoniq/item/Item.java
@@ -160,7 +160,7 @@ public abstract class Item implements SerializableItem {
     }
 
     public List<Item> getItems() {
-        throw new RuntimeException("Item is not an array or object.");
+        throw new RuntimeException("Item is not an array.");
     };
 
     public Item getItemAt(int i) {
@@ -176,7 +176,7 @@ public abstract class Item implements SerializableItem {
     };
 
     public Collection<? extends Item> getValues() {
-        throw new RuntimeException("Item is not an array or object.");
+        throw new RuntimeException("Item is not an object.");
     };
 
     public Item getItemByKey(String s) {
@@ -192,7 +192,7 @@ public abstract class Item implements SerializableItem {
     };
 
     public String getStringValue() {
-        throw new RuntimeException("getStringValue: Item is not a string.");
+        throw new RuntimeException("Item is not a string.");
     };
 
     public boolean getBooleanValue() {
--- a/src/main/java/sparksoniq/jsoniq/item/Item.java
+++ b/src/main/java/sparksoniq/jsoniq/item/Item.java
@@ -163,31 +163,57 @@ public abstract class Item implements SerializableItem {
         return compareItems(v1, v2) == 0;
     }
 
-    public abstract List<Item> getItems() throws OperationNotSupportedException;
+    public List<Item> getItems() {
+        throw new RuntimeException("Item is not an array or object.");
+    };
 
-    public abstract Item getItemAt(int i) throws OperationNotSupportedException;
+    public Item getItemAt(int i) {
+        throw new RuntimeException("Item is not an array.");
+    };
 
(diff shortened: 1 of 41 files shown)


src/main/java/sparksoniq/jsoniq/item/ObjectItem.java
```diff
@@ -77,11 +77,7 @@ else if (values.size() == 1) {
                 valueList.add(value);
             }
```
CanBerker: This is not related to the changes you've made but could you remove the @param itemMetadata from JavaDoc in line 59

src/main/java/sparksoniq/jsoniq/item/Item.java
```diff
@@ -163,31 +159,57 @@ public static boolean checkEquality(Item v1, Item v2) {
...
+    };
 
-    public abstract int getSize() throws OperationNotSupportedException;
+    public int getSize() {
+        throw new RuntimeException("Item is not an array.");
+    };
 
-    public abstract String getStringValue() throws OperationNotSupportedException;
+    public String getStringValue() {
+        throw new RuntimeException("getStringValue: Item is not a string.");
```
CanBerker: Is there a reason why the exception message for getStringValue is different than other methods? I'd suggest keeping the standard similar to other methods.

src/main/java/sparksoniq/jsoniq/item/Item.java
```diff
@@ -163,31 +163,57 @@ public static boolean checkEquality(Item v1, Item v2) {
         return compareItems(v1, v2) == 0;
     }
 
-    public abstract List<Item> getItems() throws OperationNotSupportedException;
+    public List<Item> getItems() {
+        throw new RuntimeException("Item is not an array or object.");
```
CanBerker: I think this error message should be changed to "Item is not an array" in order to not be misleading.
ObjectItem does not override getItems() method.

src/main/java/sparksoniq/jsoniq/item/Item.java
```diff
@@ -163,31 +163,57 @@ public static boolean checkEquality(Item v1, Item v2) {
...
+    };
 
-    public abstract List<String> getKeys() throws OperationNotSupportedException;
+    public List<String> getKeys() {
+        throw new RuntimeException("Item is not an object.");
+    };
 
-    public abstract Collection<? extends Item> getValues() throws OperationNotSupportedException;
+    public Collection<? extends Item> getValues() {
+        throw new RuntimeException("Item is not an array or object.");
```
CanBerker: Similar to above comment, message should be "Item is not an object."

src/main/java/sparksoniq/jsoniq/item/ObjectItem.java
```diff
@@ -77,11 +77,7 @@ else if (values.size() == 1) {
...
             else {
-                try {
-                    throw new OperationNotSupportedException("Unexpected list size found");
-                } catch (OperationNotSupportedException e) {
-                    e.printStackTrace();
-                }
+                throw new RuntimeException("Unexpected list size found");
```
CanBerker: Adding trailing "." would be nice

ghislainfourny:
Thanks, Can! All done.

PR #247, opened 2019-05-01
Reformat code by IDEA

CanBerker:
Code reformatted and imports optimized by IDEA.
Antlr generated files and input samples given in resources/queries folder are excluded


CanBerker:
Exactly this is just an automated reformatting and imports optimizing.

PR #248, opened 2019-05-01
Rearrange code by IDEA

CanBerker:
In addition to code reformatting and import optimization, code rearrangement is performed by IDEA.


CanBerker:
This PR is also fully automated but also rearranges the order the variables and functions are defined within the file to make it prettier.

PR #251, opened 2019-05-14
Implement groupby clause in SQL

CanBerker:
This PR is based on #252


src/main/resources/test_files/runtime-spark/DataFrames/GroupbyClause3.iq
```diff
@@ -0,0 +1,8 @@
...
+for $i in json-file("./src/main/resources/queries/confusion_sample.json")
+let $guess := $i."guess", $target := $i."target"
+where $guess eq $target
+group by $target
+order by $target
+return {"Language": $target,
+        "Correct guesses": count($i)}
```
CanBerker: ForLetGroup.iq renamed to this file. Result order updated. Please compare for sanity check

src/main/resources/test_files/runtime-spark/DataFrames/GroupbyClause2.iq
```diff
@@ -0,0 +1,8 @@
...
+for $i in json-file("./src/main/resources/queries/Reddit.json", 10)
+let $author := $i.author, $score := $i.score
+group by $author
+let $totalScore := sum($score)
+order by $totalScore
+return {"Author": $author,
+        "TotalScore": $totalScore}
```
CanBerker: RedditGroupBy.iq renamed to GroupbyClause2.iq   Result order updated. Please compare for sanity check

src/main/resources/test_files/runtime-spark/DataFrames/GroupbyClauseError1.iq
```diff
@@ -1,5 +1,7 @@
-(:JIQS: ShouldCrash; ErrorCode="XQST0094" :)
+(:JIQS: ShouldCrash; ErrorCode="XPTY0004" :)
 for $i in json-file("./src/main/resources/queries/conf-ex.json", 10)
 let $j := $i.choices
 group by $j := $i.choices
-return $j
\ No newline at end of file
+return $j
+
+(: non-atomic group by error :)
```
CanBerker: NonAtomic.iq renamed to GroupbyClauseError1.iq Error message updated

src/main/resources/test_files/runtime-spark/DataFrames/GroupbyClauseError2.iq
```diff
@@ -0,0 +1,6 @@
+(:JIQS: ShouldCrash; ErrorCode="XPDY0130" :)
+for $i in json-file("./src/main/resources/queries/conf-ex.json", 10)
+group by $target := $i.target
+return $target
+
+(: variable out of scope :)
```
CanBerker: This test case throws a variable not found exception shown by XPDY0130. The variable defined within the scope of group by is not accessible at the return statement. This seems correct to me but I'd appreciate a sanity check.

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
```diff
@@ -209,17 +219,194 @@ private void linearizeTuples(List<FlworTuple> keyTuplePairs) {
...
+                serializerUDFName,
+                groupbyVariableNames
+        );
+
+        df.createOrReplaceTempView("input");
+        return df.sparkSession().sql(
+                String.format(
+                        "select %s from (%s) group by `%s`",
+                        projectSQL, createColumnsSQL, appendedGroupingColumnsName
+                )
```
CanBerker: Collection and serialization of the results is performed by ProjectSQL.

In the first implementation I hadn't included a distinct operation. This resulted in duplicate values for the variables that were used in the groupby clause. To alleviate this problem, I have decided to apply a distinct operation after collecting the results for collect_list and before serializing. This distinct operation is performed only on the variables that the groupby operation is performed on. The query preparation is made in the DataFrameUtils.java file.

I'd appreciate a sanity check regarding the process of applying distinct operation on variables used in groupby

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ObjectConstructorRuntimeIterator.java
```diff
@@ -76,25 +78,23 @@ public ObjectItem next() {
...
                     //SIMILAR TO ZORBA, if value is more than one item, wrap it in an array
-                    if (currentResults.size() > 1)
+                    if (currentResults.size() > 1) {
                         values.add(new ArrayItem(currentResults));
-                    else
+                    } else if (currentResults.size() == 1) {
                         values.add(currentResults.get(0));
+                    } else {
+                        values.add(new NullItem());
+                    }
```
CanBerker: Object construction would cause exceptions for empty sequences used as values. I have added a size check and put in NullItem for empty sequences. I believe that this behavior is compliant with JSONiq specifications at:
http://www.jsoniq.org/docs/JSONiq/html-single/index.html#idm71239520

> And do not worry about the value expression: if it is is empty, null will be used as a value, and if it contains two items or more, they will be wrapped into an array.

src/main/resources/test_files/runtime-spark/DataFrames/GroupbyClauseError2.iq
ghislainfourny: This query should return successfully. Group by variables are accessible all the way to and including the return clause.

src/main/java/sparksoniq/spark/udf/GroupClauseCreateColumnsUDF.java
```diff
@@ -0,0 +1,124 @@
...
+import sparksoniq.jsoniq.runtime.iterator.primary.VariableReferenceIterator;
+import sparksoniq.semantics.DynamicContext;
+import sparksoniq.spark.DataFrameUtils;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class GroupClauseCreateColumnsUDF implements UDF1<WrappedArray, Row> {
```
ghislainfourny: Why is this class different from the UDF used for order by?

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
```diff
@@ -209,17 +219,194 @@ private void linearizeTuples(List<FlworTuple> keyTuplePairs) {
...
+        Dataset<Row> df = _child.getDataFrame(context);
+        StructType originalInputSchema = df.schema();
+        StructType inputSchema;
+        String[] columnNamesArray;
+        List<String> columnNames;
+
+        List<VariableReferenceIterator> variableAccessExpressions = new ArrayList<>();
+        for (int expressionIndex = 0; expressionIndex < _expressions.size(); expressionIndex++) {
+            GroupByClauseSparkIteratorExpression expression = _expressions.get(expressionIndex);
+            inputSchema = df.schema();
```
ghislainfourny: I can live with this because of "correctness first" -- could you make a GitHub issue for the future that it would be a nice performance improvement to be able to add several variables in one shot (for group by and let clauses)?

src/main/java/sparksoniq/spark/udf/GroupClauseDetermineTypeUDF.java
```diff
@@ -0,0 +1,101 @@
...
+import sparksoniq.exceptions.UnexpectedTypeException;
+import sparksoniq.jsoniq.item.Item;
+import sparksoniq.jsoniq.runtime.iterator.primary.VariableReferenceIterator;
+import sparksoniq.semantics.DynamicContext;
+import sparksoniq.spark.DataFrameUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class GroupClauseDetermineTypeUDF implements UDF1<WrappedArray, List> {
```
ghislainfourny: Why is this UDF different from the UDF used for order by?

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
ghislainfourny: I think this is correct. If I correctly remember, we even discussed doing so in our last meeting.

src/main/resources/test_files/runtime-spark/DataFrames/GroupbyClause2.iq
ghislainfourny: Do we understand why the order changed? Is it because we take over the order that Spark SQL's ORDER BY imposes?

CanBerker: I had checked earlier and used an online text comparison tool to be sure. The results seems to be same, just in a different order. I would assume it is caused by implementation difference.
image

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
CanBerker: Issue created

src/main/resources/test_files/runtime-spark/DataFrames/GroupbyClauseError2.iq
CanBerker: Correct behavior implemented

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
CanBerker: I'm glad to hear that

src/main/java/sparksoniq/spark/udf/GroupClauseCreateColumnsUDF.java
CanBerker: At the beginning of my work I was passing GroupBy/Orderby expressionsIterators which required different types. Since that is not the case for quite some time I don't see a reason why these two UDFs can't be merged together to something like "AggregateClause...functionality...UDF.java". I will be making the necessary changes next

src/main/java/sparksoniq/spark/udf/GroupClauseDetermineTypeUDF.java
CanBerker: I think the determineTypeUDFs are more or less identical. If desired I think these two can be merged into one. In that case we will lose some precision with the error messages or need to pass the current operation type (group vs aggregate) as a parameter. We can discuss these two points further in this week's meeting

src/main/java/sparksoniq/spark/udf/GroupClauseDetermineTypeUDF.java
```diff
@@ -0,0 +1,101 @@
...
+            } else if (_nextItem.isString()) {
+                result.add("string");
+            } else if (_nextItem.isInteger()) {
+                result.add("integer");
+            } else if (_nextItem.isDouble()) {
+                result.add("double");
+            } else if (_nextItem.isDecimal()) {
+                result.add("decimal");
+            } else if (_nextItem.isArray() || _nextItem.isObject()) {
+                throw new UnexpectedTypeException("Group by column must contain only atomics.", expression.getMetadata());
```
ghislainfourny: Encapsulate error message.

src/main/java/sparksoniq/spark/udf/GroupClauseDetermineTypeUDF.java
```diff
@@ -0,0 +1,101 @@
...
+            for (int columnIndex = 0; columnIndex < _inputColumnNames.size(); columnIndex++) {
+                _context.addVariableValue(_inputColumnNames.get(columnIndex), _deserializedParams.get(columnIndex));
+            }
+
+            // apply expression in the dynamic context
+            expression.open(_context);
+            if (expression.hasNext()) {
+                _nextItem = expression.next();
+                if (expression.hasNext()) {
+                    throw new SparksoniqRuntimeException("Group by expressions must return a singleton for each row");
```
ghislainfourny: Check that any grouping variable bound to several items gets cleanly caught at some other place by adding a test (let $a := 1 group by $b := (1, 2) return $a).

CanBerker:
Error messages encapsulated to convey only relevant information to the user.

Grouping and ordering on sequences with multiple items are prevented for Dataframe implementation.

PR #252, opened 2019-05-14
Object constructor

CanBerker:
Object construction would cause exceptions for empty sequences used as values. I have added a size check and put in NullItem for empty sequences. I believe that this behavior is compliant with JSONiq specifications at:
http://www.jsoniq.org/docs/JSONiq/html-single/index.html#idm71239520

> And do not worry about the value expression: if it is is empty, null will be used as a value, and if it contains two items or more, they will be wrapped into an array. 

Error codes added to exceptions thrown within Item class

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ObjectConstructorRuntimeIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ObjectConstructorRuntimeIterator.java
@@ -22,6 +22,7 @@ package sparksoniq.jsoniq.runtime.iterator.primary;
 import sparksoniq.exceptions.IteratorFlowException;
 import sparksoniq.jsoniq.item.ArrayItem;
 import sparksoniq.jsoniq.item.Item;
+import sparksoniq.jsoniq.item.NullItem;
 import sparksoniq.jsoniq.item.ObjectItem;
 import sparksoniq.jsoniq.item.StringItem;
 import sparksoniq.jsoniq.item.metadata.ItemMetadata;
@@ -45,6 +46,7 @@ public class ObjectConstructorRuntimeIterator extends LocalRuntimeIterator {
         this._keys = keys;
         this._values = values;
     }
+
     public ObjectConstructorRuntimeIterator(List<RuntimeIterator> childExpressions,
                                             IteratorMetadata iteratorMetadata) {
         super(null, iteratorMetadata);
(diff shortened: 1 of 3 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ObjectConstructorRuntimeIterator.java
```diff
@@ -75,25 +77,23 @@ public ObjectItem next() {
...
                 }
 
                 for (RuntimeIterator keyIterator : this._keys) {
                     keyIterator.open(this._currentDynamicContext);
-                    try {
-                        keys.add(((StringItem) keyIterator.next()).getStringValue());
-                    } catch (Exception ex) {
-                        throw new IteratorFlowException("Object must have string keys!", getMetadata());
-                    }
+                    keys.add((keyIterator.next()).getStringValue());
```
ghislainfourny: Correct use of the item API is that one needs to check that the key is a string by calling isString() here explicitly, and throw the UnexpectedTypeException here in case it is not.

The contract of use of the item API, is that getStringValue() may only be called after calling isString() and its returning true. The runtime exceptions in the item API are internal to the development (they are not supposed to happen, ever -- if they ever do it means that there is a bug in our code) and are not meant be (ab)used to throw external type errors.

This means that the renaming of RuntimeException to UnexpectedTypeException should be reverted.

Thanks a lot!

CanBerker: That makes a lot of sense. I will be updating the code accordingly. Thanks for the explanation!

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ObjectConstructorRuntimeIterator.java
```diff
@@ -75,27 +77,32 @@ public ObjectItem next() {
...
                 }
 
                 for (RuntimeIterator keyIterator : this._keys) {
                     keyIterator.open(this._currentDynamicContext);
-                    try {
-                        keys.add(((StringItem) keyIterator.next()).getStringValue());
-                    } catch (Exception ex) {
-                        throw new IteratorFlowException("Object must have string keys!", getMetadata());
+                    if (!keyIterator.hasNext()) {
+                        throw new IteratorFlowException("Key iterator must provide an item for object creation", getMetadata());
```
ghislainfourny: Is this message visible outside? If so, it is more data-independent to say "A key cannot be the empty sequence".

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ObjectConstructorRuntimeIterator.java
```diff
@@ -75,27 +77,32 @@ public ObjectItem next() {
...
+                    }
+                    Item key = keyIterator.next();
+                    if (!key.isString()) {
+                        throw new UnexpectedTypeException("Key provided for object creation must be of type String", getMetadata());
                     }
+                    keys.add(key.getStringValue());
                     if (keyIterator.hasNext())
-                        throw new IteratorFlowException("Object value must return one item!", getMetadata());
+                        throw new IteratorFlowException("Key iterator must provide at most one item to be used as key", getMetadata());
```
ghislainfourny: Is this message visible outside? If so, it is more data-independent to say "A key cannot be a sequence of more than one item".

PR #255, opened 2019-05-19
[WIP] Fix orderby and groupby implementations

CanBerker:
Implementations of orderby and groupby clauses work with sequences which should be throwing a TypeMismatchError as implemented in the Dataframe extension (#251 )

Groupby implementation doesn't throw an error when the grouping variable contains different types. TypeMismatchError should be thrown


ghislainfourny:
I think that this is now superseded by the DataFrame support.

CanBerker:
Correct, this was to be closed when DataFrames are merged

PR #256, opened 2019-05-20
Fix keySet issue with older org.json.

ghislainfourny:
The only material fix (besides upgrading version to 0.9.7) is to no longer use JSONObject.keySet(), because some Spark environments have old versions of org.json. This fix makes us backward compatible with all environments.


CanBerker:
This looks good to me. I have also double checked the documentation and leaving the link below for reference:
http://www.docjar.com/docs/api/org/json/JSONObject.html#length
public int length() 
    Get the number of keys stored in the JSONObject.

ghislainfourny:
Thanks! Also, length existed from the very first commit of org.json, so it's safer :)

PR #257, opened 2019-05-23
Distinct values rdd

ghislainfourny:


ghislainfourny:
@wscsprint3r this makes distinct-values() work on big RDDs. The distinct() transformation needs equals() and hashCode() on the Items.

The hash isn't quite perfect, but at least makes sure equal items have the same hash. Equality on object/array is implemented as a deep equal, which will be useful for the deep-equal function later.

Currently equality comparison slightly overlaps with the already implemented compareItems() (which compare atomics), but we can clean this up later by implementing the Java compare() method as well and redirecting to these builtin methods.

ghislainfourny:
Can volunteered to review as he has resources and already took a look so I am transferring the review to him (but of course, it also doesn't hurt if you look at it, Stefan).

ghislainfourny:
Thanks, Stefan!

PR #261, opened 2019-05-27
Cache intermediate view.

ghislainfourny:
For ordering queries, the same DF was computed three times. This activates caching.


src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
```diff
@@ -209,6 +209,7 @@ private void setAllLocalResults() {
         String udfSQL = DataFrameUtils.getSQL(inputSchema, -1, false);
 
         df.createOrReplaceTempView("input");
+        df.sparkSession().table("input").cache();
```
CanBerker: This is very nice, can't we apply the same idea to the groupby clause?
Cache after line 288, remove table creation at line 384

ghislainfourny: Yes, group by is next!

PR #270, opened 2019-06-03
Implement starts-with

ghislainfourny:


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/strings/StartsWithFunctionIterator.java
```diff
@@ -0,0 +1,66 @@
...
+                StringItem.class);
+            if (stringItem == null || stringItem.getStringValue().isEmpty()) {
+                return new BooleanItem(false);
+            }
+            boolean result = stringItem.getStringValue().startsWith(
+                substringItem.getStringValue());
+            return new BooleanItem(result);
+        } else
+            throw new IteratorFlowException(
+                RuntimeIterator.FLOW_EXCEPTION_MESSAGE + " substring function",
```
CanBerker: Is it intentional to use "substring function" in the error message? It could also be "starts with function"

ghislainfourny: Good catch, thanks! :)

PR #272, opened 2019-06-03
Cache intermediate DF in group by

ghislainfourny:
Dear Can,

This is the caching for group by. Does it make a difference in the measurements?


CanBerker:
It seems that the query-5-df has gone down https://github.com/CanBerker/sparksoniq-dataframe-perf-v1/pull/3/files. The fluctuations in the results at each run seem to be noticable though. I'm thinking of running the measurements around 5 times and averaging the results to get better estimates of the actual behavior. Are you planning any further performance improvements?

ghislainfourny:
Thanks a lot! Same here, I had a difference around 10%.

Yes, further improvements are in the works, like projecting away unused variables (see other PR).

PR #275, opened 2019-06-04
Variable dependencies

ghislainfourny:
Adding a function to the runtime iterator APIs to get the list of variables that any expression or clauses depends on via the dynamic context. Based on this, we will be able, in the new DataFrames implementation, to only pass those variables that are actually needed to UDFs, and to project away columns that are not needed by subsequent clauses.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/tupleiterator/RuntimeTupleIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/tupleiterator/RuntimeTupleIterator.java
@@ -101,6 +101,16 @@ public abstract class RuntimeTupleIterator implements RuntimeTupleIteratorInterf
 
     public abstract JavaRDD<FlworTuple> getRDD(DynamicContext context);
 
+    /*
+    * Variable dependencies are variables that MUST be provided in the dynamic context
+    * for successful execution.
+    * 
+    * These variables are:
+    * 1. All variables that the expression of the clause depends on (recursive call of getVariableDependencies on the expression)
+    * 2. Except those variables bound in the current FLWOR (obtained from the auxiliary method getVariablesBoundInCurrentFLWORExpression), because those are provided in the Tuples
+    * 3.Plus (recursively calling getVariableDependencies) all the Variable Dependencies of the child clause if it exists.
+    * 
+    */
     public Set<String> getVariableDependencies()
     {
         Set<String> result = new HashSet<String>();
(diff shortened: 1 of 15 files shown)


ghislainfourny:
I added the --print-iterator-tree CLI parameter. If you add --print-iterator-tree yes then instead of executing the query it will print the query plan, including variable dependencies.

src/main/java/sparksoniq/jsoniq/runtime/tupleiterator/RuntimeTupleIterator.java
```diff
@@ -96,5 +101,44 @@ public IteratorMetadata getMetadata() {
...
     public abstract JavaRDD<FlworTuple> getRDD(DynamicContext context);
 
+    public Set<String> getVariableDependencies()
+    {
+        Set<String> result = new HashSet<String>();
+        result.addAll(_child.getVariableDependencies());
+        return result;
+    }
 
+    public Set<String> getBoundVariables()
```
CanBerker: I had the intuition that BoundVariables here meant FLWOR variables and your recent addition of print function also confirms that with its style. However, my initial worry was, what if variables were defined somewhere else? This would cause a bit of confusion since the function's name suggest it gets all variable bindings in the application. Even though we don't  support it at the moment, I have noticed global variable definitions can be implemented further down the road (http://www.jsoniq.org/docs/JSONiq/html-single/index.html#idm16811680).

I would recommend indicating that BoundVariables here are actually BoundFlworVariables to bring more clarity and ease of extensibility down the road.

src/main/java/sparksoniq/spark/iterator/flowr/CountClauseSparkIterator.java
```diff
@@ -109,4 +111,30 @@ private void setNextLocalTupleResult() {
                 .mapValues(index -> index + 1)
                 .map(new CountClauseClosure(variableName, getMetadata()));
     }
+
+    public Set<String> getVariableDependencies()
+    {
+        Set<String> result = new HashSet<String>();
+        result.addAll(_child.getVariableDependencies());
```
CanBerker: I just wanted to point out that the following pattern is not applied to the CountClause. Why are the child's bound variables not removed here?
        if(_child != null)
        {
            result.removeAll(_child.getBoundVariables());
            result.addAll(_child.getVariableDependencies());
        }

Additionally, could you elaborate on the algorithm here a bit?

PR #276, opened 2019-06-04
Ends with

ghislainfourny:


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/strings/EndsWithFunctionIterator.java
```diff
@@ -0,0 +1,66 @@
...
+                StringItem.class);
+            if (stringItem == null || stringItem.getStringValue().isEmpty()) {
+                return new BooleanItem(false);
+            }
+            boolean result = stringItem.getStringValue().endsWith(
+                substringItem.getStringValue());
+            return new BooleanItem(result);
+        } else
+            throw new IteratorFlowException(
+                RuntimeIterator.FLOW_EXCEPTION_MESSAGE + " substring function",
```
CanBerker: Same comment as starts with. Is "substring function" here the intended error message?

PR #277, opened 2019-06-05
Implement contains().

ghislainfourny:


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java~
```diff
@@ -0,0 +1,521 @@
+/*
```
CanBerker: I think a temporary file has gotten mixed into the commit. This file should be removed and only Functions.java should remain

PR #288, opened 2019-06-25
Upgrade Kryo and deactivate references.

ghislainfourny:
A lot of the time was spent on Kryo references. This PR deactivates them, improving performance.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,7 @@
         <dependency>
             <groupId>com.esotericsoftware</groupId>
             <artifactId>kryo</artifactId>
-            <version>5.0.0-RC2</version>
+            <version>5.0.0-RC4</version>
         </dependency>
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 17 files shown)


ghislainfourny:
Hi @wscsprint3r , would you be willing to review this PR and the other one? It is a merge into develop-dataframes, which reimplements FLWOR on top of data frames. This PR makes it faster by deactivating references on kryo.

CanBerker:
Dear @ghislainfourny 
The documentation specifies that the references are disabled by default (link below). I believe it still adds more clarity to explicitly set them as false. But just out of curiosity, were they appearing to be enabled in our code's execution?
https://github.com/EsotericSoftware/kryo#references

ghislainfourny:
Thanks, Stefan.

Yes, this was figured out with a profiler. It was visible that serialization was the one bottleneck, and half of the serialization was spent in references.

I checked that there was a performance gain with this change -- not immense, but still enough for serialization not to be the bottleneck anymore (not it is on deserialization -- partly addressed with the other PR).

PR #289, opened 2019-06-28
Serialization pool

ghislainfourny:


ghislainfourny:
Hi Stefan,

This change makes basic selection queries (like for let where return) faster because the object serializer remembers the last serialized object and its serialized bytes, and reuses it directly upon deserialization if the bytes match (and otherwise deserializes it normally).

Query 1 in our benchmark is streaming in this sense.

For queries without grouping or ordering, the whole FLWOR expression is evaluated in a streaming fashion, one object at a time, so that this makes the parsing it much faster, and otherwise not (or unnoticeably) slower.

ghislainfourny:
Thanks!

PR #290, opened 2019-07-02
More informative output in case of general error.

ghislainfourny:
When an unexpected error happened, nothing was shown on the CLI. This was making debugging less easy.

### DIFF
--- a/src/main/java/sparksoniq/Main.java
+++ b/src/main/java/sparksoniq/Main.java
@@ -66,7 +66,8 @@ public class Main {
                 System.out.println("spark-submit --master local[*] spark-rumble-1.0.jar --query-path my-query.jq --output-path my-output.json --log-path my-log.txt");
             }
         } catch (Exception ex) {
-            throw new CliException(ex.getMessage());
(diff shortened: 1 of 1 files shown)


src/main/java/sparksoniq/Main.java
```diff
@@ -66,7 +66,9 @@ public static void main(String[] args) throws IOException {
                 System.out.println("spark-submit --master local[*] spark-rumble-1.0.jar --query-path my-query.jq --output-path my-output.json --log-path my-log.txt");
             }
         } catch (Exception ex) {
-            throw new CliException(ex.getMessage());
+            System.out.println("An error has occured: " + ex.getMessage());
+            System.out.println("We should investigate this. Please contact us or file an issue on GitHub with your query.");
```
CanBerker: I would recommend placing the link as well when referencing GitHub. It will provide ease of use as both some terminals support redirecting from such links and finding the repo for the first time might not be so easy for some users.

ghislainfourny:
Thanks a lot! done and done.

PR #293, opened 2019-07-08
Make order by faster

ghislainfourny:
- introducing same universal columns as for group by
- using an accumulator to check for type errors, to avoid superfluous jobs

The performance gain is disappointing though, so I am keeping on hold for more work (but you can still take a look and comment).


wscsprint3r:
@ghislainfourny are we still moving forward with this one?

ghislainfourny:
@wscsprint3r This one is on hold, for now, for two reasons (i) I am disappointed by the performance improvement, which is not very noticeable (even though I'd have another idea to improve further) and (ii) this change may not be forward compatible with the upcoming type support for dates, times, binaries, etc, which make lead to an explosion in the number of newly created columns.

PR #294, opened 2019-07-10
Refactor static methods of Item class

CanBerker:
I have kept 'getNumericResultType' as a static method as I felt it semantically lives on the Item class rather than a specific object

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/item/Item.java
+++ b/src/main/java/sparksoniq/jsoniq/item/Item.java
@@ -37,8 +37,8 @@ public abstract class Item implements SerializableItem {
     protected Item() {
     }
 
-    public static boolean isNumeric(Item item) {
-        return item.isInteger() || item.isDecimal() || item.isDouble();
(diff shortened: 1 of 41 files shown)


src/main/java/sparksoniq/jsoniq/item/Item.java
```diff
@@ -87,25 +87,23 @@ public static Type getNumericResultType(Item left, Item right) {
     }
 
     //returns an effective boolean value of any item type
-    public static boolean getEffectiveBooleanValue(Item item) {
-        if (item == null)
+    public boolean getEffectiveBooleanValue() {
```
ghislainfourny: Note if you want to do this even better: this method could actually be made abstract and overriden for each subclass.

CanBerker:
These are both very good suggestions. I have implemented getEffectiveBooleanValue as you have suggested. In the case of getNumericValue, I opted for throwing an exception in the base class and overriding the method in Numeric subclasses.

PR #301, opened 2019-07-29
Optimize group count

ghislainfourny:
This makes the grouping query much faster (down to 60+ seconds from 100+) by providing, in getDataFrame, information on what can be projected away. That way, the group clause, when only a count aggregate is needed, does not need to serialize big arrays of items (those from each group) any more: only the count is included in the SQL table.

Variable dependencies are extended to carry, with the needed variables, also info on what is needed: FULL sequence, COUNT, SUM, etc.

The Dynamic context has been refactored in order to be able to support queries that only ask for a count by not materializing the underlying sequence.

Each FLWOR clause has a new method that allows it to say what projection it needs from its child clause.

Currently, only the group by clause is capable of optimizing away, but in the future this opens to door to many further optimizations, as this modification is generic.


src/main/java/sparksoniq/jsoniq/runtime/tupleiterator/RuntimeTupleIterator.java
```diff
@@ -104,22 +107,32 @@ public IteratorMetadata getMetadata() {
...
+    /*
+     * The second parameter gives information on the projection needed by the caller.
+     * Currently, this is used to inform a group by clause that only a count is needed for a specific variable, which allows projecting away the actual items.
+     */
+    public abstract Dataset<Row> getDataFrame(DynamicContext context, Map<String, DynamicContext.VariableDependency> parentProjection);
+
+    /*
+     * Returns the projection needed by this iterator. The result of this method is forwarded to the child clause so it can optimize some values away.
+     */
+    public abstract Map<String, DynamicContext.VariableDependency> getProjection(Map<String, DynamicContext.VariableDependency> parentProjection);
```
CanBerker: I really like that these explanations exist.  It can be merged as it is but I have tiny remarks if you'd want to re-iterate over them.

The wording can be more concise and less subject to change (ie. stating '2. parameter' can potentially be misleading after any refactoring involving parameters) . Additionally, javadoc style can be considered.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -314,4 +320,30 @@ public void print(StringBuffer buffer, int indent)
...
     }
+    
+    public Map<String, DynamicContext.VariableDependency> getProjection(Map<String, DynamicContext.VariableDependency> parentProjection)
+    {
+        if(_child == null)
+        {
+            return null;
+        }
+
+        // passing dependencies to parent
```
CanBerker: I initially found this comment to be a bit confusing. I believe this refers to line 335 and following lines?

I would suggest, if possible, extracting such a dependency calculation inside the getProjection function into a separate function, rather than relying on a comment

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
```diff
@@ -383,7 +394,44 @@ public void print(StringBuffer buffer, int indent)
...
+        {
+            projection.remove(iterator.getVariableReference().getVariableName());
+        }
+        for(GroupByClauseSparkIteratorExpression iterator : _expressions)
+        {
+        	if(iterator.getExpression() == null)
+        	{
+        		String k = iterator.getVariableReference().getVariableName();
+                projection.put(k, DynamicContext.VariableDependency.FULL);
+                continue;
```
CanBerker: I believe k, here represents a 'key', I would prefer the explicit name rather than a single character.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
ghislainfourny: Good catch. This was a leftover before I cleaned up the API.

ghislainfourny:
Very good, thorough and useful comments. Many thanks! I have addressed everything.

src/main/java/sparksoniq/jsoniq/runtime/tupleiterator/RuntimeTupleIterator.java
ghislainfourny: Yes, javadoc is what should be done.

PR #303, opened 2019-08-07
Stick to Trusty distribution in Travis to support Java 8

ghislainfourny:


ghislainfourny:
Travis dropped support for Java 8. Moving to Java 11 LTS that Spark seems to now support.

wscsprint3r:
@ghislainfourny do we still have a cluster somewhere where we can run this and make sure it works?

wscsprint3r:
Maybe we need to package some extra dependencies or something...

ghislainfourny:
Hi Stefan,

Yes, there is AWS EMR that can be used for this. We can simply deploy a cluster on demand. I recently tested that DataFrames and Spark SQL work well with our jar there.

The DSO is no longer available, unfortunately.

Kind regards,
Ghislain

PR #304, opened 2019-08-08
Fix a lot of warnings.

ghislainfourny:


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArraySizeFunctionIterator.java
```diff
@@ -34,7 +34,11 @@
 public class ArraySizeFunctionIterator extends LocalFunctionCallIterator {
 
 
-    private RuntimeIterator arrayIterator;
+    /**
```
wscsprint3r: Is this intentional?

ghislainfourny: This was done automatically by Eclipse by using the bulk quick fix option on the missing serialization IDs. Given the very large number of files that need editing, I think the efforts required to manually remove these empty comments would outweigh the benefits. What do you think?

wscsprint3r:
Hang on, I accidentally commited the parser files...

PR #319, opened 2019-08-22
Public api

ghislainfourny:
This is the first, basic version of the Public API for those who want to use it via maven.

You can see the javadoc here: http://rumbledb.org/site/apidocs/org/rumbledb/api/package-summary.html

I moved over the Item class to this API and documented it with Javadoc. Most files modified are only a change of the imported package for the Item class.
 
In the future, we'll allow passing input to the dynamic context, and in the more remote future, binding an input variable to an RDD provided by the user.


ghislainfourny:
If you try to generate the javadocs with mvn javadoc:javadoc, there will be an error about the builtin Java modules, which is a bug of the maven javadoc plugin. But the files still get generated properly.

src/main/java/org/rumbledb/api/Item.java
```diff
@@ -109,96 +119,210 @@ public int compareTo(Item other) {
...
+	 */
     public boolean isAtomic() {
         return false;
     }
 
-    public boolean isNumber() {
-        return false;
-    }
-
+    /**
```
pjfanning: minor issue with tabs appearing the javadoc

ghislainfourny: Good catch, this is fixed.

ghislainfourny:
Thanks!

I think we can release the public API publicly (as a minor update Rumble 1.1.1) together with the maven central entry once you are ready.

PR #321, opened 2019-08-30
Publish to Maven

wscsprint3r:


ghislainfourny:
@wscsprint3r this seems to not pass the test queue. Do we know why?

wscsprint3r:
@ghislainfourny I will fix it, it's probably missing credentials for the new Maven steps, but we probably don't want those integrated into the CI process.

wscsprint3r:
@ghislainfourny I will check and see, maybe we can create a shared key for gpg signing and also have credentials for you and @CanBerker so you can access the repository. For now I can share my credentials with you.

ghislainfourny:
Thanks, Stefan! I also checked and it seems fine to use GPG to sign -- as long as we do not copy over any of its code, nor link it into our jar.

PR #323, opened 2019-10-02
Instance of operator

AndreaRinaldi1:
This branch contains the implementation of **instance of** operator (both expression and iterator).

I had to fix a few checks, like the one assigning the correct arity (question/star/plus) to _FlworVarSequenceType_ and I also implemented a helper function _getResultIfEmpty()_ that allows RuntimeIterators to decide what to return in case the result of calling (or better, the result of **not** calling) _next()_ on them, is null. That would happen for example in _() instance of string_ when _next()_ would not be called because _hasNext()_ returns false right away.

Tests as well are green.


src/main/java/sparksoniq/JsoniqQueryExecutor.java
```diff
@@ -255,7 +255,7 @@ private String getIteratorOutput(RuntimeIterator iterator) {
             result = iterator.next();
         }
         if (result == null) {
-            return "";
+            return iterator.getResultIfEmpty().serialize();
```
ghislainfourny: If result == null, then it means the iterator returned the empty sequence, which should always be serialized as an empty string.

You probably had to make this change because of a (pre-existing) mistake in the implementation of instance of : if the left hand side is (), then false should be returned by the iterator and not the empty sequence.

The current implementation of the iterator probably incorrectly returns () instead (see comment in the instance of iterator code). If you fix the iterator, the getResultIfEmpty function is no longer necessary.

src/main/java/sparksoniq/jsoniq/compiler/parser/Jsoniq.g4
```diff
@@ -168,7 +169,7 @@ keyWordBoolean
 
 atomicType
          : 'atomic' | 'string' | 'integer'
-         | 'decimal' | 'double' | keyWordBoolean | 'null';
+         | 'decimal' | 'double' | keyWordBoolean | NullLiteral;
```
ghislainfourny: Very good. Neat and clean fix.

src/main/java/sparksoniq/jsoniq/item/ArrayItem.java
```diff
@@ -73,9 +73,7 @@ public int getSize() {
 
     @Override
     public boolean isTypeOf(ItemType type) {
-        if (type.getType().equals(ItemTypes.ArrayItem) || super.isTypeOf(type))
-            return true;
-        return false;
+        return type.getType().equals(ItemTypes.ArrayItem) || super.isTypeOf(type);
```
ghislainfourny: Very good. To take it to the next level in terms of code readability, you could also avoid the super.isTypeOf recursive call and "flatten out" the tests in each class.

But this is mostly cosmetic as the super call will anyway probably be inlined by the optimizer. So I can live with the current implementation, which does the job.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/InstanceOfIterator.java
```diff
@@ -52,12 +52,17 @@ public Item next() {
                 items.add(_child.next());
             _child.close();
             this._hasNext = false;
+
+            //... instance of ()
+            if (_sequenceType.isEmpty()) return ItemFactory.getInstance().createBooleanItem(false);
```
ghislainfourny: Could we also rename isEmpty() to isEmptySequence() to make sure there is no confusion?

ghislainfourny: There is a missing test for whether items is empty. If items is empty and _sequenceType.isEmpty() is true, then the boolean item true should be returned. If items is non-empty and _sequenceType.isEmpty() is true, then the boolean item false should be returned.

For some reason that I don't understand just with a quick look, it seems that something odd happens when the input sequence is empty at the moment; fixing this will make getResultIfEmpty() superfluous.

src/main/resources/test_files/runtime/InstanceOf2.jq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="true":)
+() instance of string?
+
```
ghislainfourny: I assume that you separated these tests because of () on the left hand side. If you fix the instance of iterator to return a false boolean item in next(), this separation will no longer be needed as instance of will correctly return false/true even even if the instance of expression is nested in the query.

src/main/java/sparksoniq/JsoniqQueryExecutor.java
AndreaRinaldi1: _I will respond here to this and the following comments about the issue with the empty list._

From the second revision, i.e. commit 042b902, I removed the function _getResultIfEmpty()_ since, as @ghislainfourny rightfully said, it was superfluous. I had to change the behavior in the _open()_ function of the parent UnaryOperationIterator because it was setting **this._hasNext** to false when **child.hasNext()** was false (and EmptySequenceIterator always return false for hasNext() ). Now the check for __child.hasNext()_ happens in the _next()_ function for both UnaryOperationIterator and InstanceOfIterator.

src/main/java/sparksoniq/jsoniq/item/ArrayItem.java
AndreaRinaldi1: I think I would prefer to leave the implementation that relies on inheritance from the defined type hierarchy, rather than explicitly flattening out the different super types. I think it is cleaner and easier to understand the concept of what is being checked. Open to discussion though.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/InstanceOfIterator.java
AndreaRinaldi1: I am not sure I understood correctly what test would be missing. I added all the possible combinations of:

- x instance of y
- x instance of ()
- () instance of y
- () instance of y?
- () instance of ()

with x != () and y != ().

src/main/resources/test_files/runtime/InstanceOf2.jq
AndreaRinaldi1: Correct. 
Now, with the fixed version of InstanceOfIterator, all the tests are in a single file.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/UnaryOperationIterator.java
```diff
@@ -48,9 +48,13 @@ public Item next() {
         if (this._hasNext) {
             this._hasNext = false;
             _child.open(_currentDynamicContext);
-            Item child = _child.next();
+            Item child = null;
```
ghislainfourny: Perfect. I think it is now indeed cleaner, as the first hasNext() value depends on the semantics of the unary iterator.

Could you please add two tests

-() (returns ())
(1, -(), 1) (returns (1,1))

to check that the semantics of the unary minus operator (which does, unlike "instance of", return the empty sequence if the child returns the empty sequence) has no regression in spite of this change in the code?

Also, () instance of string* and () instance of string+ would be good corner case tests.

Thanks a lot! Then it's good to go. I'll mark as approved and you can merge after adding the above tests and checking they work.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/InstanceOfIterator.java
ghislainfourny: I think that it's all good. Only + and * were missing (see my other comment).

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/UnaryOperationIterator.java
```diff
@@ -48,9 +48,13 @@ public Item next() {
...
+            if (_child.hasNext()) {
+                child = _child.next();
+            }
             _child.close();
-
+            if (child == null)
+                return null;
```
ghislainfourny: One small but important thing here: next() should never return null, but throw an error in unexpected cases. Under normal operation, next() must return a non-null Item if hasNext() returned true, and it should not have been invoked in the first place if hasNext() returned false (e.g., for empty sequence or after the last item has been consumed).

If null is returned rather than throwing an IteratorFlowException, this may "hide" any errors in the code under the carpet, leading to bugs later on.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/InstanceOfIterator.java
AndreaRinaldi1: From revision 3, i.e. commit 16e34f9, I added those tests as well.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/UnaryOperationIterator.java
AndreaRinaldi1: I now added the tests for UnaryOperation for empty sequence as well.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/UnaryOperationIterator.java
AndreaRinaldi1: Thank you for spotting this. 
From revision 3, i.e. commit 16e34f9, I reverted the changes that I previously made to UnaryOperationIterator and now InstanceOfIterator extends UnaryOperationBaseIterator so that calling open() on an InstanceOfIterator object will use RuntimeIterator open() function and not the one overriding it in UnaryOperationIterator.

AndreaRinaldi1: As for creating branches from the main rumbleDB repository, sure, I will do. I just forked because initally I was not Contributor to the main repo. 

As soon as you approve the most recent changes, I will proceed with the merge.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/UnaryOperationIterator.java
```diff
@@ -75,11 +75,7 @@ public void open(DynamicContext context) {
         super.open(context);
 
         _child.open(_currentDynamicContext);
-        if (_child.hasNext()) {
-            this._hasNext = true;
-        } else {
-            this._hasNext = false;
-        }
+        this._hasNext = _child.hasNext();
```
ghislainfourny: Nice compression.

PR #325, opened 2019-10-09
Castable as operator

AndreaRinaldi1:
This branch includes the implementation of the **castable as** operator (iterator and expression).

From the W3C documentation:

> The expression E castable as T returns true if the result of evaluating E can be successfully cast into the target type T by using a cast expression; otherwise it returns false. If evaluation of E fails with a dynamic error, the castable expression as a whole fails.

To keep the naming of types consistent with what W3C is using, I introduced _simpleType_ and _singleType_ in Jsoniq.g4, together with the new Classes _AtomicType_, _FlworVarSingleType_ and the enum _AtomicTypes_. 

I forgot to add _NullItem_ to _AtomicTypes_ Enum, but I will do in the next version of the pr.
There is just one small naming inconsistency, that is _FlworVarSingleType_ has a _getAtomicTypes()_method, which will actually return only single types since we do not return NullItem (as NullItem is not a SingleType). I prefer this solution to adding a new Enum _SingleTypes_ that would be identical to _AtomicTypes_ except for NullItem (subclassing is not possible for Enums). Open to discussion, though.

Tests are green.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
@@ -36,6 +36,7 @@ import sparksoniq.jsoniq.compiler.translator.expr.flowr.CountClause;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworClause;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworExpression;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworVarSequenceType;
+import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworVarSingleType;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.ForClause;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.ForClauseVar;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.GroupByClause;
@@ -48,6 +49,7 @@ import sparksoniq.jsoniq.compiler.translator.expr.flowr.ReturnClause;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.WhereClause;
 import sparksoniq.jsoniq.compiler.translator.expr.operational.AdditiveExpression;
 import sparksoniq.jsoniq.compiler.translator.expr.operational.AndExpression;
+import sparksoniq.jsoniq.compiler.translator.expr.operational.CastableExpression;
 import sparksoniq.jsoniq.compiler.translator.expr.operational.ComparisonExpression;
 import sparksoniq.jsoniq.compiler.translator.expr.operational.InstanceOfExpression;
 import sparksoniq.jsoniq.compiler.translator.expr.operational.MultiplicativeExpression;
@@ -75,6 +77,7 @@ import sparksoniq.jsoniq.compiler.translator.expr.primary.VariableReference;
 import sparksoniq.jsoniq.compiler.translator.expr.quantifiers.QuantifiedExpression;
 import sparksoniq.jsoniq.compiler.translator.expr.quantifiers.QuantifiedExpressionVar;
 import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
+import sparksoniq.semantics.types.AtomicTypes;
 import sparksoniq.semantics.types.ItemTypes;
 import sparksoniq.semantics.types.SequenceType;
 
@@ -574,11 +577,11 @@ public class JsoniqExpressionTreeVisitor extends sparksoniq.jsoniq.compiler.pars
 
     @Override
     public Void visitTreatExpr(JsoniqParser.TreatExprContext ctx) {
-        UnaryExpression mainExpression;
+        CastableExpression mainExpression;
         FlworVarSequenceType sequenceType;
         TreatExpression node;
         this.visitCastableExpr(ctx.mainExpr);
-        mainExpression = (UnaryExpression) this.currentExpression;
+        mainExpression = (CastableExpression) this.currentExpression;
         if (ctx.seq != null && !ctx.seq.isEmpty()) {
             JsoniqParser.SequenceTypeContext child = ctx.seq;
             this.visitSequenceType(child);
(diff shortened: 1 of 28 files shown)


src/main/java/sparksoniq/jsoniq/compiler/parser/Jsoniq.g4
```diff
@@ -167,9 +167,15 @@ jSONItemTest
...
+singleType
+         : item=simpleType (question +='?')?;
+
+simpleType
          : 'atomic' | 'string' | 'integer'
-         | 'decimal' | 'double' | keyWordBoolean | NullLiteral;
+         | 'decimal' | 'double' | keyWordBoolean;
+
+atomicType
+         : simpleType | NullLiteral;
```
ghislainfourny: Actually null behaves like any other atomic type: it is a single type mapping the literal "null" to null, and null("null") returns null, "null" castable as null returns true. This was done so as not to introduce any special treatment for it. There is no need to introduce any special treatment by making the grammar more complicated and splitting atomicType in this respect.

src/main/java/sparksoniq/semantics/types/AtomicTypes.java
```diff
@@ -0,0 +1,11 @@
+package sparksoniq.semantics.types;
+
+public enum AtomicTypes {
+    AtomicItem,
+
+    StringItem,
+    IntegerItem,
+    DecimalItem,
+    DoubleItem,
+    BooleanItem,
```
ghislainfourny: and NullItem

src/main/java/sparksoniq/semantics/types/AtomicType.java
```diff
@@ -0,0 +1,31 @@
+package sparksoniq.semantics.types;
+
+
+import java.io.Serializable;
+
+public class AtomicType implements Serializable {
```
ghislainfourny: The name is misleading, as integer? for example is a sequence type, but not an atomic type. An atomic type is a specific kind of item type that does not have a cardinality.

-> Something along the lines of AtomicSequenceType, or CastType (or any other idea you may have)?

src/main/resources/test_files/runtime/Castable.jq
```diff
@@ -0,0 +1,22 @@
...
+2 castable as boolean,
+null castable as string?,
+() castable as string?,
+() castable as string,
+"false" castable as boolean,
+"3e4" castable as double,
+"3" castable as decimal,
+3 castable as string,
+3.4 castable as boolean
+
```
ghislainfourny: and for example null castable as null (true), "null" castable as null (true), "foo" castable as null (false), 2 castable as null (false).

The null type is formally defined in the spec as a type with a singleton value space containing only the null value, associated with the literal "null". You will not find it in the W3C website because this is introduced by JSONiq.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/CastableIterator.java
```diff
@@ -0,0 +1,69 @@
...
+        this._atomicType = atomicType;
+    }
+
+    @Override
+    public Item next() {
+        if (this._hasNext) {
+            List<Item> items = new ArrayList<>();
+            _child.open(_currentDynamicContext);
+            while (_child.hasNext())
+                items.add(_child.next());
```
ghislainfourny: We can stop after adding two values. This is enough to know what to return or throw an error and will not crash if there is a long sequence.

src/main/java/sparksoniq/jsoniq/item/StringItem.java
```diff
@@ -76,6 +80,25 @@ public boolean isTypeOf(ItemType type) {
         return type.getType().equals(ItemTypes.StringItem) || super.isTypeOf(type);
     }
 
+    @Override
+    public boolean isCastableAs(AtomicType type) {
+        if (type.getType() == AtomicTypes.StringItem)
+            return true;
+        try {
+            if (type.getType() == AtomicTypes.IntegerItem)
+                Integer.parseInt(this._value);
```
ghislainfourny: Are you positive the JVM optimizer will not throw this line away, because the result is not assigned? If we can't tell whether Java guarantees that, I can still live with that because there are tests in place, but we need to keep in mind that may change at any time.

src/main/java/sparksoniq/jsoniq/item/ObjectItem.java
```diff
@@ -142,7 +142,7 @@ public String serialize() {
             String key = _keys.get(i);
             Item value = _values.get(i);
             boolean isStringValue = value.isString();
-            sb.append("\"" + StringEscapeUtils.escapeJson(key) + "\"" + " : ");
+            sb.append("\"").append(StringEscapeUtils.escapeJson(key)).append("\"").append(" : ");
```
ghislainfourny: Nice fix.

src/main/java/sparksoniq/jsoniq/item/NullItem.java
```diff
@@ -47,6 +49,11 @@ public boolean isTypeOf(ItemType type) {
         return type.getType().equals(ItemTypes.NullItem) || super.isTypeOf(type);
     }
 
+    @Override
+    public boolean isCastableAs(AtomicType type) {
+        return false;
```
ghislainfourny: This is be true for "null" and "null?".

src/main/java/sparksoniq/jsoniq/item/StringItem.java
```diff
@@ -76,6 +80,25 @@ public boolean isTypeOf(ItemType type) {
...
+            else if (type.getType() == AtomicTypes.DecimalItem) {
+                if (this._value.contains("e") || this._value.contains("E")) return false;
+                Float.parseFloat(this._value);
+            } else if (type.getType() == AtomicTypes.DoubleItem)
+                Double.parseDouble(this._value);
+            else return isBooleanLiteral(this._value);
```
ghislainfourny: And if type.getType() == AtomicTypes.NullItem, this works if the string is "null".

src/main/java/sparksoniq/semantics/types/AtomicType.java
ghislainfourny: Thanks! SingleType sounds good.

In Zorba, it is probably something we forgot, that's a good catch. This is really an almost useless corner case and we didn't sweep in all corners. What I want to avoid above all, here, is to open a can of worms if we start treating null differently than other atomic types. Otherwise it will come back to bite us later.

In JSound too, null is a singleton-space atomic type, which we can union with other types to create "integer-or-null" types or the like.

src/main/java/sparksoniq/jsoniq/item/StringItem.java
AndreaRinaldi1: I think the JVM could skip them if those parsing method were not called outside a try-catch. Since we are catching an exception (_NumberFormatException_) that they all might potentially throw, the JVM should not skip those statement. This is also what the IDE seems to suggest me if I were to take those method calls out of the try-catch, by saying that their result is not used.

src/main/resources/test_files/runtime/Castable.jq
AndreaRinaldi1: _Added the new tests._

src/main/java/sparksoniq/jsoniq/compiler/parser/Jsoniq.g4
AndreaRinaldi1: I will reply here to this and subsequent comments on adding _NullItem_ to the list of item types we can cast to. 

From the second review iteration, I added it to the AtomicTypes (as it should have been since the beginning), I removed the SimpleType definition in the grammar, and added the necessary checks on whether items can be cast to NullItem and viceversa. Indeed now, we can do something like _x castable as null_. Only _null castable as null_, _"null" castable as null_ and the same queries with the _ZeroOrOne_ arity on the target item type, returns true. All other castable_as queries with null as target item type will return false. Also, _null castable as string_ will be true.

src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
```diff
@@ -36,6 +36,7 @@
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworClause;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworExpression;
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworVarSequenceType;
+import sparksoniq.jsoniq.compiler.translator.expr.flowr.FlworVarSingleType;
```
wscsprint3r: This class is becoming a monster, at some point we should do something about it

PR #328, opened 2019-10-13
Substring after

AndreaRinaldi1:
This branch introduces the substring-after builtin function (with two string arguments) as described in the [W3C documentation](https://www.w3.org/TR/xpath-functions-30/#func-substring-after).


wscsprint3r:
LGTM

wscsprint3r:
@ghislainfourny what do you think about adding an automatic code formatter plug-in like spotless to have consistent code style across the project?

ghislainfourny:
I agree. Would you have the time to look into this? I don't mind much about the "cosmetics" of spaces and tabulations, so you can pick what you like.

Regarding higher level style, I think we should start writing down our architectural decisions, which would make it easier for people joining the project. This is something I'll start working on.

AndreaRinaldi1:
Pulled master and pushed new version, to have a clean merge.
_Waiting for approval to merge..._

wscsprint3r:
@ghislainfourny I can have a look, yes.

PR #329, opened 2019-10-13
Change how item comparison is performed

AndreaRinaldi1:
In this branch I restructure the way comparison between items is performed. 

**Previously:**
ComparisonOperationIterator calls _comparePair(left, right)_which conditions on the left item type and then on the right item type to check for incompatible comparisons. If OK, _compareItems(left, right)_ is called, which switches over the operator to check for the truthfulness of the comparison. 

This is not extensible as different item types might support some kind of comparison (e.g. `eq` and `ne`) but not others, and it is not really good code quality to check with such cascade of _if...else_ and _switch_ for all the possible combinations of item types and corresponding supported comparisons.

**Now:**
ComparisonOperationIterator calls _comparePair(left, right)_ which checks on items being arrays, objects or atomics. If atomics, _compareItem(right, operator)_ is called on the _left_ Item. This function is overridden by all the Atomic item types. This is where, if necessary, we can perform checks on the comparison operator being supported or not by the item type. From there, the _apply(left, right)_ function is called on the operator which gets the comparison value (one from {-1,0,1}) from the _compareTo(right)_ function called on the _left_ item, and returns the boolean result of the comparison.

This way everything is modular, extensible and follows better OO programming practices.


ghislainfourny:
Thanks a lot. The code looks clean overall and it is nice to organize things a bit as more items will arrive. Much of the inheritance here can be justified with the otherwise high complexity of comparisons.

My main, overall concern is to keep a similar end-to-end performance (which takes precedence over OO practices). Could you please measure the runtime of a query running millions or billions of comparisons on the master vs. your branch to demonstrate that there is no loss in performance?

For example creating a large random dataset (e.g., 10 million) with JSON documents with all kinds of atomic values, and running a filter query based on various comparisons (eq, le, etc) that use the modified code. The great language game dataset may also be a way to start for example with strings.

ghislainfourny:
I am adding @CanBerker as a reviewer, who can help with the performance checks as he has experience with measurements. 

@Can I recommend trying out the standard queries 1, 4 and 5 with this branch and with the master. They all use comparisons in various ways.

From my perspective, the change is good to go if Can is happy with it too.

CanBerker:
Dear @ghislainfourny and @AndreaRinaldi1, 
I have put together a performance comparison tool/script for Rumble that will hopefully benefit us with this and future performance measurements.

It is accessible in the following repo with a 'readme' that is hopefully informative enough to get you up and running smoothly. If you place this repo/project adjacent to your local rumble project, you can simply call the compare_branches.sh script with the $branch_name as the only parameter. It should compare this branch's performance against the master branch and spit out the results in the aggregate_results.txt file. You can read more on available parameters(such as comparing against non-master branches) in the mentioned readme.

https://github.com/CanBerker/rumble-branch-performance-comparison

I will run the performance measurements locally on my computer, however with much shame to Microsoft, after the recent system update I have been getting random performance drops on my computer which may affect the measurements running in the VM. I would kindly ask @AndreaRinaldi1 to run the measurements on his system as well. This will hopefully both get us more reliable measurements, and get  him up to speed with this tool which I hope to be of use to him.

AndreaRinaldi1:
I have been doing some performance comparison between master branch and this, both using @CanBerker tool and a script that I created to compare not only strings, but also numerics (integers, decimals, doubles) and booleans, i.e. everything that is supported at the moment on master. 
For the former, the results are as follows:

> comparison_operator-Q1: 15205
master-Q1: 15902
comparison_operator-Q4: 57369
master-Q4: 57460
comparison_operator-Q5: 31489
master-Q5: 31760

which shows that the new branch performs slightly better than master for all queries.
As for the results from my tool, I have been performing 500M comparisons across the different -compatible- types mentioned above (5 repetitions of 100 different files, each with 1M comparisons). In general, it seems like there is no definitive winner between the two branches. You can see the plotted result of the DIFFERENCE (in nanoseconds) in execution time between master and this branch [here](https://polybox.ethz.ch/index.php/s/jBOs4rW64WmR4mo). To better clarify, a positive difference means that the master is _slower_ than this branch. 

From these results, it looks like the two branches perform similarly (sometimes this branch is better, and sometimes master is), so I would conclude there is no drop in performance.

PR #330, opened 2019-10-16
Add caching to travis build

CanBerker:
I was considering adding caching for maven imports in order to improve the speed of our travis builds. Do you see any potential problems with this addition @wscsprint3r ?


wscsprint3r:
LGTM

PR #334, opened 2019-10-20
HybridRuntimeIterator - openLocal() simplified

CanBerker:
This PR simplifies the openLocal() method of HybridRuntimeIterators by removing the context parameter while providing the same functionality.

It seems to me that the openLocal() method is only used after the open() method is called on the hybrid iterator. This operation always sets the _currentDynamicContext to the given dynamic context on open(DynamicContext context) call. Hence, passing the same parameter again to openLocal seems redundant as this is already accessible in the member field _currentDynamicContext. Furthermore, this caused various redundant assignments in openLocal() method implementations. These are now cleaned up.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
@@ -22,7 +22,6 @@ package sparksoniq.jsoniq.runtime.iterator;
 
 import org.apache.spark.api.java.JavaRDD;
 import org.rumbledb.api.Item;
-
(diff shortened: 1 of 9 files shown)


ghislainfourny:
@CanBerker Is the target branch not being master intentional?

CanBerker:
@ghislainfourny Indeed it was intentional. When I have multiple PRs for branches that chain one after another, I set the target as the previous branch to make the PR diff easier to review.

ghislainfourny:
> @ghislainfourny Indeed it was intentional. When I have multiple PRs for branches that chain one after another, I set the target as the previous branch to make the PR diff easier to review.

Alright, thank you!

PR #335, opened 2019-10-20
User Defined Functions (Static)

CanBerker:
- Expressions for mainModule, prolog, functionDeclaration are added.
- Entry point for RuntimeIteratorGenerator has been updated with mainModule (previously it was the comma expression inside the mainModule). QueryExecutors and tests are updated.
- Functions catalog  and SparksoniqFunctionSignature is refactored. FunctionIdentifier (name, arity) object defined as the key for the catalog map. 
- UserDefinedFunctions<Identifier, FunctionItem> map is added to the Functions Catalog
- New ItemType - FunctionItem added. It consists of functionName, parameter names, function signature(types for all params & return type),  functionBody (stored as expression to support recursion) and non-localVariables to support dynamic/higher-order function definitions.
  - In the current state type checking for the parameters and return type is not performed. However signature is already implemented to make this addition easier. In the UserDefinedFunctionCallIterator, the argument values are evaluated in the putArgumentValuesInDynamicContext function. Type checking can be placed after the evaluation and before placing the values into the dynamic context. 
- RuntimeIteratorVisitor methods are re-ordered inside the file to make the //region definitions consistent again.
  - VisitFunctionDeclaration is added.
  - VisitFunctionCall is updated to accommodate userDefinedFunctions


src/main/java/sparksoniq/exceptions/codes/ErrorCodes.java
```diff
@@ -33,6 +33,10 @@
 
     public static final String InvalidJsoniqVersionErrorCode = "XQST0031";
 
+    public static final String DuplicateFunctionIdentifier = "XQST0034";
```
ghislainfourny: It would be great to also update the documentation.

ghislainfourny:
A few other comments:

1. Could we have a few tests with functions with no types? (which defaults to item*)
2. Could we have tests with functions in which RDDs or DataFrames are used internally (such as nested FLWORs with json-file calls), and functions for which the body expression iterator can return an RDD?

Thanks!

CanBerker:
I am updating the docs for all the error codes with missing information. Could you please double check if they look ok ? (especially for potential discrepancies between JSONiq and XQuery error messages such as XPST0080: types)

CanBerker:
Merging the PR. If the need arises, further fixes can be handled in new PRs.

PR #336, opened 2019-10-22
Duration type

AndreaRinaldi1:
This branch includes the implementation of the **duration** type as defined in the [W3C documentation](https://www.w3.org/TR/xmlschema11-2/#duration).

The comparison operations (_eq_ and _ne_) that can be performed on such type are also implemented, following the new comparison fashion (see _compareItem()_ function in DurationItem).
You can check that there are indeed the supported operations [here](https://www.w3.org/TR/xquery-30/#mapping).

Types dayTimeDuration and yearMonthDuration will follow in another branch.


AndreaRinaldi1:
I forgot to add the Joda-Time license. Will do before merging

src/main/java/sparksoniq/jsoniq/item/DurationItem.java
```diff
@@ -0,0 +1,183 @@
...
+    public DurationItem() {
+        super();
+    }
+
+    public DurationItem(Period value) {
+        super();
+        this._value = value.normalizedStandard(PeriodType.yearMonthDayTime());
+    }
+
+    public Period getValue() {
```
ghislainfourny: It would be nice to expose this function in the Item class, renamed as getDurationValue(), as is done for all others (getIntegerValue(), getStringValue(), getBooleanValue(), etc...). As for the other getXXXValue() functions, Item.getDurationValue should throw an exception and DurationItem.getDurationValue overrides it here.

That way, this is consistent with the rest of the codebase, there is no need to cast to DurationItem (as is done at three places down there) to obtain the period, and users of the public Item API will be able to extract the period from the item if it is a duration, in the same way they extract other types of values.

ghislainfourny:
@AndreaRinaldi1 Does this branch already support, as is, queries like

duration("...") instance of duration
duration("...") treat as duration
duration("...") cast to string
"..." castable as duration

and if so, could tests be added accordingly?

If not, ignore this remark and no need to address it in this branch (one step at a time).

src/main/java/sparksoniq/jsoniq/item/DurationItem.java
AndreaRinaldi1: This sounds like a good solution. Thanks!

PR #337, opened 2019-10-22
Change how arithmetic operations are performed

AndreaRinaldi1:
In this branch I refactor how _AdditiveOperationIterator_ and _MultiplicativeOperationIterator_ deal with the different arithmetic operations on different types. This change is necessary as it is not feasible anymore to decide the resulting type based on the two operands types and, if we still did this (by folding out all the possible combinations of _left_ & _right_ operands), it would be impossible from the result type to deduce the operands type to cast to (as it is currently done). In addition to this, we would have to further branch based on the different arithmetic functions for each combination of possible _left_ & _right_ operator types.

Another reason that drives for this change, is that there will be many types to come (as you can see in [this table](https://www.w3.org/TR/xquery-30/#mapping)), that would make _AdditiveOperationIterator_ and _MultiplicativeOperationIterator_ unreadable if all the logic was there.

In this new version, similarly to the _"decentralization"_ done already for the _ComparisonIterator_, I am calling the _apply()_ function on the operator, which is then responsible for calling its corresponding arithmetic function that is then computed between the two items.

As a side change, I am also stripping away the trailing decimal zeros for decimals and doubles, allowing to have the correct result for operations like `01 || 2.0 || 3e+1 || 04 || true || null` that should yield `12304truenull` but is currently giving `12.030.04truenull` and allows for "cleaner" results for example in `(1 + 3.14 + 10e1) * 0.1` that currently yields `10.414000000000001` instead of `10.414`.


AndreaRinaldi1:
After running experiments on ~600M arithmetic operations for both master and the new branch, I saw there was quite a drop in performance. I decided to try a different implementation that was _switch_ing between the operators and calling the appropriate arithmetic function on the item, directly from AdditiveOperationIterator and MultiplicativeOperationIterator (instead of calling _apply()_ on the operator). I also removed the logic that for example IntegerItem and DoubleItem were delegating some arithmetic operations to DecimalItem if this was the result type. The result are still consistently better than the previous implementation of the new branch, but consistently worse than master. You can see in [this bar graph](https://polybox.ethz.ch/index.php/s/bM930sL3pyArVo4) the execution time for the different operations for master, the branch as you see it in review now [new branch in green] and the newer faster implementation that I just described [Better new master in orange]. 

If that drop does not seem acceptable to @ghislainfourny , I am open to suggestions on what could be other solutions.

ghislainfourny:
Many thanks for doing this, @AndreaRinaldi1 

Indeed there are a few inefficiencies that can be addressed, and removing the overhead with the new overrides in the enum, as you did, definitely helps.

Switches with direct code are known to be faster than OO dispatch, and a common wisdom in systems is to try to keep the code reasonably "flat" (i.e., a few methods with 10-20 lines of code are better than a lot of one-line methods calling each other).

I created a small demo project here that should help you understand what is typically fast and slow in Java.

https://gitlab.inf.ethz.ch/gfourny/performanceatscale

Picking the right Item method to call is definitely best done with a switch in the iterator, and I recommended pushing this change you describe in your last comment to this branch for a final review.

Now, regarding Items: having a small performance loss with the use the new overriden methods (Item.add, Item.subtract, etc) in the Item hierarchy can be justified with the fact that these are binary operators, and the size of the switch can potentially grow quadratically in the number of types. The bit of performance it would get us to inline everything in a big switch would be costly in terms of maintaining 100+ lines of switch, which could in turn affect quality and correctness, the top priority. So I find it a good compromise to have Item.add, Item.subtract, etc and to call them directly from the switch in the iterator; we can live with this slight overhead.

In the future, the plan is to turn Item into an interface rather than an abstract class, and a flat hierarchy of ***Item objects, which should improve performance with the numerous method calls (this can also be seen in the demo code above). This should give us back some of the loss.

Does it make sense?

src/main/java/sparksoniq/jsoniq/item/DoubleItem.java
```diff
@@ -157,4 +162,44 @@ public Item compareItem(Item other, OperationalExpressionBase.Operator operator,
         }
         return operator.apply(this, other);
     }
+
+    @Override
+    public Item add(Item other) {
+        if (other.isDecimal())
```
ghislainfourny: I think it is the other way round between decimal and double: for decimal and double, double wins. This is called type promotion https://www.w3.org/TR/xquery-30/#dt-type-promotion

AndreaRinaldi1: I think you are right. Thanks for spotting this!

AndreaRinaldi1: I notice that by doing this change, we get that `(1 + 3.14 + 10e1) * 0.1` results in `10.414000000000001` (just like in Master) because with doubles there is no such options as in BigDecimal to _stripTrailingZeros()_. There are two options:

1. We compute operations MUL, DIV, MOD in DoubleItem casting to Decimal and then stripping trailing zeros and then back to double like this:
` ItemFactory.getInstance().createDoubleItem(this.castToDecimalValue().multiply(other.castToDecimalValue()).doubleValue());
` for multiply() function in DoubleItem. I expect this to add some time overhead!
2. We accept the trailing zeros (or nines) as it is in master now

What do you think?

ghislainfourny: Thank you for the thorough check.

The W3C specification relies on the IEEE 754-2008 standard for double arithmetics. This is an advanced topic and, for now, simply forwarding the behavior of double operations in Java as is will be fine. If Java says 10.414000000000001, then we can just return this.

This behavior is not a regression, but an improvement over the master in terms of correctness, so that it is fine to leave some fine-tuning for later.

PR #338, opened 2019-10-23
Hex binary type

AndreaRinaldi1:
In this branch I implemented the **hexBinary** type, together with the supported comparison operations, and already supporting cast, instance of, treat as, castable as.


src/main/java/org/rumbledb/api/Item.java
```diff
@@ -242,6 +242,15 @@ public BigDecimal getDecimalValue() {
         throw new RuntimeException("Item is not a big decimal.");
     }
 
+    /**
+     * Returns the byte[] value of the item, if it is a atomic item of type hexBinary.
+     *
+     * @return the hexBinary value as an array of bytes.
+     */
+    public byte[] getHexBinaryValue() {
```
ghislainfourny: You can rename this method to getBinaryValue(), because this method can also be shared with the future base64BinaryItem; in both cases, the value extracted is byte[].

PR #339, opened 2019-10-24
Base64 binary type

AndreaRinaldi1:
In this branch I implemented the **base64Binary** type, together with the supported comparison operations, and already supporting cast, instance of, treat as, castable as.


src/main/resources/test_files/runtime/InstanceOf.jq
```diff
@@ -25,7 +25,9 @@ null instance of null?,
 () instance of string+,
 () instance of string,
 () instance of (),
-hexBinary("aabbCC") instance of hexBinary
+hexBinary("aabbCC") instance of hexBinary,
+base64Binary("0 FB8 0F+9") instance of base64Binary,
```
ghislainfourny: Something is not quite right, as these are invalid base64 encodings and these last two casts should throw errors (Zorba also doesn't accept this query). There can be no space in base64, only a-z, A-Z, 0-9, + / and =. A (here, large) base64 encoding/literal looks like this:

TWFuIGlzIGRpc3Rpbmd1aXNoZWQsIG5vdCBvbmx5IGJ5IGhpcyByZWFzb24sIGJ1dCBieSB0aGlz
IHNpbmd1bGFyIHBhc3Npb24gZnJvbSBvdGhlciBhbmltYWxzLCB3aGljaCBpcyBhIGx1c3Qgb2Yg
dGhlIG1pbmQsIHRoYXQgYnkgYSBwZXJzZXZlcmFuY2Ugb2YgZGVsaWdodCBpbiB0aGUgY29udGlu
dWVkIGFuZCBpbmRlZmF0aWdhYmxlIGdlbmVyYXRpb24gb2Yga25vd2xlZGdlLCBleGNlZWRzIHRo
ZSBzaG9ydCB2ZWhlbWVuY2Ugb2YgYW55IGNhcm5hbCBwbGVhc3VyZS4=

If there is an issue with DataTypeConverters, another option that is quite popular is org.apache.commons.codec.binary.Base64 with an encode and a decode function.

FYI a common practical mistake with base64 encoding libraries is that a string passed to and from a converter can be interpreted in two ways: as a base64 literal, or as an array of bytes (its characters). Many errors found in software amount to this confusion, as it can get easily tricky and the documentation is not always clear.

src/main/java/sparksoniq/jsoniq/item/Base64BinaryItem.java
```diff
@@ -0,0 +1,167 @@
...
+import sparksoniq.semantics.types.ItemType;
+import sparksoniq.semantics.types.ItemTypes;
+
+import javax.xml.bind.DatatypeConverter;
+import java.util.Arrays;
+import java.util.regex.Pattern;
+
+public class Base64BinaryItem extends AtomicItem {
+
+    private static final String b04char = "([AQgw])";
```
ghislainfourny: These seem to be manual computations related to base64. Base64 libraries already take care of such things.

Using an existing library (like org.apache.commons.codec.binary.Base64 ) takes care of everything with just encode and decode calls, including checking that the literal is well-formed, and in an optimized manner. You should have nothing to do yourself apart from calling the encoding and decoding functions of the library.

This manual processing may explain the issues in the tests.

AndreaRinaldi1: I completely agree with you that should be the case ... **but** if I try INVALID examples (defined [here](http://www.datypic.com/sc/xsd/t-xsd_base64Binary.html), and invalid in Zorba as well) like:
- base64Binary("FB8")
- base64Binary("==0F")
- base64Binary("Q===")

those get encoded to FB8, ==0F and Q=== respectively if I use `Base64.encodeBase64(base64BinaryString.getBytes())` or `DatatypeConverter.parseBase64Binary(base64BinaryString);` as well, without prior check on the matching pattern. 

That is why, both for hexBinary and base64Binary, I decided to implement this necessary check using the pattern defined in the [W3C documentation](https://www.w3.org/TR/xmlschema11-2/#base64Binary).

> This manual processing may explain the issues in the tests.

What are the issues in the tests?

AndreaRinaldi1:
Alright, sounds great. 
Thanks for the further inspection and approval!
_Shipping_

PR #340, opened 2019-10-27
Function references & dynamic functions 

CanBerker:
- Implemented inline function declarations
- Implemented named function references
  - function_name#arity
  - currently not supported for built-in functions as they don't fit into functionItem format
- Implemented argumentList as a postfix extension which enables dynamic function calls.
- Partial application for functions will be implemented in a separate PR.
- 'Expression' keyword is removed from module, prolog and functionDeclarations. 

@ghislainfourny Could you take a look at the behavior of the DynamicFunctionCall-Error2.jq test case. In zorba this query returns (3, 5, 8) by invoking the function for each expression in the parenthesized expression. I haven't found the relevant section in XQuery spec for invoking dynamic functions on sequences. How should we implement this behavior ?


ghislainfourny:
> @ghislainfourny Could you take a look at the behavior of the DynamicFunctionCall-Error2.jq test case. In zorba this query returns (3, 5, 8) by invoking the function for each expression in the parenthesized expression. I haven't found the relevant section in XQuery spec for invoking dynamic functions on sequences. How should we implement this behavior ?

This is specified here:

https://www.w3.org/TR/xquery-30/#id-eval-function-call

paragraph 2.b "If this yields a sequence consisting of a single function with the same arity as the arity of the ArgumentList, let F denote that function. Otherwise, a type error is raised [err:XPTY0004]."

CanBerker:
> If getting it to work with Spark (which involves serialization of Function items) needs more work, you can also raise an issue and address it in another PR.
  
I have added the tests and functions invocation on Spark has indeed proven to be a problem. I have reported an issue here which will be addressed in a separate PR:
https://github.com/RumbleDB/rumble/issues/342

Thanks for pointing out the relevant part in XQuery spec, I must have missed it. The current implementation is compliant with the spec so I will take no further actions there.

src/main/resources/test_files/runtime/FunctionUserDefinedDynamic/NamedFunctionRef-Error10.jq
```diff
@@ -0,0 +1,9 @@
+(:JIQS: ShouldCrash; ErrorCode="XPST0017"; ErrorMetadata="LINE:5:COLUMN:10:":)
+declare function f($x) {
+$x.guess eq "English"
+};
+for $o in json-lines("... language test ...")
+where f($o)
+return $o
+
+(: referencing non-existing function in a FLWOR clause :)
```
ghislainfourny: This is actually supposed to work -- any function declared in the prolog can be used everywhere in the query.

src/main/resources/test_files/runtime/FunctionUserDefinedDynamic/NamedFunctionRef-Error2.jq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldCrash; ErrorCode="XPDY0130"; :)
+max#1
+
+(: referencing built-in functions - not supported :)
```
ghislainfourny: Do we have an issue to keep track of this?

src/main/resources/test_files/runtime/FunctionUserDefinedDynamic/NamedFunctionRef-Error9.jq
```diff
@@ -0,0 +1,7 @@
+(:JIQS: ShouldCrash; ErrorCode="XPST0017"; ErrorMetadata="LINE:3:COLUMN:10:":)
+let $f := function($x) { $x.guess eq "English" }
+for $o in json-lines("... language test ...")
+where $f($o)
+return $o
+
+(: referencing non-existing function in a FLWOR clause :)
```
ghislainfourny: This is also supposed to work (but will be more work to get there).

CanBerker:
@ghislainfourny thanks for pointing these out. However, if you look at the metadata, it's actually the json-lines function that's failing rather than the user defined functions. We currently don't have an implementation for json-lines if I'm not missing something. I thought this was intentional on your side.

ghislainfourny:
> @ghislainfourny thanks for pointing these out. However, if you look at the metadata, it's actually the json-lines function that's failing rather than the user defined functions. We currently don't have an implementation for json-lines if I'm not missing something. I thought this was intentional on your side.

@CanBerker  Yes of course, that was a typo on my part :)

I did want to test this with json-file() to make sure that the function call in the where clause will eventually work both with a static and dynamic function call, like so:

let $f := function($x) { $x.guess eq "English" }
for $o in json-file("confusion-2014-03-02-small.json")
where $f($o)
return $o

and

declare function f(x) {
$x.guess eq "English"
};
for $o in json-file("confusion-2014-03-02-small.json")
where f($o)
return $o

src/main/resources/test_files/runtime/FunctionUserDefinedDynamic/NamedFunctionRef-Error2.jq
CanBerker: I will create the issue for this. I would like to propose keeping it as an active test. If disabled, it might be forgotten/unnoticed at the time of working on this issue. In the current state, it will cause an error which will raise the need for updating it once the correct implementation is in place. Does that make sense?

ghislainfourny: Yes -- please just make sure we remember to fix this and implement references to builtin functions later on. Thanks!

CanBerker: Of course. I just want to point out here that the reason why referencing isn't supported on built-in types is directly based on the fact that they don't fit into to functionItem format that clearly defines body, arguments, signature and so on. Once built-in functions are refactored to a structure that works with these concepts, it should simply be a matter of plugging them in.

CanBerker: Issue created: https://github.com/RumbleDB/rumble/issues/346

CanBerker:
I have added the updated tests in disabled format. They both fail on commaExpression serialization.

Issue #347, opened 2019-10-28
Implementation of json-doc()


ghislainfourny:
Users request the json-doc() function, which can (with a local, sequential iterator) read one value from a JSON file that is not in the JSON Lines format, that is, spread over multiple lines.

PR #348, opened 2019-10-28
Json doc

ghislainfourny:
#347


ghislainfourny:
@wscsprint3r this implements json-doc(), which allows locally reading a JSON value even when spread over multiple lines.

One test seems to fail due to some linking error with the JSON parser, but I couldn't get to the bottom of it. It would be nice if you could take a look, maybe with four eyes we find out.

Thanks!

src/main/java/sparksoniq/exceptions/codes/ErrorCodes.java
```diff
@@ -22,6 +22,8 @@
 
 public class ErrorCodes {
 
+    public static final String ErrorRetrievingResourceCode = "FODC0002";
```
CanBerker: Feels nice to see the alphabetical order in action:) 

I just wanted to point out that there is also the docs/Error codes.md file which I recently updated to match the Errorcodes.java. I don't know if it serves any other purpose than being a readme file for that part of the repo. Feel free to add this new exception there too if you feel like it.

wscsprint3r:
@ghislainfourny very 'nice' bug, I can reproduce it locally when I run the actual travis command  " - mvn -Dtest=RuntimeTests test" but not when I just mvn test

ghislainfourny:
Thanks, @wscsprint3r . If we can't fix it within the next few days, then we can review/merge with the test deactivated for now and open an issue to investigate later.

ghislainfourny:
@wscsprint3r I deactivated the test -- if you agree to merge and stick to manual testing for now, I will open an issue to fix this test later.

src/main/java/sparksoniq/exceptions/codes/ErrorCodes.java
ghislainfourny: Thanks, @CanBerker very good point. I extended the documentation list and also documented the function.

src/main/java/sparksoniq/io/json/JiqsItemParser.java
```diff
@@ -68,7 +68,12 @@ public static Item getItemFromObject(JsonIterator object, IteratorMetadata metad
                 {
                     return ItemFactory.getInstance().createDecimalItem(new BigDecimal(number));
                 }
-                return ItemFactory.getInstance().createIntegerItem(Integer.parseInt(number));
+                try {
+                	return ItemFactory.getInstance().createIntegerItem(Integer.parseInt(number));
+                } catch (NumberFormatException e)
+                {
+                	return ItemFactory.getInstance().createDecimalItem(new BigDecimal(number));
+                }
```
CanBerker: This part reminded me of this comment: https://github.com/RumbleDB/rumble/pull/350#discussion_r343552206

I wonder if there could be an alternative to implement this without the exception and if this would be beneficial for performance.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/io/JsonDocFunctionIterator.java
```diff
@@ -0,0 +1,97 @@
...
+                	FileInputStream fis = new FileInputStream(f);
+                	JsonIterator object = JsonIterator.parse(fis, 1024);
+                    return JiqsItemParser.getItemFromObject (object, getMetadata());
+                } catch (IteratorFlowException e) {
+                    throw new IteratorFlowException(e.getJSONiqErrorMessage(), getMetadata());
+                } catch (FileNotFoundException e) {
+                	throw new ErrorRetrievingResourceException("File " + path.getStringValue() + " not found.", getMetadata());
+                }
+            } else {
+                throw new UnexpectedTypeException("json-doc function has non-strong arg " +
```
CanBerker: Typo here with the error message :  "non-strong arg"

src/main/java/sparksoniq/io/json/JiqsItemParser.java
ghislainfourny: That's a very good point. Unfortunately, I haven't found an easy way to test for parsability as an int.

We could find a middle ground by testing for the length of the string before trying to cast: if it is at least 12 characters long, we can directly create a decimal.

This restricts the exception throwing to strings of 10 chars (no sign) or 11 (with sign), which is a rarer case...

What do you think?

src/main/java/sparksoniq/io/json/JiqsItemParser.java
```diff
@@ -64,11 +64,16 @@ public static Item getItemFromObject(JsonIterator object, IteratorMetadata metad
                 {
                     return ItemFactory.getInstance().createDoubleItem(Double.parseDouble(number));
                 }
-                if(number.contains("."))
+                if(number.contains(".") || number.length() >= 12)
```
ghislainfourny: This completely avoids the exceptions when it is clear it is not a 32-bit int.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/io/JsonDocFunctionIterator.java
ghislainfourny: Good catch, this is fixed too.

src/main/java/sparksoniq/io/json/JiqsItemParser.java
CanBerker: I like the idea of shrinking the surface where exceptions can be triggered but I'm not sure that I entirely grasp which exceptions we're trying to cover and optimize against here.

I see that double and decimal types which contain (e/E) and (.) respectively are already covered.
I'm assuming that non-number inputs are not arriving to this stage of execution (if they did, we would have other exceptions with parseDouble for example).
So the case that i can think about at the moment are the values for which the integer type is not large enough to store the value (assuming this causes a NumberFormatException). If so, firstly, why is number 12 chosen here? Secondly, are there any other triggers of exceptions that I'm missing?

ghislainfourny: No, only that -- this is a JSON parser that already detected that this is a number, so not much else can happen at that point.

So we want to use Decimal if 32 bits are exceeded (which happened with a dataset I am using). The NumberFormatException will happen if the 32 bits are exceeded.

We need at least 12 to be sure it does not fit because 11 could be a negative 32-bit integer: "-1000000000". We could also have an even more fine-grained test looking for a sign and the first digit (if we have 10 digits and the first one is above 5, we can also avoid the exception throwing).

Issue #346, opened 2019-10-28
Built-in functions do not fit into the functionItem format


CanBerker:
XQuery spec defines a functionItem format and guidelines for evaluationg static and dynamic functions at https://www.w3.org/TR/xquery-30/#id-eval-function-call.

Currently, the built-in functions perform operations such as calculating argument values, performing type checking and executing the function body in the iterator directly. In order to be able to reference them dynamically with functionality's such as namedFunctionReferences or partial application, built-in functions must be refactored to fit into the structure implemented in functionItem which is designed around XQuery spec. By having a clear function signature and function body definition, these functions can be passed around in the application as well for further modification (eg. on their params in partial application).

Once this refactoring takes place, partial application and namedfunction references should be enabled on built-in functions. Furthermore, please consider refactoring the Functions catalog to a nicer structure as well, since UserDefinedFunctions and BuiltinFunctions will now be compatible to be stored together.

ghislainfourny:
Update: now everything is in place. The next step is to introduce a new function item class that wraps a builtin function (e.g., just with its name) rather than a runtime iterator, so that it can be returned and invoked via named function references (e.g., count#1).

PR #350, opened 2019-10-29
Implement partial application for local evaluation

CanBerker:
- Argument placeholder added to the grammar
- Partial application implemented for function calls
  - built-in functions are not supported at the moment
  - Static (named UDFs)
  -  Dynamic (inline & named reference)


ghislainfourny:
@CanBerker the tests are failing -- is it ready for review?

CanBerker:
Sorry for late response. Indeed it is ready for review. The test was failing because of wrong line number in the metadata.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
```diff
@@ -40,15 +45,23 @@
     private List<RuntimeIterator> _fnArguments;
```
wscsprint3r: @ghislainfourny I think we should stick to regular Java names with full words -> functionArguments. What do you think?

src/main/resources/test_files/runtime/FunctionPartialApplication/PartialFunctionApplication-Error2.jq
```diff
@@ -0,0 +1,2 @@
+(:JIQS: ShouldCrash; ErrorCode="SPRKIQ0002";  ErrorMetadata="LINE:2:COLUMN:0:" :)
+max(?)
```
ghislainfourny: that should be an open issue (it's supposed to work in the long term as builtin function behave like user defined functions in this respect)

src/main/resources/test_files/runtime-spark/FunctionUserDefinedDynamic/NamedFunctionRefRDD3.jq
```diff
@@ -0,0 +1,5 @@
+(:JIQS: ShouldCrash; ErrorCode="SENR0001"; :)
+declare function fn1() {3};
+parallelize(fn1#0)
```
ghislainfourny: This would also be worth an issue (but I don't expect this to be addressed in the next couple of months as this is advanced)

src/main/resources/test_files/runtime-spark/FunctionPartialApplication/PartialFunctionApplicationRDD1.jq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="(8, 13, 18)" :)
+declare function fn ($i, $j) {$i + $j};
+for $i in parallelize((5, 10, 15))
+return fn(3, ?) ($i)
```
ghislainfourny: This is beautiful!

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
ghislainfourny: Yes, I agree.

src/main/resources/test_files/runtime/FunctionPartialApplication/PartialFunctionApplication-Error2.jq
CanBerker: I had already added this to the existing issue of built-in functions not fitting to the functionItem format. https://github.com/RumbleDB/rumble/issues/346 It is listed under Edit1 as I felt these closely relate. Would you like to have this extracted to a separate issue ?

src/main/resources/test_files/runtime-spark/FunctionUserDefinedDynamic/NamedFunctionRefRDD3.jq
CanBerker: This was actually by design on my part at the time. XQuery data model defines functions as 'Functions have no identity, cannot be compared, and have no serialization.' at https://www.w3.org/TR/xpath-datamodel-30/#dt-function-item

Do we prefer to have some sort of serialization for functions? If so, I will create an issue

src/main/resources/test_files/runtime-spark/FunctionPartialApplication/PartialFunctionApplicationRDD1.jq
CanBerker: Very glad to hear that :)

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
CanBerker: Nice catch! I will address this and double check the last few PRs as well.

src/main/resources/test_files/runtime-spark/FunctionUserDefinedDynamic/NamedFunctionRefRDD3.jq
ghislainfourny: Good point. We do not need to serialize functions -- we can simply follow the spec and throw an error code as specified for unserializable items if it is attempted to serialize a function item -- which this tests apparently already checks. You may also want to throw the appropriate error code if they are compared with value/general comparison.

src/main/resources/test_files/runtime/FunctionPartialApplication/PartialFunctionApplication-Error2.jq
ghislainfourny: Thanks for the feedback -- no need for a separate issue, it's fine that way.

src/main/resources/test_files/runtime-spark/FunctionUserDefinedDynamic/NamedFunctionRefRDD3.jq
CanBerker: @ghislainfourny The function items will throw a non-atomic comparison error (JNTY0004) in the new implementation. 

I will merge the PR on your approval of the tests & error codes.

src/main/java/sparksoniq/semantics/visitor/RuntimeIteratorVisitor.java
```diff
@@ -357,27 +362,39 @@ public RuntimeIterator visitFunctionDeclaration(FunctionDeclaration expression,
...
+                throw new UnsupportedFeatureException(
+                        "Partial application on built-in functions are not supported.",
+                        expression.getMetadata()
                 );
-            } else {
-                throw new RuntimeException(ex1.getMessage());
             }
+            Constructor<? extends RuntimeIterator> ctor = functionClass.getConstructor(List.class, IteratorMetadata.class);
+            return ctor.newInstance(arguments, iteratorMetadata);
+        } catch (UnknownFunctionCallException e) {
```
ghislainfourny: Can you make a note/issue for the future to use a test rather than use exceptions to switch between builtin/UDF? It makes a difference in performance. Exceptions should only be used for actual errors.

It's good for now (correctness and design of test cases had priority) but we should keep it in mind.

We could talk about this next week to clean up the mechanism.

PR #353, opened 2019-10-31
Rdd execution support for function Items

CanBerker:
- Function Items are refactored to store RuntimeIterators rather than body expressions
- FunctionItemIterator renamed to FunctionRuntimeIterator for consistency
- Types references by FunctionItem (FunctionItem, FunctionIdentifier, SequenceType, SequenceType.arity and ItemType) added to kryo serialization
- Dynamic Function Call - FunctionItem input verification improved
- forClauseSparkIterator's  was failing to pass Dynamic context properly. Implementation is changed to match other clause iterators.
- RuntimeIterator's hasNext, isOpen and DynamicContext fields are made transient and removed from Kryo serialization.
- FunctionItem refactored to use Java.io serialization instead of Kryo for serializing the runtimeIterator of the function body.


src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -45,12 +45,15 @@
```
ghislainfourny: That's what I missed -- it is actually serialized with kryo, please forget my last question in the email.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -45,12 +45,15 @@
 
 	private static final long serialVersionUID = 1L;
 	protected static final String FLOW_EXCEPTION_MESSAGE = "Invalid next() call; ";
-    private final IteratorMetadata metadata;
+    private IteratorMetadata metadata;
     protected boolean _hasNext;
     protected boolean _isOpen;
     protected List<RuntimeIterator> _children;
     protected DynamicContext _currentDynamicContext;
```
ghislainfourny: This can/should be transient.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -45,12 +45,15 @@
 
 	private static final long serialVersionUID = 1L;
 	protected static final String FLOW_EXCEPTION_MESSAGE = "Invalid next() call; ";
-    private final IteratorMetadata metadata;
+    private IteratorMetadata metadata;
     protected boolean _hasNext;
     protected boolean _isOpen;
```
ghislainfourny: This can/should be transient, set to false upon reading.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -45,12 +45,15 @@
 
 	private static final long serialVersionUID = 1L;
 	protected static final String FLOW_EXCEPTION_MESSAGE = "Invalid next() call; ";
-    private final IteratorMetadata metadata;
+    private IteratorMetadata metadata;
     protected boolean _hasNext;
```
ghislainfourny: this can/should transient, set to the same as what the regular constructor does upon reading.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -45,12 +45,15 @@
...
     protected boolean _isOpen;
     protected List<RuntimeIterator> _children;
```
ghislainfourny: This is the "meat", together with the IteratorMetadata, the only thing that needs to be serialized (i.e., not transient) here.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -134,7 +137,7 @@ public void reset(DynamicContext context) {
     public void write(Kryo kryo, Output output) {
         output.writeBoolean(_hasNext);
```
ghislainfourny: Writing _hasNext, _isOpen and _currentDynamicContext can be completely dropped (s. email).

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -143,7 +146,7 @@ public void write(Kryo kryo, Output output) {
     public void read(Kryo kryo, Input input) {
         this._hasNext = input.readBoolean();
```
ghislainfourny: _hasNext can be set to false as well as _isOpen. _currentDynamicContext can be set to null.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
ghislainfourny: transient

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
ghislainfourny: transient

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
ghislainfourny: transient

src/main/java/sparksoniq/jsoniq/item/FunctionItem.java
```diff
@@ -135,7 +132,8 @@ public void write(Kryo kryo, Output output) {
         kryo.writeObject(output, this.identifier);
         kryo.writeObject(output, this.parameterNames);
         kryo.writeObject(output, this.signature);
-        kryo.writeObject(output, this.bodyExpression);
+        kryo.writeObject(output, this.bodyIterator.getClass());
+        kryo.writeObject(output, this.bodyIterator);
```
ghislainfourny: You could try to serialize bodyIterator to byte[] using java.io.Serializable, and then write this as a byte array to kryo. That way you circumvent the (possibly untested and broken) kryo serialization of iterators. This is what I suggested before.

src/main/java/sparksoniq/jsoniq/item/FunctionItem.java
```diff
@@ -145,7 +143,8 @@ public void read(Kryo kryo, Input input) {
         this.identifier = kryo.readObject(input, FunctionIdentifier.class);
         this.parameterNames = kryo.readObject(input, ArrayList.class);
         this.signature = kryo.readObject(input, ArrayList.class);
-        this.bodyExpression = kryo.readObject(input, Expression.class);
+        Class<? extends RuntimeIterator> iteratorClass = kryo.readObject(input, Class.class);
+        this.bodyIterator = kryo.readObject(input, iteratorClass);
```
ghislainfourny: and here the reverse: read an array of byte[], and deserialize it with java.io.Serializable.

src/main/java/sparksoniq/jsoniq/item/FunctionItem.java
ghislainfourny: If you do, however, want to use kryo for iterators, I think that writeClassAndObject may be what you are looking for here (and possibly the cause of error).

src/main/java/sparksoniq/jsoniq/item/FunctionItem.java
ghislainfourny: or, if you really want to use Kryo, readClassAndObject

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
CanBerker: Done

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
CanBerker: Done

CanBerker:
Java serializer works like a charm. Putting the PR for review now

 In a quick test using writeClassAndObject was causing a classNotFound exception as before for the subclass of RuntimeIterator that was used. I didn't investigate this further for the time being.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
CanBerker: I tried setting breakpoints for the write and read methods of RuntimeIterator but nothing was hit when i tried a few queries manually and our regular tests as well. I also tried to falsify the values in the read method (so that isOpen and hasNext would be set to true upon object instantiation) but code ran normally again.

When I tried to set similar breakpoints for write and read methods of BooleanItem, the breakpoints were actually hit in some of the tests so I believe that the kryo serialization for RuntimeIterators are in fact not used.

Please let me know if you have a better testing method in mind or if there is something missing in my understanding.

ghislainfourny: A place to look at would be the Spark configuration. There are two things that should be configurable: the serialization of results (the Items in the RDDs), which Stefan made use Kryo already during his thesis (plus you did the same for manual serialization in DataFrames), and on the other hand the serialization of all closures passed to Spark in transformations (and this is were the Iterators are serialized), which we are unsure uses Java IO or Kryo.

While I do not have the exact parameters on my mind, this may be a useful reading to start with: 

https://ogirardot.wordpress.com/2015/01/09/changing-sparks-default-java-serialization-to-kryo/

However, since for now the matter with FunctionItems is fixed with using Java IO for the iterator, there is no urgency in this. You can look into this when you see fit.

ghislainfourny:
> Java serializer works like a charm. Putting the PR up for review now
> 
> In a quick test using writeClassAndObject was causing a classNotFound exception as before for the subclass of RuntimeIterator that was used. I didn't investigate this further for the time being.

Yes, this can be left aside for now. You can include this problem in the issue to not lose this information.

PR #357, opened 2019-11-05
Implement parser for Parquet.

ghislainfourny:
Parquet parser; ongoing effort, nested objects are not tested yet (but nested arrays now are). That first try is not very efficient and performance will need to be considerably improved later, by finding ways of bypassing those parts of the DataFrames API that force us to cast or construct objects, and possibly by sharing the schema info across entire partitions.


ghislainfourny:
@CanBerker this is not yet in a state to be merged, but you can take a look. The conversion from any DataFrame (not necessarily Parquet) to an RDD of Items will be directly reusable (and will need to be improved) when the Iterator API is extended to DataFrames execution in addition to local execution and RDD execution.

When an iterator does not support DataFrames, it can use this map transformation to fall back from DataFrames to RDDs. This is similar to falling back to local execution when RDDs are not supported, by collecting the RDD to a local list.

CanBerker:
This looks very nice! I just wanted to point out one minor detail about style: having the starting curly braces on a separate line is inconsistent with the rest of the project and will appear as changes once an auto-formatting is run on the file.

ghislainfourny:
Thanks, @CanBerker. Yes, this is true -- I do not have a strong opinion on curly braces and will simply follow whichever decision you and Stefan decides for linting.

CanBerker:
@ghislainfourny could you update this branch with the current master so that I can use getItemFromRow function for extended hybridRuntimeIterator

docs/Function library.md
```diff
@@ -582,6 +582,16 @@ count(
 
 (Also see examples for json-file for host and port, sets of files and working directory).
 
+### parquet-file (Rumble specific)
+
+Parses one or more parquet files and returns a sequence of objects.
```
CanBerker: Does the function currently handle multiple parquet file inputs? The implementation seemed like it reads a single file unless I'm missing something.

ghislainfourny: @CanBerker it does work with several files, as for json-file. For this it suffices to use wildcards, etc. This is exactly the same syntax used by the Spark input functions.

ghislainfourny: I have extended the doc to make sure this is clear.

PR #358, opened 2019-11-07
Date type

AndreaRinaldi1:
In this branch I introduce the date type according to the [specifications](https://www.w3.org/TR/xmlschema11-2/#date) defined by W3C. 

All the [arithmetic and comparison operations](https://www.w3.org/TR/xquery-30/#mapping), as well as the [allowed casts](https://www.w3.org/TR/xpath-functions-30/#casting-from-primitive-to-primitive) involving this type, are implemented.

Also tests are in place.

There is one extra bugfix for object lookup, since if we were doing _o.date_ where _date_ is a field of object _o_, the parser would fail, thinking that date was a call to the homonymous function.


AndreaRinaldi1:
I completely agree, I should have waited for #361, #364 and #365 but I realized that when I already opened the PR. Sorry for the inconvenience.

PR #359, opened 2019-11-07
RDD version InstanceOf iterator

AndreaRinaldi1:
The branch includes the implementation of Spark action on RDD for instanceOf iterator.


src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/InstanceOfIterator.java
```diff
@@ -20,59 +20,69 @@
...
+                    if (!item.isTypeOf(itemType)) {
+                        return ItemFactory.getInstance().createBooleanItem(false);
+                    }
                 }
+                return ItemFactory.getInstance().createBooleanItem(true);
+            } else {
+                JavaRDD<Item> childRDD = _child.getRDD(_currentDynamicContext);
+                this._hasNext = false;
+                if (isInvalidArity(childRDD.count())) return ItemFactory.getInstance().createBooleanItem(false);
+                JavaRDD<Boolean> result = childRDD.map(new InstanceOfClosure(_sequenceType.getItemType()));
```
ghislainfourny: This looks very good.

There is optimization potential to make it even faster by avoiding the shuffle overhead in the distinct() transformation : you could also use a map transformation that filters out all valid items, and only keeps invalid items, which does not involve any shuffling. The corresponding closure would probably be only a small change in InstanceOfClosure to return invalid items rather than all item types.

Then all that would remain to be done is call the top() action (or equivalent to just pick any item), check for emptiness of the resulting RDD and return false if it is not empty.

You can either address this in this PR, or merge it and raise an issue to remember optimizing this later.

AndreaRinaldi1: After thinking again about the different possibilities and coming from your suggestion, I think the one that could be the fastest is returning true if an item is invalid and filtering on that transformation. We will have that all the items are valid iff the resulting rdd is empty, otherwise there must have been at least one invalid item that the filter kept in the result.

PR #360, opened 2019-11-07
Hybrid treat as

AndreaRinaldi1:
In this branch I change the TreatAs Iterator to be a subtype of HybridRuntimeIterator and work with Spark RDD when the child iterator isRDD.


src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
```diff
@@ -35,43 +69,57 @@ public Item next() {
...
+    }
 
-        if (_nextResult != null && !_nextResult.isTypeOf(itemType)) {
-            throw new TreatException(" " + ItemTypes.getItemTypeName(_nextResult.getClass().getSimpleName()) + " cannot be treated as type " + sequenceTypeName + _sequenceType.getArity().getSymbol(), getMetadata());
+    @Override
+    public JavaRDD<Item> getRDD(DynamicContext dynamicContext) {
+        _currentDynamicContext = dynamicContext;
+        JavaRDD<Item> childRDD = _iterator.getRDD(dynamicContext);
+
+        long count = childRDD.count();
```
ghislainfourny: There is optimization potential here: you could use an action that only takes up to two items, and use the local count of what is returned (0, 1 or 2). Two indicates more than one in the original sequence, and there is no need for the exact count for the purpose of this iterator.

AndreaRinaldi1:
Thank you @ghislainfourny for this suggestion. I think taking the first two element from the rdd istead of computing the count() could be beneficial. I would have really liked to experiment on this, but I was not able to run such queries on the cluster. I am getting a _java.lang.IllegalArgumentException: Unsupported class file major version 57_ exception. I tried to downgrade to older java version as it is suggested online up to Java 9, but this is not solving the issue. 

If you have any suggestion on how to run queries on rumble that make use of spark, any help would be appreciated.

ghislainfourny:
> Thank you @ghislainfourny for this suggestion. I think taking the first two element from the rdd istead of computing the count() could be beneficial. I would have really liked to experiment on this, but I was not able to run such queries on the cluster. I am getting a _java.lang.IllegalArgumentException: Unsupported class file major version 57_ exception. I tried to downgrade to older java version as it is suggested online up to Java 9, but this is not solving the issue.
> 
> If you have any suggestion on how to run queries on rumble that make use of spark, any help would be appreciated.

Of course, I would suggest for example

json-file("path to some JSON Lines file") treat as object*
json-file("path to some JSON Lines file") treat as object+
json-file("path to some JSON Lines file") treat as array* (error)
json-file("path to some JSON Lines file") treat as item*
json-file("path to some JSON Lines file") treat as integer* (error)
json-file("path to the language game JSON Lines file").guess treat as string+
json-file("path to the language game JSON Lines file").choices treat as array+
json-file("path to the language game JSON Lines file").choices[] treat as string*
json-file("path to the language game JSON Lines file").choices[] treat as string (error)
json-file("path to the language game JSON Lines file").choices[] treat as string? (error)
json-file("path to the language game JSON Lines file").target treat as integer* (error)

Does it help?

AndreaRinaldi1:
I previously tried to downgrade the java version from the 13 up to 9 indeed. Now I tried with 8 as suggested and it worked. 

I run experiments for the two cases on a total of ~2.5B queries using **parallelize(<_range in the tens of millions_>) treat as decimal+** and the resulting average execution times to complete ~40M queries are as follows:
- 18254.8059701ms using count() on the RDD
- 10881.9253731ms using take(2).size() on the RDD

It really seems like the second option performs much better for large RDDs. I would expect the advantage of take() action to be somewhat directly proportional to the RDD' size, as take(n) is fixed to _<= n_ regardless of the size of the RDD, while count() heavily depends on the actual size. 

Thank you for the suggestion @ghislainfourny . If you agree, I would proceed with this implementation then.

ghislainfourny:
This sounds great and your interpretation of the difference makes sense. Yes, you can proceed with the faster approach.

ghislainfourny:
@AndreaRinaldi1 Thank you for doing this. As a follow up for some later work, I think there is a way to make treat as iterators even faster if the input is an RDD and the target sequence type has a cardinality of either one or zero-or-one. Do you see it?

AndreaRinaldi1:
There is the opportunity to improve performance if the cardinality is *, as the checks on size are not necessary at all (we don't even need to set dummy values). 
As for the case when the sequenceType arity is one or OneOrZero, I cannot see any improvement. @ghislainfourny

ghislainfourny:
> As for the case when the sequenceType arity is one or OneOrZero, I cannot see any improvement. @ghislainfourny

These cases (which are known statically because the target type is known statically) can be handled by initializing isRDD to false, and returning the item (or the empty sequence) with the local iterator API, as is done in the instance of iterator to return the single boolean.

That way, the item will not be collected twice (once by the treat as iterator, and a second time later on in the iterator chain), but only once with the take action. "Branches" in the RDD dataflow are generally costly -- this is probably unavoidable for the + cardinality, but for the other three cardinalities it can be avoided (for * by doing as you state, for 1 and ? as discussed below).

In short:
- initIsRDD() only needs to be set to true if the child iterator is an RDD and the cardinality of the target type is + or *, otherwise false
- in the local iterator, add a switch in setNextResult that uses take(2) to collect the items, check cardinality and return either 0 or 1 items (similar to the instance of iterator, the count function, etc).

Queries like

let $o := json-file(".../file.json")[1] treat as object
return size($o.choices)

(json-file(".../file.json")[1].guess treat as string?) || " is the target language"

should then have better performance, because the item is only collected once, instead of twice.

Does it make sense?

PR #361, opened 2019-11-07
Time type

AndreaRinaldi1:
In this branch I introduce the time type according to the [specifications](https://www.w3.org/TR/xmlschema11-2/#time) defined by W3C. 

All the [arithmetic and comparison operations](https://www.w3.org/TR/xquery-30/#mapping), as well as the [allowed casts](https://www.w3.org/TR/xpath-functions-30/#casting-from-primitive-to-primitive) involving this type, are implemented.

Also tests are in place.


src/main/java/org/rumbledb/api/Item.java
```diff
@@ -261,6 +261,24 @@ public DateTime getDateTimeValue() {
...
+
+    /**
+     * Returns the dateTime value of the item, if it is a atomic item of type timeItem.
+     *
+     * @return the dateTime value as a DateTime.
+     */
+    public DateTime getTimeValue() {
+        throw new RuntimeException("Item is not a Time.");
+    }
+
```
ghislainfourny: Do we need two methods? Could there be a single method implemented by all items that can return their value as a dateTime?

AndreaRinaldi1: Absolutely! I replaced getDateValue and getTimeValue with getDateTimeValue().

PR #364, opened 2019-11-07
Extract fields from date time

AndreaRinaldi1:
This branch includes the implementation of the builtin functions to do component extraction on durations, as defined in [W3C specs](https://www.w3.org/TR/xpath-functions-30/#component-extraction-dateTime).


src/main/java/org/rumbledb/api/Item.java
```diff
@@ -261,6 +261,24 @@ public DateTime getDateTimeValue() {
         throw new RuntimeException("Item is not a DateTime.");
     }
 
+    /**
```
ghislainfourny: Same comment as in the other PR: can we reuse getDateTimeValue() across all these items?

PR #365, opened 2019-11-07
Adjust time zone

AndreaRinaldi1:
In this branch I include the builtin functions to adjust dateTimes, dates and times according to a specified (or not specified) timezone, as described thoroughly in [W3C specs](https://www.w3.org/TR/xpath-functions-30/#timezone.functions). 

_Spoiler:_ there are quite some rules to determine the result based on the different parameters.


ghislainfourny:
> _Spoiler:_ there are quite some rules to determine the result based on the different parameters.

Yes -- date and time is a very complex matter. This is why many languages neglect them, in spite of their being paramount to a lot of users.

PR #366, opened 2019-11-12
getDataFrame support for Hybrid RI

CanBerker:
- getDataFrame(...) return type fixed.
- getRDDAux(...) method defined on Hybrid RI

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
@@ -46,9 +46,6 @@ public abstract class HybridRuntimeIterator extends RuntimeIterator {
     protected boolean isRDDInitialized = false;
     protected boolean _isRDD;
 
-    protected Dataset<Row> _dataFrame;
-    protected boolean _isDataFrame = false;
-
     protected HybridRuntimeIterator(List<RuntimeIterator> children, IteratorMetadata iteratorMetadata) {
         super(children, iteratorMetadata);
         this.parser = new JiqsItemParser();
@@ -65,7 +62,7 @@ public abstract class HybridRuntimeIterator extends RuntimeIterator {
 
     @Override
     public boolean isDataFrame() {
-        return _isDataFrame;
+        return false;
     }
 
     @Override
@@ -144,18 +141,18 @@ public abstract class HybridRuntimeIterator extends RuntimeIterator {
 
 
     @Override
-    public JavaRDD<Item> getRDD(DynamicContext dynamicContext) {
-        if (_isDataFrame) {
-            Dataset<Row> df = this.getDataFrame(dynamicContext);
+    public JavaRDD<Item> getRDD(DynamicContext context) {
+        if (isDataFrame()) {
+            Dataset<Row> df = this.getDataFrame(context);
             JavaRDD<Row> rowRDD = df.javaRDD();
             return rowRDD.map(new RowToItemMapper(getMetadata()));
         } else {
-            return getRDDAux(dynamicContext);
+            return getRDDAux(context);
         }
     }
 
     @Override
-    public Dataset<Row> getDataFrame(DynamicContext dynamicContext) {
+    public Dataset<Row> getDataFrame(DynamicContext context) {
         throw new SparkRuntimeException("DataFrames are not implemented for the iterator", getMetadata());
     }
 
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
@@ -23,7 +23,6 @@ package sparksoniq.jsoniq.runtime.iterator;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
-import org.apache.spark.sql.types.StructType;
 import org.rumbledb.api.Item;
 import sparksoniq.Main;
 import sparksoniq.exceptions.IteratorFlowException;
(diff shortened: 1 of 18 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
```diff
@@ -101,22 +110,45 @@ public boolean hasNext() {
...
+                        } else {
+                            Main.terminal.output("\nWarning: Results have been truncated to: " + SparkSessionManager.COLLECT_ITEM_LIMIT
+                                    + " items. This value can be configured with the --result-size parameter at startup.\n");
+                        }
                     }
+                } else {
+                    rows = (Row[]) _dataFrame.collect();
+                }
+                for (Row row: rows) {
+                    result.add(JiqsItemParser.getItemFromRow(row, getMetadata()));
```
ghislainfourny: Wondering if this portion of the code is necessary thinking on the medium term.

The DataFrame -> RDD fallback will be implemented in getRDD() soon as part of these efforts (which will compute the RDD from the new overridable getRDDAux() method if no dataframes are available, and otherwise computes the RDD from getDataFrame() with a getItemFromRow map).

-> This will get us the DataFrame -> local fall back for free: the get RDD() call two lines below would take care of the DF conversion already (cascading), with a modular, two-level architecture:

DF    ---- fallback ---->    RDD    ---- fallback ----->     local

What do you think? Anything I am missing?

src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
```diff
@@ -140,8 +142,27 @@ public Item next() {
...
 
+
+    @Override
+    public JavaRDD<Item> getRDD(DynamicContext dynamicContext) {
+        if (_isDataFrame) {
+            Dataset<Row> df = this.getDataFrame(dynamicContext);
+            JavaRDD<Row> rowRDD = df.javaRDD();
+            return rowRDD.map(new RowToItemMapper(getMetadata()));
+        } else {
+            return getRDDAux(dynamicContext);
```
ghislainfourny: Wonderful.

CanBerker:
@ghislainfourny ParquetFileFunctionIterator seems to have extended from SparkFunctionCallIterator (extends from SparkRuntimeIterator). Just to confirm: Should I make parquetFileFunctionIterator a HybridRuntimeIterator? This means I will need to also implemented local methods and initisRDD as well

src/main/java/sparksoniq/spark/iterator/function/ParquetFileFunctionIterator.java
```diff
@@ -54,7 +54,22 @@ public void open(DynamicContext context) {
...
+        RuntimeIterator urlIterator = this._children.get(0);
+        urlIterator.open(context);
+        Dataset<Row> rows = SparkSessionManager.getInstance().getOrCreateSession()
+                .read().parquet(urlIterator.next().getStringValue());
+        urlIterator.close();
+        return rows;
+    }
+
+    @Override
+    public JavaRDD<Item> getRDDAux(DynamicContext context) {
```
ghislainfourny: This method can now be completely dropped. Indeed, is DataFrame always returns true and getRDDAux will never be called.

This is a general rule: an iterator that always return true to isDataFrame() does not need to provide getRDDAux. an iterator that always return true to isRDD() does not need to provide local iterators.

Only iterators whose answer to isDataFrame() may sometimes be yes, sometimes be no, need to provide getRDDAux.

Does it make sense?

ghislainfourny:
> @ghislainfourny ParquetFileFunctionIterator seems to have extended from SparkFunctionCallIterator (extends from SparkRuntimeIterator). Just to confirm: Should I make parquetFileFunctionIterator a HybridRuntimeIterator? This means I will need to also implemented local methods and initisRDD as well.
> 
> On a first look, I think the alternative can be also extending the SparkRuntimeIterator with getRDDAux method to add DataFrame support without bothering with local execution for the spark only iterator functionality
> 
> Edit: Second approach made much more sense and I implemented it in the latest comment. SparkRuntimeIterator also supports getDataFrame and getRDDAux with this implementation.

Yes, exactly, Hybrid is the right approach. As said in the review, no need to implement anything besides isDataFrame and getDataFrame.

The other methods (local, getRDDAux) will throw exceptions because they are not overriden, but this is fine because they will never be called by the HybridRuntimeIterator class anyway.

src/main/java/sparksoniq/jsoniq/runtime/iterator/SparkRuntimeIterator.java
```diff
@@ -113,4 +110,22 @@ public Item next() {
...
     }
+
+    @Override
+    public JavaRDD<Item> getRDD(DynamicContext context) {
+        if (isDataFrame()) {
+            Dataset<Row> df = this.getDataFrame(context);
+            JavaRDD<Row> rowRDD = df.javaRDD();
+            return rowRDD.map(new RowToItemMapper(getMetadata()));
+        } else {
+            return getRDDAux(context);
```
ghislainfourny: As a pointer to illustrate my comment: here, this is a duplicate implementation of what is already done in the hybrid iterator -> the SparkRuntimeIterator is no longer needed, because the hybrid iterator does that job, and even better.

CanBerker:
@ghislainfourny I agree with the points you make, I noticed this yesterday as well but did not want to go into the refactoring effort without prior discussion. Having agreed on the need for deprecating SparkRuntimeIterator and using HybridRuntimeIterator, I'm going to suggest that we merge this PR as is and I will handle the refactoring in a new PR, keeping the same functionality with cleaner code. Does it sound good ?

ghislainfourny:
> @ghislainfourny I agree with the points you make, I noticed this as well yesterday but did not want to go into the refactoring effort without prior discussion. Having agreed on the need for deprecating SparkRuntimeIterator and using HybridRuntimeIterator, I'm going to suggest that we merge this PR as is and I will handle the refactoring in a new PR, keeping the same functionality with cleaner code. Does it sound good ?

Absolutely! Let us do that.

I expect the follow-up PR to be relatively small, mostly deleted lines :)

PR #368, opened 2019-11-13
Release1.3

ghislainfourny:
Preparation of the next release


wscsprint3r:
@ghislainfourny do we push to maven?

ghislainfourny:
@wscsprint3r why not? We can do it the flexible way, releasing the master as it is regularly.

Issue #120, opened 2018-11-21
position() and last()


ghislainfourny:
In filters, position() and last() are not supported yet.

CliffordAnderson:
@ghislainfourny We ran into the lack of those functions today. It would be great to add them to the language. Just a 👍.

ghislainfourny:
Many thanks @CliffordAnderson for your feedback! :)

ghislainfourny:
.. and done :)

CliffordAnderson:
@ghislainfourny Thank you so much! I really appreciate your implementing these features so quickly. Can't wait to try them out.

ghislainfourny:
@CliffordAnderson this will be in the next release, but I would be happy to prepare a jar with the current master for you, if you would like to already try this now.

ghislainfourny:
@CliffordAnderson many thanks for your kind words!

PR #369, opened 2019-11-14
Position and Last (local and RDD)

ghislainfourny:
Support for position() and last() in predicates.

https://github.com/RumbleDB/rumble/issues/120

This also works with RDDs (with zipWithIndex for position() and a count action for last()).

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/context/PositionFunctionIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/context/PositionFunctionIterator.java
@@ -53,17 +53,12 @@ public class PositionFunctionIterator extends LocalFunctionCallIterator {
     public Item next() {
         if (this._hasNext) {
             this._hasNext = false;
-            long position = _currentDynamicContext.getPosition();
-            if(position == -1)
+            Item result = _currentDynamicContext.getPosition();
+            if(result == null)
             {
-	            throw new AbsentPartOfDynamicContextException("Context indefined (position) ", getMetadata());
+	            throw new AbsentPartOfDynamicContextException("Context undefined (last) ", getMetadata());
             }
-            if(position < Integer.MAX_VALUE)
-            {
-            	return ItemFactory.getInstance().createIntegerItem((int)position);
-            } else {
-            	return ItemFactory.getInstance().createDecimalItem(new BigDecimal(position));
-            }
+            return result;
         }
         return null;
     }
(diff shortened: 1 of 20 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateIterator.java
```diff
@@ -43,12 +43,16 @@
     private RuntimeIterator _iterator;
     private RuntimeIterator _filter;
     private Item _nextResult;
+    private long _position;
+    private boolean _mustMaintainPosition;
+    private DynamicContext _filterDynamicContext;
```
ghislainfourny: We need to chain dynamic contexts. The context passed to the filter must be a new context, so as not to modify in place the parent context passed to the left hand side and potentially other places.

src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateIterator.java
```diff
@@ -82,9 +97,31 @@ protected void openLocal() {
...
+            _position = 0;
+            _mustMaintainPosition = true;
+        }
         _iterator.open(_currentDynamicContext);
         setNextResult();
     }
+    
+    private void setLast() {
+        long last = 0;
+        _iterator.open(_currentDynamicContext);
```
ghislainfourny: This means that the iteration on the lhs happens twice -- however, this will only actually happen if last() is called in the filter (thanks to the variable dependency check), causing no performance regression on existing queries that were not using last().

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/context/PositionFunctionIterator.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public Item next() {
+        if (this._hasNext) {
+            this._hasNext = false;
+            Item result = _currentDynamicContext.getPosition();
+            if(result == null)
+            {
+	            throw new AbsentPartOfDynamicContextException("Context undefined (last) ", getMetadata());
```
CanBerker: Context undefined (position) ?

ghislainfourny: Good catch!

ghislainfourny:
> Looks great! There are tiny style differences with indentation and parenthesis at a few places but these can be ignored for now. I am thinking of running a project wide style fix in a separate PR some time soon.

Thanks! FYI @wscsprint3r  is also working on a lint, which may help.

Issue #193, opened 2019-02-27
Support for filter with integers on Spark


ghislainfourny:
json-file(".../doc.json")[10]

is not currently supported. It is tricky because we don't want to cause a loss of performance with zipWithIndex transformations if the filter is a boolean test.

The plan is to statically "catch" in the translator the most obvious cases where the filter expression always returns a number (new recursive function isAlwaysInteger() in the expression tree returning true for number constants, arithmetic operations), and in this case, create a new runtime iterator that specifically tests for position.

Likewise, we need to catch the most obvious cases where the filter expression always returns a boolean (new recursive function isAlwaysBoolean() in the expression tree returning true for boolean literals, comparison and boolean logics), and in this case, create the existing iterator that performs filtering.

The remaining cases need to be redirected to an "inefficient, hybrid" iterator that both tests for position and boolean depending on each evaluation.

PR #370, opened 2019-11-14
Allow numbers in predicates also on Spark.

ghislainfourny:
Fixes

https://github.com/RumbleDB/rumble/issues/193

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateClosureZipped.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateClosureZipped.java
@@ -55,7 +55,12 @@ public class PredicateClosureZipped implements Function<Tuple2<Item, Long>, Bool
         _expression.open(dynamicContext);
         Item result = _expression.next();
         _expression.close();
-        return result.getBooleanValue();
+        if(result.isNumeric())
+        {
+        	return new Boolean(result.equals(dynamicContext.getPosition()));
+        } else {
+        	return result.getBooleanValue();
+        }
     }
 
 };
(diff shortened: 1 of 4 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateIterator.java
```diff
@@ -176,7 +181,22 @@ private void setNextResult() {
...
+        		&& !filter.getVariableDependencies().containsKey("last")
+        		&& (
+        			filter instanceof BooleanRuntimeIterator
+        			|| filter instanceof AndOperationIterator
+        			|| filter instanceof OrOperationIterator
+        			|| filter instanceof NotOperationIterator
+        			|| filter instanceof ComparisonOperationIterator
+        		)
+        ) {
+        	Function<Item, Boolean> transformation = new PredicateClosure(filter, dynamicContext);
```
ghislainfourny: This is the optimized (unzipped part) if positions are not needed. Runs as before.

src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateIterator.java
```diff
@@ -176,7 +181,22 @@ private void setNextResult() {
...
+            JavaRDD<Item> resultRDD = childRDD.filter(transformation);
+            return resultRDD;
+        }
+        else
         {
```
ghislainfourny: In all other cases -> we zip.

src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/PredicateClosureZipped.java
```diff
@@ -55,7 +55,12 @@ public Boolean call(Tuple2<Item, Long> v1) throws Exception {
         _expression.open(dynamicContext);
         Item result = _expression.next();
         _expression.close();
-        return result.getBooleanValue();
+        if(result.isNumeric())
+        {
+        	return new Boolean(result.equals(dynamicContext.getPosition()));
```
ghislainfourny: Now follows the semantics of filters

https://www.w3.org/TR/xquery-30/#id-filter-expression

ghislainfourny:
Thanks, @CanBerker

PR #371, opened 2019-11-14
Remove SparkFuncCallIter, Extend SparkRI from HybridRI

CanBerker:


src/main/java/sparksoniq/spark/iterator/function/ParseJsonFunctionIterator.java
```diff
@@ -22,20 +22,20 @@
```
CanBerker: @ghislainfourny  in this file and in ParseTextFunctionIterator under the getRDDAux() method I have noticed that the rdd gets cached in _rdd field. Could you remind me in which scenerio this caching helps us ?

src/main/java/sparksoniq/jsoniq/runtime/iterator/SparkRuntimeIterator.java
```diff
@@ -21,111 +21,54 @@
...
-    @Override
-    public boolean isDataFrame() {
-        return false;
-    }
-
-    @Override
-    public void reset(DynamicContext context) {
-        super.reset(context);
-        result = null;
+    protected JavaRDD<Item> getRDDAux(DynamicContext context) {
```
ghislainfourny: How about moving the exception-throwing implementations of getRDDAux, openLocal, closeLocal, etc to the HybridRuntimeIterator class, so that all iterators (including this one) automatically throw these exceptions by default, without having to implement them if they do not need them?

src/main/java/sparksoniq/spark/iterator/function/ParseJsonFunctionIterator.java
ghislainfourny: None. :)

This was from the early alpha version, where we did not pass any dynamic context to getRDD(), making the RDD always the same. But this was incorrect, because the RDD is, actually, dependent on the dynamic context.

So, in the end, unless there is a good reason specific to an iterator, all these caches should go away at some point.

src/main/java/sparksoniq/jsoniq/runtime/iterator/SparkRuntimeIterator.java
CanBerker: I'm really glad you've asked this because I have considered this as well. Leaving them abstract gives us static errors which I thought might be helpful during development as new iterators are added since developer will exactly know which methods he/she has to implement. Moving exceptions to HybridRI will serve the same purpose as well with less code, but errors for forgotten method implementations will only pop-up at runtime. I don't really have a strong preference. How should we proceed?

ghislainfourny: Both approaches make sense. It is a compromise between on the one hand, assurance not to forget any implementation, and on the other hand, productivity.

My feeling is that, since we have tests for pretty much any iterator, any error that would no longer be thrown statically would otherwise be caught by the tests, which makes me lean towards productivity. It seems we can get the best of both worlds.

Does it make sense?

src/main/java/sparksoniq/spark/iterator/function/ParseJsonFunctionIterator.java
CanBerker: Great! I was thinking of refactoring the _rdd field from the class as well as I saw no obvious benefits until I ran across these two caching snippets which made me question their usefulness.

src/main/java/sparksoniq/jsoniq/runtime/iterator/SparkRuntimeIterator.java
ghislainfourny: Your reasoning makes sense as well. I like it.

Here is a suggestion along the lines of your latest proposal: keep your implementation the way it is, but rename the SparkRuntimeIterator to RDDRuntimeIterator, and use this iterator as the base class of all iterators that can always return RDDs.

Likewise, you can have an iterator DataFrameRuntimeIterator that always returns true for isDataFrames, and throws an exception for getRDDAux. This iterator could in particular be the base class of ParquetFileblabla, as this one is the typical example of an iterator that can always return a DataFrame.

Would this work?

CanBerker: I had the same idea as well but as I'm not very familiar with the performance analysis of method lookups across subclasses I was hesitant to grow the subclass chain further.  So I didn't follow through with this at that moment thinking it might be a premature organization, so to say, in the code as we have only 3-4 SparkOnly (RDD or DF) methods so far.

If a latency is not of a concern here, I really like this idea.

src/main/java/sparksoniq/spark/iterator/function/ParseJsonFunctionIterator.java
ghislainfourny: We are in Switzerland. You are my guest if you want to clean this up.

src/main/java/sparksoniq/jsoniq/runtime/iterator/SparkRuntimeIterator.java
ghislainfourny: Thanks a lot for having performance in mind, this is a very good thought. I agree. Let us keep the DFRuntimeIterator idea aside for now and proceed with only the RDDRuntimeIterator.

(My response to you for later: you do have a point that too big hierarchies and too many function calls are bad for performance. Overall I am especially concerned about the latency of the local iterator methods, because these are the ones called 1000000s of times. The RDD/DF methods are only called once per sequence, and the latency there is less likely to be the bottleneck. In the very-long term, the design with the fastest potential is having a flat hierarchy with RuntimeIterator as an interface with all iterators directly implementing it, with the hybrid code called with a has-a rather than an is-a mechanism. This is something that we can have a discussion on in the long-term.)

CanBerker: I like the name RDDRuntimeIterator, it brings much better clarity.

Regarding making RDDAux abstract in RDDRI however, would affect productivity negatively I think. In that scenerio, for every DF only iterator, we would have to manually add the identical exception throwing method for getRDDAux. Eg. ParquetFunctionIterator would need to define this method explicitly.

ghislainfourny: Perfect. Very clear thinking. I'll just mark as approve and you can go ahead.

PR #373, opened 2019-11-14
Added spotless plugin with custom XML code formatter 

wscsprint3r:
@ghislainfourny we can pick what format we want, I set it to Google Java Format 1.7:

https://google.github.io/styleguide/javaguide.html

Travis will check for formatting violations in the PRs and fail the build. 
The auto-formatter can be run using 

**mvn spotless:apply**


pom.xml
```diff
@@ -148,6 +148,27 @@
                     </dependency>
                 </dependencies>
             </plugin>
+            <plugin>
```
wscsprint3r: This is the only change, the rest are formatting changes

wscsprint3r:
@ghislainfourny please have a look and let me know if you want me to pick another format or just tweak the current one :)

ghislainfourny:
@CanBerker @AndreaRinaldi1

What is your opinion on the details of code formatting? This is the moment to indicate your taste on the matter and reach a consensus, and then we'll stick to the chosen style.

CanBerker:
Firstly thanks a lot @wscsprint3r  for taking the time to look at this. I think this will be very beneficial for keeping style consistent across multiple people and IDEs.

It's quite the of formatting change for me, especially with the switch to double spaces instead of tabs.
I haven't formed strong preferences yet regarding the style itself and I think I can get used to the google style over time.

I would like to point out a few things with regard to friction in the IDEs and productivity though. I have been using IntelliJ IDEA for quite some time and as far as I know that's @AndreaRinaldi1's choice of development environment as well. I have found the following XML schemes for IntelliJ and eclipse https://github.com/HPI-Information-Systems/Metanome/wiki/Installing-the-google-styleguide-settings-in-intellij-and-eclipse
Alternatively google-java-formatter can be helpful for IntelliJ users but it only improves the autoformatting functionality of the IDE (the XML style also fixes the editing as it's the style for the IDE).

I have experimented with both of them and while the result is quite similar, it's **not identical**. It will be for the best and necessary to run the maven style application command prior to pushing.^

I have also ran into one **strange behavior**. The google style-guide shared below specifies that braces are always used even if the body is empty.
https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used
In GroupByClauseSparkIterator line 224, an if clause without braces exist and this doesn't trigger any issues with the Spotless plugin's check and is not fixed by apply. When I run autoformatting with IntelliJ, this gets fixed and braces are added. The Spotless check is still passing and apply doesn't make any further changes. Can @wscsprint3r check why this deviation from the google specification happens with spotless plugin?

AndreaRinaldi1:
Just like @CanBerker said, the google style could work for me as well, even though that is not what I am used to. 
In particular I would prefer 4 spaces block indentation and indent continuation at +8 from original line, that would shift the column limit to 120 instead of 100. Although, if we do not want to tweak the google style format to be able to use existing plugins for the different IDEs, I think I can get used to it.

Thank you @wscsprint3r for taking the time to deal with this.

wscsprint3r:
@CanBerker @ghislainfourny  @AndreaRinaldi1 thanks for the feedback.

I agree that the 2 space indentation is a bit strange, I propose as an alternative the Eclipse JDT Java Format. 

Have a look :)

CanBerker:
I want to point out two things with this new formatting:
1- I would really prefer to use 4 spaces instead of tabs (JDT format inserts tabs)
2- Switch to eclipse formatter was performed after the google formatting. We still have the <p> tags in all javadocs. I'm ok with this if this is desired, but there might be other remnants of google styling which may not be desirable. I'd recommend we apply Eclipse styling on a clean slate.

wscsprint3r:
Added custom XML, we can tweak it as we go if we find we don't like it. It's in **root/codetemplates-eclipse.xml**

wscsprint3r:
So the current status is:
- 120 long lines
- 4 spaces indent
- Wrapping as follows (I have mixed feelings about it):
Screenshot from 2019-11-15 16-34-49

CanBerker:
I like this latest style way more than the google one. Thanks a lot @wscsprint3r !

I looked into whether we can "force braces" (Intellij terminology) or "Use blocks in if/while/for/do statements" (Eclipse terminology) but couldn't see an option for it. This is not essential but I thought it could be a nice to have feature.

wscsprint3r:
@CanBerker I will merge once you approve :)

PR #374, opened 2019-11-15
Typeswitch expression

AndreaRinaldi1:
In this branch I implement the typeswitch expression as defined in the [W3C specs](https://www.w3.org/TR/xquery-30/#id-typeswitch).

The tests are in place as well.


src/main/resources/test_files/runtime/ControlIterators/TypeswitchExpression1.jq
```diff
@@ -0,0 +1,9 @@
+(:JIQS: ShouldRun; Output="string" :)
```
ghislainfourny: Is there a reason why there are only ItemType tests (i.e., SequenceTypes with an exact cardinality of one)?

AndreaRinaldi1: No there was not real reason for that. I just did not realize. 
I added the arity now.
Thank you

PR #375, opened 2019-11-19
Add RDD support to FLWOR tuple

CanBerker:
RDD and DF support added to FLWOR tuples with individual maps.
DF -> RDD -> Local fallback implemented (IteratorMetadata used for DF -> Fallback)
Both data structures have their own get and put methods. These methods also benefit from iteratormetadata for better error messages.
Collecting an RDD is extracted to a method in SparkContextManager.
Logging messages extracted to a method in the Main class.

### DIFF
--- a/src/main/java/sparksoniq/semantics/DynamicContext.java
+++ b/src/main/java/sparksoniq/semantics/DynamicContext.java
@@ -95,8 +95,7 @@ public class DynamicContext implements Serializable, KryoSerializable {
             return _variableCounts.get(varName);
         }
         if (_variableValues.containsKey(varName)) {
-            Item count = ItemFactory.getInstance().createIntegerItem(_variableValues.get(varName).size());
-            return count;
+            return ItemFactory.getInstance().createIntegerItem(_variableValues.get(varName).size());
         }
         if (_parent != null) {
             return _parent.getVariableCount(varName);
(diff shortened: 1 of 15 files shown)


src/main/java/sparksoniq/semantics/DynamicContext.java
```diff
@@ -35,21 +37,9 @@
...
-
-    public enum VariableDependency {
-        FULL,
-        COUNT,
-        SUM,
-        AVG,
-        MAX,
-        MIN
-    }
-
```
CanBerker: Moved within class with formatting

ghislainfourny:
@CanBerker this looks very good. Could you double-check performance at least on the great language game dataset? Thanks!

CanBerker:
You can see the performance logs here in the most recent builds. 
https://travis-ci.org/CanBerker/rumble-branch-performance-comparison/builds

There doesn't seem to be a performance loss with the q1. q4 and q5 are arguably ok, but there is a lot of fluctuation to say for certain from these logs.

The local execution average over 3 iterations give the following results for which the TupleRDD branch looks fine, although there is a fluctuation with the q1 now. 
master-Q1: 87305.66666666667
tupleRDD-Q1: 77112.0
master-Q4: 227966.0
tupleRDD-Q4: 232248.0
master-Q5: 152052.66666666666
tupleRDD-Q5: 148597.66666666666

With this recent investigation I think that the implementation in this branch doesn't bring a performance overhead.

PR #377, opened 2019-11-20
Provide entry point similar to builtin functions for both dynamic calls and user-defined function calls. 

ghislainfourny:


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
```diff
@@ -37,11 +37,10 @@
 import java.util.List;
 import java.util.Map;
 
-public class UserDefinedFunctionCallIterator extends HybridRuntimeIterator {
+public class FunctionItemCallIterator extends HybridRuntimeIterator {
 
     private static final long serialVersionUID = 1L;
     // parametrized fields
-    private FunctionIdentifier _functionIdentifier;
```
CanBerker: Can you move _functionitem(defined below) to thisline to keep consistency with the comments regarding parametrized fields or calculated fields. If these comments are not helpful, you can also remove them if you wish

ghislainfourny: Done! Good catch. The comments make sense.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -292,6 +293,22 @@ public static RuntimeIterator getBuiltInFunctionIterator(
         }
     }
 
+    public static RuntimeIterator getUserDefinedFunctionIterator(
+            FunctionIdentifier identifier,
+            IteratorMetadata metadata,
+            List<RuntimeIterator> arguments
+    ) {
+        return buildUserDefinedFunctionIterator(getUserDefinedFunction(identifier, metadata), metadata, arguments);
```
CanBerker: I think getUserDefinedFunction(identifier, metadata) can be simplified to not have the metadata parameter. It seems unused

ghislainfourny: Also good catch, done.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicallyResolvedFunctionCallIterator.java
```diff
@@ -0,0 +1,147 @@
...
+
+    }
+
+    @Override
+    public void openLocal() {
+        _functionCallIterator = Functions.getUserDefinedFunctionIterator(
+            _functionIdentifier,
+            getMetadata(),
+            _functionArguments
+        );
```
CanBerker: If I'm not mistaken the assignment to _functionCalliterator is redundant. This lookup is already performed in initIsRdd (which is always called before openLocal() or getRDDAux() ) method and saved to class state.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
```diff
@@ -73,82 +74,10 @@ public DynamicFunctionCallIterator(
     @Override
     public void openLocal() {
         setFunctionItemAndIteratorWithCurrentContext();
```
ghislainfourny: @CanBerker does this apply here too that it is redundant?

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicallyResolvedFunctionCallIterator.java
ghislainfourny: Good catch, thanks.

Note to self: resolving the function at init time may prevent recursive call. We may need to move it back to open and getRDD later on to enable recursion.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
CanBerker: Precisely!

ghislainfourny: Very good. For dynamic calls recursion is not an issue as I do not think it is (as of the current version of the W3C spec) possible for a function item to call itself.

ghislainfourny: I reverted, it seems that it does make a difference. We could investigate later on.

ghislainfourny: I think I found the issue. It is because there was this partial application test, but it is redundant with the other iterator. Now it should work.

ghislainfourny: I put it back for now -- there seems to be an issue with serializing and deserializing function items that forces us to reset the function when opening, which we can investigate later.

CanBerker:
The PR looks great overall. I created a new PR #378 (lined up to merge into this one) to discuss the naming of the new classes. If you wish, we can directly merge this PR too, I'll be waiting on your feedback to proceed.

ghislainfourny:
@CanBerker one more fix; the dynamic function call iterator must rebuild the function item iterator every time, because it is dynamic-context-dependent. I added a test that was broken and now fixed.

https://github.com/RumbleDB/rumble/pull/377/commits/a3791ee12793fe8f845d627d3405aa1791161a0c

PR #378, opened 2019-11-20
Improve naming for User defined function call classes

CanBerker:
I tried to improve the naming of classes for the PR #377 . Please feel free to address anything that doesn't suit you as it's mostly a matter of preference in the road to clarity.

The issues I tried to address mainly were
- The duality of DynamicallyResolvedFunctionCallIterator and DynamicFunctionCallIterator as I assumed the difference would not be immediately clear in my opinion for a new pair of eyes. These are now renamed to StaticUserDefinedFunctionCallIterator and DynamicUserDefinedFunctionCallIterator respectively.
- The ambiguity of functionItemCalliterator and the name of the iterator  field of this class . I believe naming these UserDefinedFunctionCallIterator and functionBodyIterator makes the intention of the class much more clearer. 

There could be a new issue with the new class names (UserDefinedFunctionCallIterator, StaticUserDefinedFunctionCallIterator, DynamicUserDefinedFunctionCallIterator) as to making the developer expect that there is some sort of a class hierarchy. In order to address this, the names could be made even more explicit such as "UserDefinedFunctionCallStaticResolutionIterator" and "UserDefinedFunctionCallDynamicResolutionIterator". This would outline the main objective of these two classes which is handling the function resolution, however I didn't proceed with this as I was not sure if it would be over-explaining/ too long names :)

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
@@ -22,7 +22,6 @@ package sparksoniq.jsoniq.runtime.iterator.functions.base;
 
 import sparksoniq.exceptions.DuplicateFunctionIdentifierException;
 import sparksoniq.exceptions.SparksoniqRuntimeException;
-import sparksoniq.exceptions.UnknownFunctionCallException;
 import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
 import sparksoniq.jsoniq.item.FunctionItem;
 import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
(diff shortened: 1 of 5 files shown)


ghislainfourny:
@CanBerker Many thanks for triggering this discussion.

1. I like the idea of naming the classes Static... vs. Dynamic... This reflects well what this is.

2. Regarding FunctionItem vs. UserDefinedFunction: here there is an important distinction. A user defined function is a function that is not builtin. A function item may be a builtin function or a user-defined function. This is why I suggested "FunctionItem" rather than "UserDefinedFunction" for the shared iterator.

3. Regarding FunctionIterator vs. FunctionCallIterator: here too, there is a distinction. I saw that you created classes ending in FunctionIterator that always return function items, therefore the "Call" to distinguish.

Here is a proposal for the class names:

1. StaticUserDefinedFunctionCallIterator (this can never be builtin, because builtin functions have their own iterators each). This is renamed from DynamicallyResolvedFunctionCallIterator.
2. DynamicFunctionCallIterator (maybe a builtin function obtained through a name-ref)

sharing:
3. FunctionItemCallIterator: because all function items (shared between the above)

What do you think?

CanBerker:
Thanks for the explanation in discussion-(2). I had considered the named-refs but felt the false comfort of the current non-built-in function supporting state in my naming choice.

I absolutely agree that call should be in the named with discussion-(3)

I appreciate your proposals. The PR is updated as such.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -123,8 +139,109 @@
 import java.util.HashMap;
 import java.util.List;
 
-
-import static sparksoniq.jsoniq.runtime.iterator.functions.base.Functions.FunctionNames.*;
+import static sparksoniq.jsoniq.runtime.iterator.functions.base.Functions.FunctionNames.ABS;
```
ghislainfourny: Is this intentional? I saw this change go back and forth. I don't mind either way, but it would be nice to settle it.

ghislainfourny:
> Thanks for the explanation in discussion-(2). I had considered the named-refs but felt the false comfort of the current non-built-in function supporting state in my naming choice.

This is perfectly understandable.

> I absolutely agree that call should be in the named with discussion-(3)
> 
> I appreciate your proposals. The PR is updated as such.


Thanks a lot. Please feel free, after solving conflicts, to merge into the other PR. Also a small comment on a change in the imports that may have been triggered by a tool.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
CanBerker: Nice catch! I configured in IntelliJ some time ago to not use wildcard imports. I just checked a PR of @AndreaRinaldi1 which I assume to be using the default import styling of IntelliJ which enables wildcard imports. 

No wildcard imports prevent unexpected collisions in the namespace, however this can be regarded as a somewhat rare case. Wildcard imports reduce the number of lines in our imports significantly. Which may or may not be regarded as important for readability. 

I have no strong preferences over either one. What do you think @AndreaRinaldi1 ?

AndreaRinaldi1: Yes I think I had also configured IntelliJ to use single imports but probably we surpassed the count that I previously set for '*' imports. I just increased the counter to 999. 
We can use single imports, if we prefer.

PR #379, opened 2019-11-21
Group and order typechecked new types

AndreaRinaldi1:
In this branch there are two main features that are implemented:

* When using **as sequenceType** in for, let and group by clauses, the type of the passed assignment expression is checked against the provided type and an exception is thrown if the item is not compatible with that type, or the arity does not match. Currently the exception that is thrown is TreatException, but it should be UnexpectedTypeException. I was thinking of parametrizing TreatIterator constructor to throw different exceptions, because using a try-catch (and raise a new exception) in every place in _{FOR/LET/GROUPBY}ClauseSparkIterator_ where the exception can get raised, would make the code quite unreadable. Let me know what you think about this.

* Group by and Order by now works with the types that I implemented in the last months (Duration, DayTimeDuration, YearMonthDuration, DateTime, Date, Time, Base64Binary, HexBinary). Both the clauses work when the evaluation is local. 
In the other case, there is currently two issues: 
    * When grouping Duration (and subtypes), because items like _duration("PT0M")_ and _yearMonthDuration("P0Y")_ that both have 0 millis, should be grouped together but they do not. This issue does not exist when grouping for example _duration("PT0M")_ and _duration("P0Y")_.
    * When grouping DateTime (or Date or Time), because if I write the timezone from Kryo, two items like _date("2004-04-12")_ and _date("2004-04-12Z")_ that should be group together, would then not be grouped together. If, on the other hand, I do not write that information from Kryo, the two would be grouped together, but when reading back the object, I lost knowledge about whether the initial items originally had the timezone or not, so the timezone is added to all the items (as you can see in the tests), as the default value for _hasTimeZone_ is true.

_I have to change dateTimeGroupIndex to 6 in GroupClauseCreateColumnsUDF file. Will do in next commit_


ghislainfourny:
> I was thinking of parametrizing TreatIterator constructor to throw different exceptions, because using a try-catch (and raise a new exception) in every place in _{FOR/LET/GROUPBY}ClauseSparkIterator_ where the exception can get raised, would make the code quite unreadable. Let me know what you think about this.

Yes, passing which code is to be thrown in the constructor of the same, reused iterator is a good idea.

Indeed, it is a good idea to keep the changes simple and the design lean.

ghislainfourny:
>   * When grouping Duration (and subtypes), because items like _duration("PT0M")_ and _yearMonthDuration("P0Y")_ that both have 0 millis, should be grouped together but they do not. This issue does not exist when grouping for example _duration("PT0M")_ and _duration("P0Y")_.

This is indeed an pre-existing bug in the master, which is solved here independently of durations, but that should take care of the issue.

https://github.com/RumbleDB/rumble/pull/380

I hope this helps you further.

ghislainfourny:
> When grouping DateTime (or Date or Time), because if I write the timezone from Kryo, two items like _date("2004-04-12")_ and _date("2004-04-12Z")_ that should be group together, would then not be grouped together. If, on the other hand, I do not write that information from Kryo, the two would be grouped together, but when reading back the object, I lost knowledge about whether the initial items originally had the timezone or not, so the timezone is added to all the items (as you can see in the tests), as the default value for _hasTimeZone_ is true.

I spent some time investingating this in the specs. My findings are this:

- As far as kryo is concerned, the boolean must be included, because a dateTime with a timeZone is  distinct (in terms of information content) from a dateTime without a timeZone. Kryo should thus preserve this information when writing and reading back.

**Reference:**
https://www.w3.org/TR/xmlschema11-2/#dateTime
"dateTime uses the date/timeSevenPropertyModel, with no properties except ·timezoneOffset· permitted to be absent. The ·timezoneOffset· property remains ·optional·."

- As far as grouping is concerned (which is addressed in the Long column): in order to compare a dateTime with a time zone with a dateTime without a time zone, the latter must be (for the purpose of the comparison) extended with the implicit timezone (which if I correctly recollect, you chose to be GMT).

This means that, indeed, _date("2004-04-12")_ and _date("2004-04-12Z")_, should be grouped together because the former is automatically (only for the purpose of creating the grouping key) extended with Z. The long number added to the DataFrame column should be the same for both. If such is the case, then they will be grouped together.

This issue may also be related to the bug we just fixed on durations, which could explain why it seems that they are not grouped together. Does the issue still arise now that _date("2004-04-12")_ and _date("2004-04-12Z")_ both appear in the output?

**Reference:**
https://www.w3.org/TR/xpath-functions-30/#func-dateTime-equal
"If either $arg1 or $arg2 has no timezone component, the effective value of the argument is obtained by substituting the implicit timezone from the dynamic evaluation context."

I hope this helps.

AndreaRinaldi1:
Thank you very much @ghislainfourny for the investigation.
Indeed, in the commit that I just pushed earlier, I managed to have the two datetimes (one with timezone GMT and the other without) grouped together, as you can see in the tests. The result will not show the timezone in this case (same behavior as Zorba). 
For order-by as well, now the datetime does not show the timezone if it was not originally there.
All that I just mentioned also applies naturally for dates and times.

What actually unlocked this solution is indeed the changes from #380, as I just now needed to add the boolean to the kryo output, and read it back.

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClauseVar.java
```diff
@@ -45,6 +46,12 @@ public ForClauseVar(
         super(FLWOR_CLAUSES.FOR_VAR, varRef, seq, expression, metadataFromContext);
         this.allowEmpty = emptyFlag;
         this.positionalVariableReference = atVarRef;
+        if (seq != null)
+            this.asSequence = new FlworVarSequenceType(
+                    seq.getSequence().getItemType().getType(),
```
ghislainfourny: I guess this will then use the static variable introduced in the other PR, correct?

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/durations/DurationFunctionIterator.java
```diff
@@ -1,5 +1,6 @@
 package sparksoniq.jsoniq.runtime.iterator.functions.durations;
 
+import org.joda.time.Instant;
```
ghislainfourny: A long-term comment: I read that Joda time was used before Java 8, that it was then largely taken over into Java 8, and that the developers of Joda time recommend migrating to the Java 8 library. This may be worth looking into later on (no urgency).

https://www.baeldung.com/joda-time

src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
```diff
@@ -90,13 +90,6 @@ public int compareWithFlworKey(FlworKey flworKey) {
             ) {
                 throw new SparksoniqRuntimeException("Non atomic key not allowed");
             }
```
ghislainfourny: Is there a reason why this check is being removed -- is this checked somewhere else (e.g., the compareTo() call)?

src/main/java/sparksoniq/semantics/visitor/RuntimeIteratorVisitor.java
```diff
@@ -214,7 +214,14 @@ private RuntimeTupleIterator visitFlowrClause(
...
             for (ForClauseVar var : ((ForClause) clause).getForVariables()) {
-                RuntimeIterator assignmentExpression = this.visit(var.getExpression(), argument);
+                RuntimeIterator assignmentExpression = this.visit(var.getExpression(), argument);;
+                if (var.getAsSequence() != null && var.getAsSequence().getSequence() != null)
+                    assignmentExpression = new TreatIterator(
+                            assignmentExpression,
+                            var.getAsSequence().getSequence(),
+                            false,
+                            createIteratorMetadata(clause)
+                    );
```
ghislainfourny: Very nice reuse of the treat as iterator.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -294,11 +293,10 @@ public void print(StringBuffer buffer, int indent) {
         }
 
         // start with an empty projection.
-        Map<String, DynamicContext.VariableDependency> projection =
-            new TreeMap<String, DynamicContext.VariableDependency>();
 
         // copy over the projection needed by the parent clause.
-        projection.putAll(parentProjection);
+        Map<String, DynamicContext.VariableDependency> projection =
+            new TreeMap<>(parentProjection);
```
ghislainfourny: Thank you for catching this.

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
```diff
@@ -160,6 +161,19 @@ private void setAllLocalResults() {
                                     expression.getIteratorMetadata().getExpressionMetadata()
                             );
                         }
+                        if (resultItem.isDuration())
```
ghislainfourny: Why are we converting duration/dateTime/binary grouping keys to strings?

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
```diff
@@ -152,11 +153,23 @@ private void setAllLocalResults() {
...
-                                "Order by keys must be atomics",
-                                orderByExpression.getIteratorMetadata().getExpressionMetadata()
-                        );
+                    if (resultItem != null) {
+                        if (!resultItem.isAtomic())
+                            throw new NonAtomicKeyException(
+                                    "Order by keys must be atomics",
+                                    orderByExpression.getIteratorMetadata().getExpressionMetadata()
+                            );
+                        if (resultItem.isBinary()) {
```
ghislainfourny: Thank you for paying attention to this level of details.

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClauseVar.java
AndreaRinaldi1: Currently asSequence property in FlworVarDecl is null by default and all the places that need it, they first check for nullity. We could stick to this, and then also this part of the code is fine at its current state, or we set the default for this property to a new FlworVarSequenceType with the _mostGeneralSequenceType_ introduced in the other PR, and change all checks for nullity with checks for this new most-general FlworVarSequenceType

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
AndreaRinaldi1: We need to do this because otherwise in local evaluation, items like time("13:20:00") and time("13:20:00Z") or items like dayTimeDuration("PT0M") and yearMonthDuration("P0Y") will have different key and they would not be grouped together. Using their long value and wrapping it into a StringItem allows us to actually compare the long values, and since they are the same for the two pairs of examples mentioned above, they would be grouped together.

src/main/java/sparksoniq/jsoniq/tuple/FlworKey.java
AndreaRinaldi1: Thank you for noticing. Indeed we cannot use such check anymore as we want to allow the comparison between items with different class names (e.g. DayTimeDuration and Duration). Even though, as you rightfully mentioned, it is compareTo() function that takes care of the current comparison fashion, I still need to add a catch for exception, in case this is thrown by compareTo(). I added such check for dataframes as well. 
Thanks you

src/main/java/sparksoniq/spark/iterator/flowr/GroupByClauseSparkIterator.java
AndreaRinaldi1: I now changed FlworKey's hashCode() function to call hashCode() on its keys instead of serialize() as before. Now there is no need for casting to string in GroupByClauseSparkIterator. 
The other change is for the overriding hashCode() function of Duration.
@ghislainfourny

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
```diff
@@ -242,6 +255,18 @@ private void setAllLocalResults() {
                         } else {
                             // do nothing, type is already set to integer
                         }
+                    } else if (currentColumnType.equals("duration")) {
```
ghislainfourny: I am not sure I understand what is going on here.

If columnType is not a duration, for example, will the unexpected type exception in the last else if no longer be thrown?

Also, the puts are idempotents (except yearMonthDuration replaced with dayTimeDuration, which may be an overlook) and seem to write back the value that was already there.

What issue is being addressed here that requires these extra else ifs?

src/main/java/sparksoniq/spark/udf/OrderClauseCreateColumnsUDF.java
```diff
@@ -120,19 +122,46 @@ public Row call(WrappedArray<byte[]> wrappedParameters) {
...
-                        _results.add(nextItem.castToIntegerValue());
-                    } else if (typeName.equals("double")) {
-                        _results.add(nextItem.castToDoubleValue());
-                    } else if (typeName.equals("decimal")) {
-                        _results.add(nextItem.castToDecimalValue());
-                    } else {
-                        throw new SparksoniqRuntimeException("Unexpected ordering type found while creating columns.");
+                    try {
+                        switch (typeName) {
+                            case "bool":
```
ghislainfourny: is "bool" instead of "boolean" an overlook from the past? (all other types correspond to JSONiq types).

src/main/java/sparksoniq/spark/udf/OrderClauseDetermineTypeUDF.java
```diff
@@ -114,11 +115,33 @@ public OrderClauseDetermineTypeUDF(
                 result.add("double");
             } else if (_nextItem.isDecimal()) {
                 result.add("decimal");
+            } else if (_nextItem.isDayTimeDuration()) {
+                result.add("yearMonthDuration");
```
ghislainfourny: did these two get swapped?

src/main/java/sparksoniq/jsoniq/item/Base64BinaryItem.java
```diff
@@ -183,12 +185,14 @@ private String serializeValue() {
 
     @Override
     public void write(Kryo kryo, Output output) {
-        kryo.writeObject(output, this.getStringValue());
+        output.writeInt(this.getValue().length);
+        output.writeBytes(this.getValue());
```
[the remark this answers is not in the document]
AndreaRinaldi1: I agree on keeping the status quo (null for Item*).
I will also leave some comment on the default value of asSequence and why we need to do that change in Arity for the for clause.

src/main/java/sparksoniq/jsoniq/item/DurationItem.java
```diff
@@ -119,7 +119,9 @@ public boolean equals(Object otherObject) {
 
     @Override
     public int hashCode() {
-        return this.getValue().hashCode();
+        return String.valueOf(
+            this.getValue().toDurationFrom(Instant.now()).getMillis()
+        ).hashCode();
```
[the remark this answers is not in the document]
AndreaRinaldi1: I cannot just call _this.getValue().hashCode()_ because the groupBy will not work effectively across Duration and its subtypes. 
I can avoid the cast to String and call the hashCode() on the long value of millis. This should be faster. Thank you

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/durations/DayTimeDurationFunctionIterator.java
```diff
@@ -72,4 +72,61 @@ public void open(DynamicContext context) {
...
+        @Override
+        public void open(DynamicContext context) {
+            super.open(context);
+            try {
+                _durationStringItem = this.getSingleItemOfTypeFromIterator(
+                    this._children.get(0),
+                    StringItem.class,
+                    new UnknownFunctionCallException("yearMonthDuration", this._children.size(), getMetadata())
+                );
+            } catch (UnknownFunctionCallException e) {
```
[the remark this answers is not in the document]
AndreaRinaldi1: Yes, I will shortly open a new PR where I remove all these "late" checks as they are not necessary after the introduction of builtin typechecking. 
As you rightfully said getSingleItemOfTypeFromIterator will then return the item, without checks and casts.

src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
AndreaRinaldi1: You are right, thank you for pointing this out to me. 
Indeed all those else-ifs are not necessary. Although, for durations, I have to introduce something similar to what is being done for numbers. 
Now the UnexpectedTypeException is thrown if there are sorting keys of different types (see _test_files/runtime-spark/DataFrames/OrderbyClauseError.jq_)

src/main/java/sparksoniq/spark/udf/OrderClauseCreateColumnsUDF.java
AndreaRinaldi1: Yes, it looks like an overlook from the past. Fixed

src/main/java/sparksoniq/spark/udf/OrderClauseDetermineTypeUDF.java
AndreaRinaldi1: Yes, thank you for spotting it!

src/main/java/sparksoniq/jsoniq/item/DurationItem.java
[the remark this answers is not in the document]
ghislainfourny: Thank, this reasoning makes sense. Good optimization on using Long.hashCode().

AndreaRinaldi1:
I run the experiment with 20 repetitions of the query5 on the confusion5m.json file (total of 100M queries) both for master and this branch and the former had an avg execution time of 22974.0ms and the latter 23194.4. With less than 1% difference between the two values, I would then conclude that this is not statistically significant and that there is no loss in performance with these changes.
@ghislainfourny

wscsprint3r:
Very good idea to test the performance before merge, it may be difficult to track performance regressions afterwards....

PR #380, opened 2019-11-21
Make sure that only one grouping key is taken when keys in the same group are different.

ghislainfourny:
The current implementation was incorrectly assuming that all equivalent keys within a group would be exactly the same.

When this is not the case, one of the values should be arbitrarily picked.

Replacing in Spark SQL

array_distinct(...)

with array(first(...))

This problem is already present in the master with, for example, integers and decimals (see added test).

### DIFF
--- a/src/main/java/sparksoniq/spark/DataFrameUtils.java
+++ b/src/main/java/sparksoniq/spark/DataFrameUtils.java
@@ -229,7 +229,7 @@ public class DataFrameUtils {
                 queryColumnString.append(serializerUdfName);
                 queryColumnString.append("(");
                 queryColumnString.append("array(");
-                queryColumnString.append("first(");
+                queryColumnString.append("first(`");
             } else {
                 queryColumnString.append(serializerUdfName);
                 queryColumnString.append("(");
@@ -241,7 +241,7 @@ public class DataFrameUtils {
             if (applyCount) {
                 queryColumnString.append("`)");
             } else if (applyDistinct) {
-                queryColumnString.append(")");
+                queryColumnString.append("`)");
                 queryColumnString.append(")");
                 queryColumnString.append(")");
             } else {
--- a/src/main/java/sparksoniq/spark/DataFrameUtils.java
+++ b/src/main/java/sparksoniq/spark/DataFrameUtils.java
@@ -225,27 +225,26 @@ public class DataFrameUtils {
             }
             if (applyCount) {
                 queryColumnString.append("count(`");
-            } else if (applyDistinct) {
-                queryColumnString.append(serializerUdfName);
-                queryColumnString.append("(");
-                queryColumnString.append("array(");
-                queryColumnString.append("first(`");
             } else {
                 queryColumnString.append(serializerUdfName);
                 queryColumnString.append("(");
-                queryColumnString.append("collect_list(`");
+                if (applyDistinct) {
+                    queryColumnString.append("array(");
+                    queryColumnString.append("first(`");
+                } else {
+                    queryColumnString.append("collect_list(`");
+                }
             }
 
             queryColumnString.append(columnName);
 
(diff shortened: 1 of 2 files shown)


ghislainfourny:
@CanBerker this may even have a positive impact on performance for grouping queries, it is worth looking at the measurements.

src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -225,24 +225,25 @@ public static String getGroupbyProjectSQL(
...
+            } else if (applyDistinct) {
+                queryColumnString.append(serializerUdfName);
+                queryColumnString.append("(");
+                queryColumnString.append("first(");
             } else {
                 queryColumnString.append(serializerUdfName);
                 queryColumnString.append("(");
-                if (applyDistinct) {
-                    queryColumnString.append("array_distinct(");
-                }
```
CanBerker: I don't think this needs to be extracted to a separate if condition as the part with serializerUDF and collect_list are still identical. 

I think it would still work if only the line with array_distinct was replaced with array(first(

src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -225,24 +225,25 @@ public static String getGroupbyProjectSQL(
...
             if (applyCount) {
                 queryColumnString.append("`)");
+            } else if (applyDistinct) {
+                queryColumnString.append(")");
+                queryColumnString.append(")");
             } else {
                 queryColumnString.append("`)");
-                if (applyDistinct) {
-                    queryColumnString.append(")");
-                }
```
CanBerker: Same comment as above

src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -225,24 +225,25 @@ public static String getGroupbyProjectSQL(
...
             if (applyCount) {
                 queryColumnString.append("count(`");
+            } else if (applyDistinct) {
+                queryColumnString.append(serializerUdfName);
+                queryColumnString.append("(");
+                queryColumnString.append("first(");
             } else {
                 queryColumnString.append(serializerUdfName);
                 queryColumnString.append("(");
-                if (applyDistinct) {
```
CanBerker: I think it might be nice to rename this variable to something else since we're not using a distinct operation anymore.

CanBerker:
This is quite irrelevant to this PR but I think it might be nice to piggyback a tiny cleanup for the same file: Could you remove the deprecated javadocs for 'inputSchema' and 'duplicateVariableIndex' on line 168-169.

src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -225,27 +225,26 @@ public static String getGroupbyProjectSQL(
...
-                queryColumnString.append("first(`");
             } else {
                 queryColumnString.append(serializerUdfName);
                 queryColumnString.append("(");
-                queryColumnString.append("collect_list(`");
+                if (applyDistinct) {
+                    queryColumnString.append("array(");
+                    queryColumnString.append("first(`");
+                } else {
+                    queryColumnString.append("collect_list(`");
```
CanBerker: I think this should not be in an else statement

ghislainfourny: Actually this is intentional -- first and collect_list are both aggregation operators. This is also why I think performance may improve significantly, because there is no need to collect anything any more.

We only take the first value and put it into an array.

CanBerker: Thanks for the explanation, sounds great !

ghislainfourny:
Many thanks, @AndreaRinaldi1 and @CanBerker

PR #381, opened 2019-11-21
Typecheck builtin functions

AndreaRinaldi1:
Implementation of a function signature API, extension of the builtin function catalogue with
signatures. General type checking of parameters and return values, including promotion. 

Every parameter to every builtin function, as well as the returned item, will have to have a defined SequenceType (I just changed a few functions in the catalogue, that's why tests are failing) that they have to match against. 

To check the arguments types, each argument iterator is wrapped into a TypePromotionIterator that has a TreatIterator child. The only change to TreatIterator is that it will set a boolean variable according to whether TypePromotionIterator has to check for promotion or not. TreatIterator will behave as previously, and if an exception is thrown, TypePromotionIterator will check if promotion is possible or not and either promote the item to that type and "suppress" the TreatException, or throw the exception.

In order to check for the return type, the BuiltinFunctionCallIterator (counterpart of the UserDefinedFunctionCallIterator for builtin functions) is also wrapped into a TypePromotionIterator and this in turn is wrapped into a CheckReturnTypeIterator that throws the appropriate exception if the return type does not match the SequenceType defined for it. 

As for the code structure for functions, there is a BuiltinFunction class that has identifier, the FunctionSignature (a class that contains signature and return type), and the class of the functionCallIterator.


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/FunctionSignature.java
```diff
@@ -0,0 +1,60 @@
...
+    public FunctionSignature(
+            List<SequenceType> parameters,
+            SequenceType returnType
+    ) {
+        this.parameters = parameters;
+        this.returnType = returnType;
+    }
+
+
+    public List<SequenceType> getParameters() {
```
ghislainfourny: This class looks very good. Maybe getParameterTypes as the name of this method?

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/BuiltinFunction.java
```diff
@@ -0,0 +1,47 @@
+package sparksoniq.jsoniq.runtime.iterator.functions.base;
+
+import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
+
+public class BuiltinFunction {
+
+    private FunctionIdentifier identifier;
```
ghislainfourny: This class also looks good.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -22,253 +22,164 @@
...
 public class Functions {
-    private static HashMap<FunctionIdentifier, Class<? extends RuntimeIterator>> builtInFunctions;
+    private static final HashMap<FunctionIdentifier, BuiltinFunction> builtInFunctions;
     private static HashMap<FunctionIdentifier, FunctionItem> userDefinedFunctions;
 
+    private static final Map<String, ItemType> itemTypes;
+
+    static {
+        itemTypes = new HashMap<>();
+        itemTypes.put("String", new ItemType(ItemTypes.StringItem));
```
ghislainfourny: Would it make sense to stick to lower-case (string, hexBinary...), to have the same syntax as in the language and avoid future confusion?

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -277,33 +188,44 @@ public static boolean checkBuiltInFunctionExists(FunctionIdentifier identifier)
...
+                        metadata
+                )
             );
-            return ctor.newInstance(arguments, metadata);
-        } catch (Exception e) {
-            throw new RuntimeException(e.getMessage());
         }
+        return new CheckReturnTypeIterator(
+                new TypePromotionIterator(
+                        new BuiltinFunctionCallIterator(builtinFunction, arguments, metadata),
```
ghislainfourny: I am not sure I understand why this new wrapping object is created, rather than simply sticking to

ctor.newInstance(arguments, metadata)

Could you explain what your reasoning is, or if you ran into any troubles just using the iterator instance as is?

This creates a runtime overhead with additional object constructions, which is costly. For builtin functions, the iterator class is "ready to use off the shelf" without extra machinery. This makes them faster as, again, they may be called billions of times.

Normally, there is no need to modify anything in the function calls iterators themselves -- this is outside of the scope of your thesis. This is the point of having this method as the point of entry to simply wrap iterators inside the treat as and type promotion operators.

ghislainfourny:
> In order to check for the return type, the BuiltinFunctionCallIterator (counterpart of the UserDefinedFunctionCallIterator for builtin functions)

I think this comment helps me understand your thought pattern on this. The counterpart of the UserDefinedFunctionCallIterator is already present in the current state: these are all the iterators associated with the builtin functions in the map (which you put inside BuiltInFunction).

The reason for the dynamic instantiation of iterators inside the class UserDefinedFunctionCallIterator has to do with recursion, possible in user-defined functions.

Builtin functions are not recursive and thus their iterators can and should directly be used "as is".

Removing the BuiltinFunctionCallIterator class would both simplify the code and make execution faster.

Please let me know if there remains any questions on this or if anything is unclear.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
AndreaRinaldi1: I introduced that iterator for a couple of reasons:

- I thought that merging builtin functions and user-defined functions was a plan for the future, so I tried to make them already as similar as possible in the way they are used as well, i.e. introducing this point-of-entry for builtin functions.

- To have a single place where I could catch the TreatException and throw a corresponding (i.e. UnexpectedTypeException with message _"invalid argument type"_). That is the same reason for CheckReturnTypeIterator to throw UnexpectedTypeException with message _"invalid return type"_. If we don't want that I guess we will have to say something more general like "string cannot be promoted to type double" without saying that this error comes from the argument or from the return type (if we also want to get rid of CheckReturnTypeIterator).

ghislainfourny: Thank you for your answer and these details. This is very useful.

> I introduced that iterator for a couple of reasons:
> 
> * I thought that merging builtin functions and user-defined functions was a plan for the future, so I tried to make them already as similar as possible in the way they are used as well, i.e. introducing this point-of-entry for builtin functions.

The two APIs have already been unified, in the sense that both kinds of iterators are created from methods in the function catalogue that both pass a list of runtime iterator children, and return a runtimeiterator.

This methods are called getBuiltInFunctionIterator and buildUserDefinedFunctionCallIterator.

Doing the type checks in these two methods (in the same way) will take care of absolutely everything -- even dynamic function calls. The change in buildUserDefinedFunctionCallIterator, however, can be done in a later PR once this one is finished, but I anticipate it should only be a few lines of code.

There is no plan to merge the iterators beyond that, because of performance: builtin functions must stay streamlined in the way they are as they will be called billions of times. Any extra wrapping machinery would slow down the system.

Thank you for your kind understanding that function call iterators are out of scope in the context of type checking, and should be left untouched. It is important that the other thesis remains unimpacted by these changes, and that everybody can focus on their topic.

> * To have a single place where I could catch the TreatException and throw a corresponding (i.e. UnexpectedTypeException with message _"invalid argument type"_). That is the same reason for CheckReturnTypeIterator to throw UnexpectedTypeException with message _"invalid return type"_. If we don't want that I guess we will have to say something more general like "string cannot be promoted to type double" without saying that this error comes from the argument or from the return type (if we also want to get rid of CheckReturnTypeIterator).

I understand, thank you for this explanation. The design with catching and throwing again has an issue: in getRDDAux, no exception would be thrown as this is a lazy evaluation: the exception would only pop up at the next action in the pipeline.

So, any information that must appear in the error message could/should also be passed as an argument to the treat as iterator, which can directly throw the appropriate error codes early on without catching them later in the pipeline.

Indeed, CheckReturnTypeIterator would then also be superfluous, which also simplifies the code and improves performance (less method calls, less exception catching). This should also considerably diminish the size of this PR (which is the goal of these entry points, which isolate type checking from the function call machinery).

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
AndreaRinaldi1: Yes, thanks for spotting this.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
AndreaRinaldi1: > There is no plan to merge the iterators beyond that, because of performance: builtin functions must stay streamlined in the way they are as they will be called billions of times. Any extra wrapping machinery would slow down the system.

Ok, got it! I thought the final plan was to have a single iterator that would be the entry point for both, but it actually would lose in performance to add another layer of iterator as you rightfully pointed out. 
I removed BuiltinFunctionCallIterator and CheckReturnTypeIterator so, now, TypePromotionIterator is parametrized with the exception message that should be thrown depending on whether this iterator is wrapping parameters or the function iterator (so checking the return type). 


> The design with catching, and throwing again a different exception, has one issue: in getRDDAux, no exception would be thrown as this is a lazy evaluation: the exception would only pop up at the next action in the pipeline.
So, any information that must appear in the error message could/should also be passed as an argument to the treat as iterator, which can directly throw the appropriate error codes early on without catching them later in the pipeline.

I parametrized TreatAsIterator to either throw TreatException or UnexpectedTypeException (as discussed in #379 as well). Although TypePromotionException will have to go through the try-catch of the TreatException because it will have to check for promotion before forwarding the exception further into the pipeline. Do you think the problem you mentioned because of lazy evaluation would still persist? If so, does that mean that we cannot check for promotion with TypePromotionIterator if we are not in local evaluation? I would appreciate if you could point out to me any error (if you see any) in the getRDDAux() function and of TypePromotionIterator, and its closure. Thank you

AndreaRinaldi1:
> Overall it looks very good and clean -- just a comment on the superfluous extra layer of call iterator around builtin functions iterators (see my comments).
> 
> After this first round of review is done, we should also proceed with some runtime measurements to check that the treat-as/promotion do not bring performance regressions on large numbers of calls.

My hypothesis is that there could be some (hopefully quite negligible) decrease in performance as we are adding one layer of iterator to check return type plus one for every argument, but I will make sure to test this.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -352,6 +286,24 @@ public static FunctionItem getUserDefinedFunction(FunctionIdentifier identifier,
          * function that parallelizes item collections into a Spark RDD
          */
         public static final String PARALLELIZE = "parallelize";
+        public static final BuiltinFunction parallelizeFunction1 = new BuiltinFunction(
```
ghislainfourny: I have a productivity comment, to help with the easy population of the builtin function catalogue.

Here, it may be useful to factor out the declarations inside utility methods, inside convenience functions. Then we could have a very compact list list of declarations, looking like this for the four methods already appearing here:

putBuiltInFunction(PARALLELIZE, ParallelizeFunctionIterator.class, "item*", "item*");
putBuiltInFunction(PARALLELIZE, ParallelizeFunctionIterator.class, "item*", "integer", "item*");
putBuiltInFunction(ENDSWITH, EndsWithFunctionIterator.class, "string?", "string?", "boolean");
putBuiltInFunction(HEXBINARY, HexBinaryFunctionIterator.class, "string?", "hexBinary?");

with overloaded putBuiltInFunction methods (one for each arity between 1 and 4, I don't think we have builtin functions for more than 4 -- and the arity is implicitly passed via the number of arguments). The methods putBuiltInFunction can then take care of everything: creating the BuiltinFunction objects as well as calling builtInFunctions.put().

That way, it would be very easy and quick to create the entire list of signatures: these would be all one liners.

Does it help?

AndreaRinaldi1: Dear Prof. Fourny, 
I refactored the whole builtin functions catalogue with the necessary types. Please let me know if the way I implemented it looks good or you prefer to have the put-into-map inside the same function that creates the builtin function. I kept the two things separate but I have no strong preference for one over the other. 

There are quite some file changes because, as I went over the different iterators, I did some code refactoring when necessary to have a cleaner code. 
I also fixed the substring iterator that was not behaving according to specs

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -225,14 +226,14 @@ public IteratorMetadata getMetadata() {
             }
         }
         iterator.close();
-        if (result != null && !(type.isInstance(result)))
+        if (result != null && !(result.canBePromotedTo(new ItemType(ItemTypes.valueOf(type.getSimpleName())))))
```
ghislainfourny: This method is used at many different places in the Rumble codebase. Promotion, however, only applies to specific places such as function calls, order by and some operators.

https://www.w3.org/TR/xquery-30/#id-type-promotion-and-operator-mapping

If you would like to do this change, it should be checked that all callers of this function correspond to these cases (many probably do). If this is not the case for some callers, then a boolean could be added as an extra parameter to allow returning an item that is only promotable to the desired type, or not.

https://www.w3.org/TR/xquery-30/#id-type-promotion-and-operator-mapping

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/StaticUserDefinedFunctionCallIterator.java
```diff
@@ -128,8 +128,7 @@ public boolean initIsRDD() {
 
     public Map<String, DynamicContext.VariableDependency> getVariableDependencies() {
         Map<String, DynamicContext.VariableDependency> result =
-            new TreeMap<String, DynamicContext.VariableDependency>();
-        result.putAll(super.getVariableDependencies());
+                new TreeMap<>(super.getVariableDependencies());
```
ghislainfourny: Very nice. I like this refactoring, as well as the clean ups at other places.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArrayFlattenFunctionIterator.java
```diff
@@ -74,9 +75,7 @@ public void open(DynamicContext context) {
     public void setNextResult() {
         while (_iterator.hasNext()) {
             Item item = _iterator.next();
-            List<Item> singleItemList = new ArrayList<>();
-            singleItemList.add(item);
-            flatten(singleItemList);
+            flatten(Collections.singletonList(item));
```
ghislainfourny: Also a good catch.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/arrays/ArrayFlattenFunctionIterator.java
```diff
@@ -89,7 +88,7 @@ public void setNextResult() {
         }
     }
 
-    public void flatten(List<Item> items) {
+    private void flatten(List<Item> items) {
```
ghislainfourny: I like this too, better encapsulation.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -135,259 +148,349 @@
...
 public class Functions {
-    private static HashMap<FunctionIdentifier, Class<? extends RuntimeIterator>> builtInFunctions;
+    private static final HashMap<FunctionIdentifier, BuiltinFunction> builtInFunctions;
     private static HashMap<FunctionIdentifier, FunctionItem> userDefinedFunctions;
 
+    private static final Map<String, ItemType> itemTypes;
+
+    static {
+        itemTypes = new HashMap<>();
+        itemTypes.put("item", new ItemType(ItemTypes.Item));
```
ghislainfourny: Caching types and indexing them in this way is a very good idea both in terms of performance and productivity.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -135,259 +148,349 @@
...
-        builtInFunctions.put(new FunctionIdentifier(ATAN, 1), ATanFunctionIterator.class);
-        builtInFunctions.put(new FunctionIdentifier(ATAN2, 2), ATan2FunctionIterator.class);
-
-        builtInFunctions.put(new FunctionIdentifier(SUBSTRING, 2), SubstringFunctionIterator.class);
-        builtInFunctions.put(new FunctionIdentifier(SUBSTRING, 3), SubstringFunctionIterator.class);
-        builtInFunctions.put(new FunctionIdentifier(SUBSTRING_BEFORE, 2), SubstringBeforeFunctionIterator.class);
-        builtInFunctions.put(new FunctionIdentifier(SUBSTRING_AFTER, 2), SubstringAfterFunctionIterator.class);
-        for (int i = 0; i <= 100; i++) {
-            builtInFunctions.put(new FunctionIdentifier(CONCAT, i), ConcatFunctionIterator.class);
+        builtInFunctions.put(position.getIdentifier(), position);
```
ghislainfourny: This looks good. Yes, I am fine with keeping the put outside the creation functions.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -396,19 +499,41 @@ public static boolean checkBuiltInFunctionExists(FunctionIdentifier identifier)
 
     public static RuntimeIterator getBuiltInFunctionIterator(
             FunctionIdentifier identifier,
-            IteratorMetadata metadata,
-            List<RuntimeIterator> arguments
+            List<RuntimeIterator> arguments,
+            IteratorMetadata metadata
```
ghislainfourny: This makes sense, the metadata is often passed last.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -432,19 +557,94 @@ public static void clearUserDefinedFunctions() {
...
     }
 
     public static boolean checkUserDefinedFunctionExists(FunctionIdentifier identifier) {
         return userDefinedFunctions.containsKey(identifier);
     }
 
+    private static BuiltinFunction createBuiltinFunction(
+            String functionName,
+            String returnType,
+            Class<? extends RuntimeIterator> functionIteratorClass
```
ghislainfourny: This looks very nice.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -456,467 +656,1085 @@ public static FunctionItem getUserDefinedFunction(FunctionIdentifier identifier)
...
         }
     }
 
-    public static class FunctionNames {
+    static final class FunctionNames {
         /**
          * function that returns the context position
          */
-        public static final String POSITION = "position";
+        static final BuiltinFunction position = createBuiltinFunction(
```
ghislainfourny: Nice.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/object/ObjectProjectFunctionIterator.java
```diff
@@ -97,7 +97,7 @@ public void setNextResult() {
         }
     }
 
-    public Item getProjection(Item objItem, List<Item> keys) {
+    private Item getProjection(Item objItem, List<Item> keys) {
```
ghislainfourny: Nice encapsulation.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/strings/ConcatFunctionIterator.java
```diff
@@ -59,7 +59,7 @@ public Item next() {
                                 getMetadata()
                         );
                     }
-                    if (stringValue != "") {
+                    if (!stringValue.equals("")) {
```
ghislainfourny: Very good catch. This was definitely a bug.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/strings/SubstringFunctionIterator.java
```diff
@@ -50,27 +51,31 @@ public Item next() {
             if (stringItem == null) {
                 return ItemFactory.getInstance().createStringItem("");
             }
-            IntegerItem indexItem = this.getSingleItemOfTypeFromIterator(this._children.get(1), IntegerItem.class);
+            DoubleItem indexItem = this.getSingleItemOfTypeFromIterator(this._children.get(1), DoubleItem.class);
```
ghislainfourny: Also a good catch.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
```diff
@@ -86,15 +97,13 @@ private void setNextResult() {
...
-            throw new TreatException(
-                    " "
-                        + ItemTypes.getItemTypeName(_nextResult.getClass().getSimpleName())
-                        + " cannot be treated as type "
-                        + sequenceTypeName
-                        + _sequenceType.getArity().getSymbol(),
-                    getMetadata()
-            );
-
+            String message = ItemTypes.getItemTypeName(_nextResult.getClass().getSimpleName())
```
ghislainfourny: Very nice. This iterator looks clean now.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public Item nextLocal() {
+        Item result;
+        try {
+            result = _treatIterator.next();
+        } catch (TreatException e) {
+            this.checkTypePromotion(e.getJSONiqErrorMessage());
+            result = _treatIterator._currentResult;
```
ghislainfourny: I see two small but important issues here:
1. Breach of encapsulation in the iterator API (the state of an iterator should not be modified by the consumer).
2. Use of an exception in "normal use" of the engine (that is, without the query having any issue).

In order to solve them, there is a trick commonly used by other iterator-based XQuery/JSONiq engines (so we do not need to reinvent a new design): swapping the treat as and promotion iterators, i.e., having the treat-as iterator calling the promotion iterator rather than the other way round.

That way,
- the promotion iterator's only job is to check whether each item that doesn't match the target type can be promoted or not (with the new canBePromotedTo method you introduced), and if so, promote it.
- the treat as iterator's only job is to check for types (as it already does) and always returns sequences that match the target type (no breach of contract).

With this clean separation of concerns between promotion and type checking, the iterators can be chained completely from outside, in the function catalogue, with no need from one (promotion) to internally instantiate the other (treat as). See my other comments in the course catalogue.

Then
The TreatAsException will only be thrown if no promotion happened in the child iterator, and does not need to be caught, which solves issue 2.
And the TypePromotionIterator no longer needs to alter the state of the TypePromotionIterator, as the latter can take care of deciding to promote all by itself, which solves issue 1.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -396,19 +499,41 @@ public static boolean checkBuiltInFunctionExists(FunctionIdentifier identifier)
...
-        try {
-            Constructor<? extends RuntimeIterator> ctor = functionClass.getConstructor(
-                List.class,
-                IteratorMetadata.class
+        BuiltinFunction builtinFunction = builtInFunctions.get(identifier);
+
+        for (int i = 0; i < arguments.size(); i++) {
+            arguments.set(
+                i,
+                new TypePromotionIterator(
```
ghislainfourny: After disentangling (see my other comment), this can become:

new TreatAsIterator(
...
   new TypePromotionIterator(
        ...
        arguments.get(i),
    ...
)

where the iterators cleanly call each other without even being aware of each other, and without any encapsulation breach.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -396,19 +499,41 @@ public static boolean checkBuiltInFunctionExists(FunctionIdentifier identifier)
...
+                .getConstructor(
+                    List.class,
+                    IteratorMetadata.class
+                );
+            functionCallIterator = constructor.newInstance(arguments, metadata);
+        } catch (ReflectiveOperationException e) {
+            throw new UnknownFunctionCallException(identifier.getName(), arguments.size(), metadata);
+        }
+
+        return new TypePromotionIterator(
```
ghislainfourny: And likewise, after disentangling (see my other comment), this can become:

new TreatAsIterator(
    ...
    new TypePromotionIterator(
        ...
        functionCallIterator
    ...
)

where the iterators cleanly call each other without even being aware of each other, and without any encapsulation breach.

src/main/resources/test_files/runtime/FunctionBinaries/FunctionBase64BinaryError2.jq
```diff
@@ -1,4 +1,4 @@
-(:JIQS: ShouldCrash; ErrorCode="XPTY0004"; ErrorMetadata="LINE:2:COLUMN:13:" :)
+(:JIQS: ShouldCrash; ErrorCode="XPTY0004"; ErrorMetadata="LINE:2:COLUMN:0:" :)
```
ghislainfourny: The fix of swapping the two iterators to disentangle them may possibly help get the right location hint back, because then each iterator works standalone again.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
ghislainfourny: > Do you think the problem you mentioned because of lazy evaluation would still persist? If so, does that mean that we cannot check for promotion with TypePromotionIterator if we are not in local evaluation? I would appreciate if you could point out to me any error (if you see any) in the getRDDAux() function and of TypePromotionIterator, and its closure. Thank you

I think that my comments should address this question, as this is closely related. The implementation of getRDDAux will then follow very naturally.

ghislainfourny: > Ok, got it! I thought the final plan was to have a single iterator that would be the entry point for both, but it actually would lose in performance to add another layer of iterator as you rightfully pointed out.
> I removed BuiltinFunctionCallIterator and CheckReturnTypeIterator so, now, 

Very nice. Thanks a lot for taking care of this.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
AndreaRinaldi1: If I understand correctly, in order to do what you suggest, I would have to basically copy the same behavior of TreatIterator in TypePromotionIterator and add the check for promotion there. As a matter of fact, before checking for promotion I would have to perform the check on the type (_isTypeOf_) that TreatIterator is doing, and I would also have to check for correct arity (and check on empty sequence) i.e. all the checks that TreatIterator does now. The only difference between the two would be that TypePromotionIterator checks for promotion in case the check _isTypeOf_ is false and promote if possible. 
If that is what we end up doing, though, we do not actually need to encapsulate such type of iterator with TreatIterator again.

ghislainfourny: > If that is what we end up doing, though, we do not actually need to encapsulate such type of iterator with TreatIterator again.

Correct -- I would suggest to give it a try. That's good thinking.

One way to do that could be to either clone the treat as iterator and extend the cloned iterator with promotion if a type check fails. Then, only use this cloned iterator.

Another way is to add a boolean parameter to the treat as iterator and extend it with promotion if that boolean is true., Then, use this treat as iterator setting the boolean to true.

In both cases, only a single iterator would suffice.

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
AndreaRinaldi1: I realized that I do not really need this change as the promotion is happening before, thus the check with isInstance works already. 
_change reverted_

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
AndreaRinaldi1: In order to keep to keep the two iterator separate (as they are also throwing exceptions with a different message), I went for the first option. Even thought there is some code duplication, I think this disentanglement makes the iterator hierarchy cleaner and we get the separation of concerns we aimed for. 
As a side benefit, we also got rid of one layer of additional and unnecessary iterator.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -0,0 +1,173 @@
...
+            return _currentResult;
+        } else
+            throw new IteratorFlowException(RuntimeIterator.FLOW_EXCEPTION_MESSAGE, getMetadata());
+    }
+
+    private void setNextResult() {
+        _nextResult = null;
+        if (_iterator.hasNext()) {
+            _nextResult = _iterator.next();
+            if (_nextResult != null)
```
ghislainfourny: next() never returns null if hasNext() was true, so this condition should always be true.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private void checkEmptySequence(long size) {
+        if (
+            size == 0
+                && (_sequenceType.getArity() == SequenceType.Arity.One
+                    ||
+                    _sequenceType.getArity() == SequenceType.Arity.OneOrMore)
+        ) {
+            throw new UnexpectedTypeException(
+                    "Empty sequence cannot be promoted to type "
```
ghislainfourny: The string in the error message is not quite correct: only atomic values can be promoted ( https://www.w3.org/TR/xquery-30/#id-type-promotion-and-operator-mapping ). If an empty sequence is found and the arity is at least 1, then this is a type error.

How about : "Expecting at least one item, but the value provided is the empty sequence." or along these lines?

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -0,0 +1,173 @@
...
+                    getMetadata()
+            );
+        }
+    }
+
+    private void checkItemsSize(long size) {
+        if (size > 0 && _sequenceType.isEmptySequence()) {
+            throw new UnexpectedTypeException(
+                    ItemTypes.getItemTypeName(_nextResult.getClass().getSimpleName())
+                        + " cannot be promoted to type empty-sequence()",
```
ghislainfourny: Same here: "Expecting empty sequence, but the value provided has at least one item."

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -0,0 +1,173 @@
...
+
+
+        if (
+            size > 1
+                && (_sequenceType.getArity() == SequenceType.Arity.One
+                    ||
+                    _sequenceType.getArity() == SequenceType.Arity.OneOrZero)
+        ) {
+            throw new UnexpectedTypeException(
+                    "Sequences of more than one item cannot be promoted to type "
```
ghislainfourny: Expecting at most one item, but the value provided had at least two items.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -0,0 +1,173 @@
...
+        Function<Item, Item> transformation = new TypePromotionClosure(_exceptionMessage, _sequenceType, getMetadata());
+        return childRDD.map(transformation);
+    }
+
+    private void checkTypePromotion() {
+        if (!_nextResult.canBePromotedTo(_sequenceType.getItemType()))
+            throw new UnexpectedTypeException(
+                    _exceptionMessage
+                        + ItemTypes.getItemTypeName(_nextResult.getClass().getSimpleName())
+                        + " cannot be promoted to type "
```
ghislainfourny: This message is correct -- this is an atomic value that cannot be promoted to a type.

src/main/java/sparksoniq/semantics/types/SequenceType.java
```diff
@@ -30,17 +30,20 @@
     private ItemType _itemType;
     private boolean _isEmptySequence = false;
 
+    public final static SequenceType mostGeneralSequenceType = new SequenceType(
```
ghislainfourny: Good optimization.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
```diff
@@ -15,40 +16,48 @@
...
-    private final SequenceType _sequenceType;
     private int _childIndex;
-    private final ItemType itemType;
-    private final String sequenceTypeName;
 
-    public TreatIterator(RuntimeIterator iterator, SequenceType sequenceType, IteratorMetadata iteratorMetadata) {
+    public TreatIterator(
+            RuntimeIterator iterator,
+            SequenceType sequenceType,
+            boolean shouldThrowTreatException,
```
ghislainfourny: Is this boolean still needed, now that function type checking is handled with an iterator of its own?

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -399,19 +512,46 @@ public static boolean checkBuiltInFunctionExists(FunctionIdentifier identifier)
...
+
+        for (int i = 0; i < arguments.size(); i++) {
+            if (builtinFunction.getSignature().getParameterTypes().get(i) != mostGeneralSequenceType) {
+                arguments.set(
+                    i,
+                    new TypePromotionIterator(
+                            "Invalid argument for function " + identifier.getName() + ". ",
+                            arguments.get(i),
+                            builtinFunction.getSignature().getParameterTypes().get(i),
+                            metadata
```
ghislainfourny: Guessing: would passing the metadata of iterator arguments.get(i), rather than that of the parent, solve the location hints, in the tests?

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
AndreaRinaldi1: I am fine with that as well. I just want to point out that Zorba is throwing 

> type error [err:XPTY0004]: empty-sequence() can not be promoted to parameter type xs:integer of function round-half-to-even()

when evaluating `round-half-to-even(1.125, ())` and

> type error [err:XPTY0004]: sequence of more than one item can not be promoted to parameter type xs:integer of function round-half-to-even()

when evaluating `round-half-to-even(1.125, (1,2,3))`.

If you feel that this is not entirely correct, I will proceed with the message you suggested.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
AndreaRinaldi1: The boolean is not needed for this branch, but I will push the TreatIterator with the boolean in #379, so I guess I will just leave it here for faster merge/conflict resolution.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
AndreaRinaldi1: Using the arguments.get(i).getMetadata() for arguments TypePromotionIterator and functionCallIterator.getMetadata() for return type TypePromotionIterator indeed solves this, bringing back the right column number. Thank you for your help.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
AndreaRinaldi1: That is right, thank you for spotting this

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
ghislainfourny: Thank you for the thorough check with Zorba. It is always a good idea to double-check.

I looked at the specification again and think the error message in Zorba was an abuse of language. I cannot find any mention in the spec of the promotion of sequences.

It would be best to not follow Zorba here, and output error messages that do not use the term "promotion", excepted when referring to a single atomic item and atomic type (which is the case for one message).

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
ghislainfourny: Yes, this works. Thank you for the precision.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
ghislainfourny: Very happy to hear it solved the issue! Thank you for fixing this.

AndreaRinaldi1:
I have been experimenting on the performance of this branch compared to master. I run 100 reps of the following query:
`for $i in json-file("confusion.json")`
`return ends-with(substring(substring-after($i.target, ()), string-length($i.guess) - string-length($i.country), size($i.choices)), $i.country)`
where the json file has 500k objects, making it a total of 50M queries. 
This branch has an average execution time of 26953.69ms while for master the value is 23267.27 making it a 13.6% increase in execution time. I would just like to highlight that this should be the "cap" in performance decrease (i.e. worst case scenario) as it just includes functions that have an itemType different from Item* both in the arguments and in the return type, meaning that TypePromotionIterator is needed for every one of those. In addition to this, I decided to use substring function, because type promotion is actually needed there (because I pass an integer where a double is expected), while master is expecting an integer (and getting an integer). This scenario is adding some time execution to the total.

In all the functions where the argument, or the return type, is Item*, TypePromotionIterator is not needed "around" the argument, or the functionCallIterator respectively, and the performance will be the same as for master.

src/main/resources/test_files/runtime/FunctionString/SubstringError1.jq
```diff
@@ -1,4 +1,4 @@
-(:JIQS: ShouldCrash; ErrorCode="XPTY0004"; ErrorMetadata="LINE:2:COLUMN:0:" :)
+(:JIQS: ShouldCrash; ErrorCode="XPTY0004"; ErrorMetadata="LINE:2:COLUMN:22:" :)
```
ghislainfourny: Very nice that some tests now include improved error info.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/strings/StartsWithFunctionIterator.java
```diff
@@ -51,20 +51,20 @@ public Item next() {
                 StringItem.class
             );
             if (substringItem == null || substringItem.getStringValue().isEmpty()) {
-                return new BooleanItem(true);
+                return ItemFactory.getInstance().createBooleanItem(true);
```
ghislainfourny: Very good catch. Boolean items are pooled, so this will improve performance.

src/main/java/sparksoniq/jsoniq/item/FunctionItem.java
```diff
@@ -59,7 +60,7 @@ protected FunctionItem() {
     public FunctionItem(
             FunctionIdentifier identifier,
             List<String> parameterNames,
-            List<SequenceType> signature,
+            FunctionSignature signature,
```
ghislainfourny: @CanBerker do you approve this minor refactoring (a proper class for function signatures, which will also be convenient at many other places).

CanBerker: This is perfectly fine by me. I was actually anticipating a class as a signature but did not want to implement that myself as it would most likely be subject to a similar refactoring when Andrea would work on it.

ghislainfourny: Thanks @CanBerker 

@AndreaRinaldi1 then it's good to go.

AndreaRinaldi1:
> This hypothesis (that with item* types only there is no drop in performance) would be nice to test and validate, to have an experimental confirmation (this is also related to our discussion on the dynamic call on a function with no type spec, which should not instantiate any TypePromotionIterators either). Such experimental confirmations reinforce the quality of the thesis and final report

I absolutely agree. Having experimental backup for your hypothesis is crucial. 
I experimented, running a total of 50M of this query:

`for $i in (json-file("confusion.json"))`
`for $j in (subsequence(reverse(flatten(insert-before(flatten(insert-before(flatten(insert-before(members($i.choices), 0, remove(members($i.choices), 0))), 0, keys($i))), 0, values($i)))), 0, 100))`
`return $j`

where I tried to include as many functions as possible that have _item*_ both as arguments and return types. Indeed all of those function (except for json-file) satisfy this condition. 

I run this query on the confusion.json file from the language game.

The hypothesis is indeed verified, as master (the old branch without these changes) has an avg. execution time of 20601.07ms and this branch (also without the late checks on item type, which I will shortly go to review with) scored an avg execution time of 20451.14ms.

@ghislainfourny What do you think?

PR #382, opened 2019-11-22
Support initial let clause RDDs and DFs without instant materialization

CanBerker:
Current base is tupleRDD branch

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
@@ -20,12 +20,13 @@
 
 package sparksoniq.jsoniq.runtime.iterator.functions.sequences.aggregate;
 
+import org.rumbledb.api.Item;
 import sparksoniq.exceptions.IteratorFlowException;
 import sparksoniq.exceptions.SparksoniqRuntimeException;
 import sparksoniq.jsoniq.item.ItemFactory;
 import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
-import sparksoniq.jsoniq.runtime.iterator.primary.VariableReferenceIterator;
 import sparksoniq.jsoniq.runtime.iterator.functions.base.LocalFunctionCallIterator;
+import sparksoniq.jsoniq.runtime.iterator.primary.VariableReferenceIterator;
 import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
 import sparksoniq.semantics.DynamicContext;
 
@@ -33,11 +34,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
-import org.rumbledb.api.Item;
-
 public class CountFunctionIterator extends LocalFunctionCallIterator {
     /**
-     * 
+     *
      */
     private static final long serialVersionUID = 1L;
 
@@ -50,8 +49,8 @@ public class CountFunctionIterator extends LocalFunctionCallIterator {
         if (this._hasNext) {
             RuntimeIterator iterator = this._children.get(0);
             if (!iterator.isRDD()) {
-                if (_children.get(0) instanceof VariableReferenceIterator) {
-                    VariableReferenceIterator expr = (VariableReferenceIterator) _children.get(0);
+                if (iterator instanceof VariableReferenceIterator) {
+                    VariableReferenceIterator expr = (VariableReferenceIterator) iterator;
                     this._hasNext = false;
                     return _currentDynamicContext.getVariableCount(expr.getVariableName());
                 }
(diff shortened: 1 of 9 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
```diff
@@ -50,36 +49,36 @@ public Item next() {
...
             } else {
-                Long count = iterator.getRDD(_currentDynamicContext).count();
+                long count = iterator.getRDD(_currentDynamicContext).count();
                 this._hasNext = false;
                 if (count > (long) Integer.MAX_VALUE) {
                     // TODO: handle too big x values
                     throw new SparksoniqRuntimeException("The count value is too big to convert to integer type.");
                 } else {
-                    return ItemFactory.getInstance().createIntegerItem(count.intValue());
+                    return ItemFactory.getInstance().createIntegerItem((int) count);
```
ghislainfourny: A side note for later: we may want at some point to make integer items 64-bit rather than 32, so as to avoid these conversions.

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
```diff
@@ -31,68 +35,87 @@
...
-        this.sequence = seq;
+        _variableName = "$" + variableName;
+        _sequence = seq;
+    }
+
+    @Override
+    protected boolean initIsRDD() {
+        // TODO: figure out how to set isRDD for varRefIter.
+        //return _currentDynamicContext.isRDD(_variableName, getMetadata());
+        return false;
```
CanBerker: @ghislainfourny Do you have any tips on how isRDD should be determined for VariableReferenceIterator?

src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/ObjectLookupIterator.java
```diff
@@ -142,7 +139,7 @@ protected void closeLocal() {
 
     @Override
     public Item nextLocal() {
-        if (_hasNext == true) {
+        if (_hasNext) {
```
ghislainfourny: Can spotless catch these things automatically, or was it manual? Just curious.

CanBerker: I don't know if spotless has a configuration for this, but this one was manual

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
```diff
@@ -31,68 +35,87 @@
...
         super(null, iteratorMetadata);
-        this._variableName = "$" + variableName;
-        this.sequence = seq;
+        _variableName = "$" + variableName;
+        _sequence = seq;
+    }
+
+    @Override
+    protected boolean initIsRDD() {
+        // TODO: figure out how to set isRDD for varRefIter.
```
ghislainfourny: This is related to a discussion we had previously on dynamic function calls, which have the same issue: isRDD is actually context dependent, although for many expressions it is not (which is why, for now, we did not pass the context here). Could you open an issue to address it separately?

src/main/java/sparksoniq/jsoniq/runtime/iterator/postfix/ObjectLookupIterator.java
ghislainfourny: Very good. Thanks!

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
ghislainfourny: My comment crossed your question. :)

CanBerker:
This is the only failing test at the moment. It can be disabled and this PR can be merged if we're going to address the isRDD issue you have mentioned in a separate PR.

let $i := json-file("./src/main/resources/queries/conf-ex.json") return count($i)
(: first let clause with an RDD(gets materialized) :)

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
CanBerker: Thanks a lot! I will open the issue. I believe this issue will be the last mile in fully supporting DF&RDD enabled initial let clauses.

ghislainfourny:
> This is the only failing test at the moment. It can be disabled and this PR can be merged if we're going to address the isRDD issue you have mentioned in a separate PR.
> 
> let $i := json-file("./src/main/resources/queries/conf-ex.json") return count($i)
> (: first let clause with an RDD(gets materialized) :)

Is this test a regression, or a new test?

CanBerker:
Funnily enough it is a regression on the test, however, it can also be seen as a new test. We actually supported the initial let clauses operating on RDDs by instantly materializing them. So for the new implementation's purposes, it is a new test, but it is regressing as the instant materialization is not performed anymore.

ghislainfourny:
> Funnily enough it is a regression on the test, however, it can also be seen as a new test. We actually supported the initial let clauses operating on RDDs by instantly materializing them. So for the new implementation's purposes, it is a new test, but it is regressing as the instant materialization is not performed anymore.

Thank you. I see. What is the new behavior of the test: error, or different result?

CanBerker:
Currently errors out as the RDD variable needs to be looked up, but since the RDD status of VariableReferenceIterator is not detected, this does not happen. Local lookup can't find the variable $i

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
```diff
@@ -31,68 +35,87 @@
...
     }
 
     @Override
-    public void reset(DynamicContext context) {
-        super.reset(context);
-        this.currentIndex = 0;
-        this.items = null;
+    public void openLocal() {
+        _currentIndex = 0;
+        _items = _currentDynamicContext.getLocalVariableValue(_variableName, getMetadata());
```
ghislainfourny: Thank you. Could you help me understand why, here, it does not automatically materialize the RDD inside the dynamic context?

I remember that we had implemented these methods as an automatic cascade, i.e., the dynamic context would automatically materialize the value if it is an RDD. Am I missing something?

CanBerker: Could you take a look at the CountFunctionIterator? I believe that may answer your question as the execution in the test case flows through there. 

Ps: I have to run somewhere now, we can pick up this discussion on Monday when you're available. Have a nice weekend!

ghislainfourny: Happy to do so. Have a nice weekend!

ghislainfourny: (And for taking it from there on Monday, here is my fix: https://github.com/RumbleDB/rumble/pull/384 )

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
CanBerker: Issue created at #385

CanBerker:
Just to double-check, does this look good to go ?

ghislainfourny:
> Just to double-check, does this look good to go ?

Yes it is -- I didn't pay attention this question was asked on a different PR.

Thanks!

PR #384, opened 2019-11-22
Fix test.

ghislainfourny:
Fixing the test. The dynamic context should automatically cascade the count from the RDD.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
@@ -20,12 +20,13 @@
 
 package sparksoniq.jsoniq.runtime.iterator.functions.sequences.aggregate;
 
+import org.rumbledb.api.Item;
 import sparksoniq.exceptions.IteratorFlowException;
 import sparksoniq.exceptions.SparksoniqRuntimeException;
 import sparksoniq.jsoniq.item.ItemFactory;
 import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
-import sparksoniq.jsoniq.runtime.iterator.primary.VariableReferenceIterator;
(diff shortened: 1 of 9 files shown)


src/main/java/sparksoniq/semantics/DynamicContext.java
```diff
@@ -202,6 +202,9 @@ public Item getLocalVariableCount(String varName) {
         if (_localVariableCounts.containsKey(varName)) {
```
ghislainfourny: The name of this function could simply by getVariableCount(). It just returns the count irregardless of whether the variable is local, RDD or DF.

CanBerker: Very nice ! I like the name and the functionality

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
```diff
@@ -52,7 +52,7 @@ public Item next() {
                 if (iterator instanceof VariableReferenceIterator) {
                     VariableReferenceIterator expr = (VariableReferenceIterator) iterator;
                     this._hasNext = false;
-                    return _currentDynamicContext.getLocalVariableCount(expr.getVariableName());
+                    return _currentDynamicContext.getVariableCount(expr.getVariableName());
                 }
```
CanBerker: I feel that this entire if block with the instanceof check can be taken out of the if(!iterator.isRDD()) block. The current implementation with getVariableCount materializes through count() or size() in any scenerio, making the isRDD check redundant for the execution of VariableReferenceIterator within CountFunctionIterator.

ghislainfourny: This is a very good point. I have reorganized the cascade accordingly.

I also removed a few elses, which are unnecessary when the blocks return from the method.

ghislainfourny:
@CanBerker feel free to merge this into your branch, and then we can proceed with merging your branch into the master.

PR #387, opened 2019-11-25
Start renaming packages and deprecated ShellStart.

ghislainfourny:
In the long term we'll rename the packages to rumble instead of sparksoniq.

Giving that many people are working on the codebase, we can take this slowly, starting with barely touched files.


pom.xml
```diff
@@ -66,7 +66,7 @@
                     <archive>
                         <manifest>
                             <addClasspath>true</addClasspath>
-                            <mainClass>sparksoniq.Main</mainClass>
+                            <mainClass>org.mongodb.Main</mainClass>
```
CanBerker: I know it's not review ready yet and I may be missing something here but shouldn't this be rumbledb? I didn't understand the mongodb part

ghislainfourny: Absolutely. Good catch!

PR #388, opened 2019-11-25
Improve iterators performances

AndreaRinaldi1:
In this branch TreatIterator is changed so that, when the sequenceType arity is One or OneOrZero, we use the local iterator API. This way, the item will not be collected twice (once by the treat as iterator, and a second time later on in the iterator chain), but only once with the take action. "Branches" in the RDD dataflow are generally costly. These changes are based on the [discussion](https://github.com/RumbleDB/rumble/pull/360#issuecomment-555527545) with @ghislainfourny in PR #360 . 

The other performance improvements come from replacing count() action with take(2).size() when an exact count of the items is not necessary, but we just need to know if there is more than one item in the RDD.


src/main/java/sparksoniq/JsoniqQueryExecutor.java
```diff
@@ -287,7 +287,7 @@ private String getIteratorOutput(RuntimeIterator iterator) {
     private String getRDDResults(RuntimeIterator result) {
         JavaRDD<Item> rdd = result.getRDD(new DynamicContext());
         JavaRDD<String> output = rdd.map(o -> o.serialize());
-        long resultCount = output.count();
+        long resultCount = output.take(2).size();
```
ghislainfourny: Very nice catch.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/value/DistinctValuesFunctionIterator.java
```diff
@@ -110,7 +110,7 @@ public void setNextResult() {
         _currentDynamicContext = dynamicContext;
         JavaRDD<Item> childRDD = _sequenceIterator.getRDD(dynamicContext);
         Function<Item, Boolean> transformation = new FilterNonAtomicClosure();
-        if (childRDD.filter(transformation).count() == 0) {
+        if (childRDD.filter(transformation).isEmpty()) {
```
ghislainfourny: Also nice catch.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/InstanceOfIterator.java
```diff
@@ -70,8 +73,10 @@ public Item next() {
             } else {
                 JavaRDD<Item> childRDD = _child.getRDD(_currentDynamicContext);
                 this._hasNext = false;
-                if (isInvalidArity(childRDD.count()))
+
+                if (isInvalidArity(childRDD.take(2).size()))
```
ghislainfourny: Good improvement too.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
```diff
@@ -153,7 +165,9 @@ private void checkItemsSize(long size) {
 
     @Override
     public boolean initIsRDD() {
-        return _iterator.isRDD();
+        return _sequenceType.getArity() != SequenceType.Arity.One
```
ghislainfourny: Perfect. Nice and clean.

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
```diff
@@ -23,6 +23,7 @@
     private static final long serialVersionUID = 1L;
     private RuntimeIterator _iterator;
     private Item _nextResult;
+    private Item _currentResult;
```
ghislainfourny: A cosmetic (and, to some extent, performance-related) comment: since the only test involving this variable is whether it is null, a boolean may suffice.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/general/TreatAsClosure.java
```diff
@@ -27,6 +27,12 @@ public Boolean call(Item input) throws Exception {
                         + sequenceType.getArity().getSymbol(),
                     metadata
             );
+        if (input != null && sequenceType.isEmptySequence())
```
ghislainfourny: Can input ever be null (here and above)? If the input RDD is empty, call() will not be called at all...

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
AndreaRinaldi1: Using the _\_currentResult_ item, we can also avoid creating a new short-lived _Item result_ object in every iteration of the loop, and use _\_currentResult_ to store the value of _nextResult_ before the next() call. I agree that we can also go back to using a new inside-loop variable as before and using a boolean to check whether we are in the first iteration of the loop. I have no strong preference for either of the two options here. If you feel the latter is better, I will do that.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/general/TreatAsClosure.java
AndreaRinaldi1: That's right, there is no real need for this extra check. Thank you

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TreatIterator.java
ghislainfourny: Your reasoning makes a lot of sense. I like to see this as an optimization.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/general/TreatAsClosure.java
ghislainfourny: Thanks a lot. I cannot see any pushed commits removing the "input != null" part though (the branch looks unchanged). Is it an overlook?

AndreaRinaldi1: No it's not an overlook, I was waiting to reach an agreement on the "_currentResult_" discussion to push both changes.

PR #392, opened 2019-11-27
Add count calculation for DF

CanBerker:
This PR adds a faster DF fallback for the count calculation by avoiding the DF -> RDD conversion.


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/aggregate/CountFunctionIterator.java
```diff
@@ -64,8 +64,12 @@ public Item next() {
                 return ItemFactory.getInstance().createIntegerItem(results.size());
             }
 
-            // it is an RDD
-            long count = iterator.getRDD(_currentDynamicContext).count();
+            long count;
```
ghislainfourny: This is excellent. Exactly the right mindset.

ghislainfourny:
@CanBerker how about a test? (like count(parquet-file(...)) and count(structured-json-file...)

PR #393, opened 2019-11-27
Add dataframe support to keys() function

CanBerker:
Only works with top level keys, needs nesting implementation.

@ghislainfourny I was investigating the DataFrame extensions for iterators recently and I have a questionn. The parquetFileFunctionIterator and structuredJsonFileFunctionIterator read DataFrames resulting in DataTypes of native Java types. However, these are never converted into Items and not serialized for that matter. Don't we need to generate the Item types with these iterators as well, just as we did with the iterators of FLWOR clauses previously?


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/object/ObjectKeysFunctionIterator.java
```diff
@@ -118,15 +135,35 @@ public void setNextResult() {
...
+        JavaRDD<Item> childRDD = _iterator.getRDD(context);
         FlatMapFunction<Item, Item> transformation = new ObjectKeysClosure();
         return childRDD.flatMap(transformation).distinct();
     }
 
     @Override
-    public boolean initIsRDD() {
-        return _iterator.isRDD();
+    public Dataset<Row> getDataFrame(DynamicContext context) {
+        _currentDynamicContext = context;
```
ghislainfourny: That's a very good place to start.

Now, thinking further, I think that getting the keys of a DataFrame is much faster than getting those inside an RDD, because (i) they can directly be read from the schema into a local list and (ii) and it will always be a small list because this is homogeneous data (so outputting as a DF is always an overkill).

How about having the keys simply be output via the local iterator (similar to count()) if the child iterator is a DF?

I.e.:

- isRDD() and isDF() would both return false if the child iterator is a DF
- in the local iterator, test for _iterator.isDF(), and if so, collect the keys from the schema locally to a local list, and return them one by one (i.e., a big switch in setNextResult() for the local child case and the DF child case).

That way, no heavy machinery is needed to populate a new (and small) DF.

Does it make sense?

getRDDAux() can and should of course be left unchanged, because if the input is an RDD, we may have a very heterogeneous sequence with as many keys as objects :)

This discussion is specific to DataFrames.

CanBerker:
> The mindset is that DF should use the native types directly as is the case now: we are looking at highly structured data. So, when we generate new DFs from an iterator, we can stick to the native DF types.
 
The JSON lines format and the respective read methods of DF reader for json files allow nested objects and arrays as expected. As we discussed a couple of months ago, columns which contain such JSON types, or columns that have multiple types in them are cast into strings. By our assumption above, we're expecting the user to deal with these when they read a parquet-file or a structured-json-file. Is this correct? 

If so, should we try and add a log message for the user when a mismatch of type found in one of the columns resulting in a string column for parquet-file and structured-json-file functions?

ghislainfourny:
> > The mindset is that DF should use the native types directly as is the case now: we are looking at highly structured data. So, when we generate new DFs from an iterator, we can stick to the native DF types.
> 
> The JSON lines format and the respective read methods of DF reader for json files allow nested objects and arrays as expected. As we discussed a couple of months ago, columns which contain such JSON types, or columns that have multiple types in them are cast into strings. By our assumption above, we're expecting the user to deal with these when they read a parquet-file or a structured-json-file. Is this correct?

Yes, this is correct. It is the responsibility of the user to only call the structured-json-file() function on a JSON dataset that they know is highly structured. If they have any doubt, then they should use json-file(). Maybe an explicit hint in the documentation explaining this could remove any confusion. In the future, we could provide a function called around json-file() that returns a boolean that says whether it is highly structured or not.

> If so, should we try and add a log message for the user when a mismatch of type found in one of the columns resulting in a string column for parquet-file and structured-json-file functions?

Yes, this is a good idea -- are you aware of any mechanism through which Spark tells us when this happens?

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/object/ObjectKeysFunctionIterator.java
```diff
@@ -118,15 +107,45 @@ public void setNextResult() {
...
-    public boolean initIsRDD() {
-        return _iterator.isRDD();
+    protected boolean hasNextLocal() {
+        return _hasNext;
+    }
+
+    @Override
+    protected void resetLocal(DynamicContext context) {
+        _iterator.reset(_currentDynamicContext);
+        setResultsFromNextObjectItem();
```
ghislainfourny: Here the same switch should be made as in openLocal(), as it could be a DF.

CanBerker:
>>If so, should we try and add a log message for the user when a mismatch of type found in one of >>the columns resulting in a string column for parquet-file and structured-json-file functions?
>Yes, this is a good idea -- are you aware of any mechanism through which Spark tells us when this >happens?
Unfortunately I couldn't find any such mechanism.

> In the future, we could provide a function called around json-file() that returns a boolean that says > whether it is highly structured or not.

That's a great idea. It might be a very viable option to alleviate the issue above as well.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/object/ObjectKeysFunctionIterator.java
CanBerker: Fixed now

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/object/ObjectKeysFunctionIterator.java
```diff
@@ -132,8 +132,16 @@ protected boolean hasNextLocal() {
...
     protected void resetLocal(DynamicContext context) {
+        _alreadyFoundKeys = new ArrayList<>();
+        _nextResults = new LinkedList<>();
         _iterator.reset(_currentDynamicContext);
-        setResultsFromNextObjectItem();
+
+        if (_iterator.isDataFrame()) {
+            setResultsFromDF();
+        } else {
+            _iterator.open(_currentDynamicContext);
```
ghislainfourny: There is still a redundancy with _iterator.reset() and _iterator.open(). A reset is logically equivalent to close+open, so that this line is in fact re-opening an already open iterator. Also, it is not necessary to reset _iterator if its DF API was used and will continue to be used. Maybe move reset to here and remove open?

docs/Function library.md
```diff
@@ -618,7 +618,7 @@ return $my-json
 
 ### structured-json-file (Rumble specific)
 
-Parses one or more json files that follow [JSON-lines](http://jsonlines.org/) format and returns a sequence of objects. This enables better performance with fully structured data and is recommended to use when such data is available. This is also similar to Spark's spark.read.json().
+Parses one or more json files that follow [JSON-lines](http://jsonlines.org/) format and returns a sequence of objects. This enables better performance with fully structured data and is recommended to use only when such data is available. When data has nested structure or multiple types for the same field, this field and contained values will be treated as strings. This is also similar to Spark's spark.read.json().
```
ghislainfourny: Here, nested structures will be handled well by this function (nestedness is 100% DataFrames-compatible)  -- only multiple types are an issue.

PR #394, opened 2019-11-27
Fix order clause empty df handling

CanBerker:
Based on #393. This PR introduces a tiny change that fixes the crashes caused by empty dataframe being passed as input to order by clause. This would result in an empty sql statement which causes the crash.

Tests added.

@ghislainfourny Should the short circuit mentioned above (seen in OrderByClauseSparkIterator.java file) be added to other FLWOR clauses as well? It could save a bit of unnecessary computation and querying when operating on empty sequences. However this introduces an additional count() action which might be costly.

### DIFF
--- a/src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/OrderByClauseSparkIterator.java
@@ -187,6 +187,9 @@ public class OrderByClauseSparkIterator extends RuntimeTupleIterator {
             throw new SparksoniqRuntimeException("Invalid orderby clause.");
         }
         Dataset<Row> df = _child.getDataFrame(context, getProjection(parentProjection));
+        if (df.count() == 0) {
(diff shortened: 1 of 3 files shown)


ghislainfourny:
> Based on #393. This PR introduces a tiny change that fixes the crashes caused by empty dataframe being passed as input to order by clause. This would result in an empty sql statement which causes the crash.
> 
> Tests added.
> 
> @ghislainfourny Should the short circuit mentioned above (seen in OrderByClauseSparkIterator.java file) be added to other FLWOR clauses as well? It could save a bit of unnecessary computation and querying when operating on empty sequences. However this introduces an additional count() action which might be costly.

Here I think performance measurements would tell which is better. For an empty DF, the closures would not be called anyway, so that the count() action may be more costly than the gain. Maybe you could open an issue on this?

CanBerker:
Fortunately, there were no crashes with the other clauses. 

I have generated the issue for performance investigations of empty DFs for later.

PR #395, opened 2019-11-28
Add context parameter to isRDD isDF and initisRDD methods

CanBerker:
- Refactored RuntimeIterator::isRDD(), RuntimeIterator::isDataFrame() and RuntimeTupleIterator::isDataFrame() methods to have DynamicContext parameter.
- Renamed getIterators() method in JsoniqQueryExecutor to getLocalExecutionResults()

@ghislainfourny  there are still some issues we need to address with VariableReferenceIterator's isRDD status. 
The following query gets locally evaluated now, even though, for clause iterates on an RDD
>let $data := json-file("./src/main/resources/queries/conf-ex.json")
>for $i in $data
>return $i

It seems to me that, in the presence of a variable reference iterator, we can't know whether some query should be executed locally or with spark, until we actually evaluate that variable. Prior to that, we're in a 'maybe' state with regard to VariableReferenceIterator. However, picking a default value for the 'maybe' state is not satisfactory either in my opinion. If we were to pick isRDD=false as the default value for 'maybe' state (which is the current implementation), the above query doesn't utilize RDDs. 

In the current state the following query gets locally evaluated (please confirm whether this is the desired behavior or not)
>let $data := json-file("./src/main/resources/queries/conf-ex.json")
>let $var := $data
>return $var

If we were to select isRDD=true as the default value for 'maybe' state, then this query would be evaluated with the RDD API.


ghislainfourny:
> * Refactored RuntimeIterator::isRDD(), RuntimeIterator::isDataFrame() and RuntimeTupleIterator::isDataFrame() methods to have DynamicContext parameter.
> * Renamed getIterators() method in JsoniqQueryExecutor to getLocalExecutionResults()
> 
> @ghislainfourny there are still some issues we need to address with VariableReferenceIterator's isRDD status.
> The following query gets locally evaluated now, even though, for clause iterates on an RDD
> 
> > let $data := json-file("./src/main/resources/queries/conf-ex.json")
> > for $i in $data
> > return $i
> 
> It seems to me that, in the presence of a variable reference iterator, we can't know whether some query should be executed locally or with spark, until we actually evaluate that variable. Prior to that, we're in a 'maybe' state with regard to VariableReferenceIterator. However, picking a default value for the 'maybe' state is not satisfactory either in my opinion. If we were to pick isRDD=false as the default value for 'maybe' state (which is the current implementation), the above query doesn't utilize RDDs.
> 
> In the current state the following query gets locally evaluated (please confirm whether this is the desired behavior or not)
> 
> > let $data := json-file("./src/main/resources/queries/conf-ex.json")
> > let $var := $data
> > return $var
> 
> If we were to select isRDD=true as the default value for 'maybe' state, then this query would be evaluated with the RDD API.

Yes, indeed. Many thanks for creating this PR to give us a concrete understanding of the issue, this helps a lot.

I will get back to you on this next week. I think that the way to overcome the issue is to do the determination between DF, RDD or local statically, via the static context, by extending each variable in the static context with a DF-RDD-or-local flag. Then, the static context (which each runtime iterator will have) can be read by initRDD in the variable-ref to know if each variable is local, an RDD or a DF.

In other terms: isRDD() would be context-dependent, but only statically, not dynamically.

ghislainfourny:
As discussed today we will discuss the static context approach next week.

ghislainfourny:
@CanBerker do you agree we can now drop this "investigation" PR?

CanBerker:
Absolutely, cleaning this up now. We will continue development in #415 \

PR #398, opened 2019-12-02
Use EBV in where clauses and predicates when pushing down to DF.

ghislainfourny:
The EBV was not considered, although where clauses and predicates should.

https://www.w3.org/TR/xquery-30/#id-ebv

For predicates, there is the special case where the sequence is a single numeric item. Then the EBV is not taken, but the position is checked instead.

https://www.w3.org/TR/xquery-30/#id-filter-expression


src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -53,6 +53,7 @@ public boolean hasNextLocal() {
     @Override
     public void resetLocal(DynamicContext context) {
         _iterator.reset(_currentDynamicContext);
+        this._childIndex = 0;
```
ghislainfourny: @AndreaRinaldi1 this was probably an overlook.

ghislainfourny:
Thank you!

PR #400, opened 2019-12-02
Primitive types builtin functions

AndreaRinaldi1:
In this branch I implement the builtin function for string, integer, decimal and double.

You can now instantiate such item like so:
integer(1.234),
decimal(5.67),
double(8e9),
string(date("2001-12-12-10:00"))


src/main/resources/test_files/runtime/FunctionNumerics/NumericFunctions.jq
```diff
@@ -0,0 +1,13 @@
+(:JIQS: ShouldRun; Output="(1, 2, 3456, 7, 8.9, 1011.12, 13, 14.15, 1617.18)" :)
+integer(1),
```
ghislainfourny: Very nice. Could we also test casting from string? like integer("1"), etc.

PR #401, opened 2019-12-02
Remove additional late typechecks

AndreaRinaldi1:
In this branch I remove the additional typechecks that are performed after calling (the old version of) _getSingleItemOfTypeFromIterator()_. 
Moreover, that function is now not checking for the (previously) specified item type and it is not performing any cast. 

This is allowed by the redesigned builtin function definition in the catalogue, that is in particular specifying the arguments and return types, and thanks to the builtin type checking and type promotion introduced in #381 .


ghislainfourny:
@AndreaRinaldi1 Out of curiosity, is there a noticeable performance improvement on function calls with actual type checks?

AndreaRinaldi1:
@ghislainfourny As for the performance comparison between the old master and the new builtin function with early typecheck on arguments and return type, the results are as follows:
- the branch from #381 was seeing an increase in avg execution time of 15.5% compared to the previous master
- this branch let us reduce the performance drop by ~2 percentage points, as it has an increase of 13.6% in avg execution time compared to the old master. 

The results that I shared in [this](https://github.com/RumbleDB/rumble/pull/381#issuecomment-559235241) comment refers to this branch indeed (i.e. the one without the late checks already).

Overall, that drop of 13.6% in performance that this experiment showed, is the worst case scenario, since every argument and return type need to be checked --and sometimes promoted--.

In another experiment that I explained in [this other comment](https://github.com/RumbleDB/rumble/pull/381#issuecomment-560464996) I confirm that when there is no need for typechecking (nor promotion) the performance of **this** branch and master are not different.

PR #405, opened 2019-12-05
Typecheck user-defined functions

AndreaRinaldi1:
In this branch I implement the early typecheck for arguments and ultimately for return type for user-defined functions. 

Just like for builtin functions, TypePromotionIterator wraps every argument that is not null and that has a SequenceType type definition in the function signature different from Item*. 
Analogously, it also wraps the whole function to check on its return type if different from Item*. 

As you can see from the tests, this works both for static user-defined function calls and dynamic function calls.


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
```diff
@@ -23,6 +23,7 @@
 import org.apache.spark.api.java.JavaRDD;
 import org.rumbledb.api.Item;
 import sparksoniq.exceptions.IteratorFlowException;
+import sparksoniq.exceptions.SequenceExceptionExactlyOne;
```
wscsprint3r: Is this needed?

src/main/java/sparksoniq/jsoniq/runtime/iterator/operational/TypePromotionIterator.java
```diff
@@ -161,6 +159,8 @@ public boolean initIsRDD() {
     }
 
     private void checkTypePromotion() {
+        if (_nextResult.isFunction())
```
ghislainfourny: Good catch. Thanks!

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -579,7 +579,20 @@ public static RuntimeIterator buildUserDefinedFunctionCallIterator(
             IteratorMetadata metadata,
             List<RuntimeIterator> arguments
     ) {
-        return new FunctionItemCallIterator(functionItem, arguments, metadata);
+        FunctionItemCallIterator functionCallIterator = new FunctionItemCallIterator(functionItem, arguments, metadata);
```
ghislainfourny: Perfect!

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
```diff
@@ -92,6 +95,25 @@ private void processArguments() {
             );
         }
 
+        if (_functionItem.getSignature().getParameterTypes() != null) {
+            for (int i = 0; i < _functionArguments.size(); i++) {
+                if (
+                    _functionArguments.get(i) != null
+                        && !_functionItem.getSignature().getParameterTypes().get(i).equals(mostGeneralSequenceType)
+                ) {
+                    _functionArguments.set(
```
ghislainfourny: Very good and clean.

src/main/java/sparksoniq/exceptions/UnknownFunctionCallException.java
```diff
@@ -39,4 +39,8 @@ public UnknownFunctionCallException(String fnName, int arity, IteratorMetadata i
             iteratorMetadata.getExpressionMetadata()
         );
     }
+
+    public UnknownFunctionCallException(String message, IteratorMetadata iteratorMetadata) {
```
ghislainfourny: Where is this used? I cannot see any new call in the diff.

src/main/java/sparksoniq/semantics/visitor/RuntimeIteratorVisitor.java
```diff
@@ -430,23 +427,12 @@ public RuntimeIterator visitFunctionDeclaration(FunctionDeclaration expression,
...
         FunctionItem function = new FunctionItem(
-                new FunctionIdentifier(expression.get_name(), parameters.size()),
-                parametersNames,
-                new FunctionSignature(
-                        parameters,
-                        returnType
-                ),
-                bodyIterator,
-                new HashMap<>()
+                expression.get_name(),
```
ghislainfourny: Very nice, indeed this is a code duplication.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
AndreaRinaldi1: No, probably just something added by mistake. 
Thank you

src/main/java/sparksoniq/exceptions/UnknownFunctionCallException.java
AndreaRinaldi1: You are right, I probably added that when I added the exception for non-existing function, but then I decided to stick to the existing one. 
Thank you for spotting

Issue #244, opened 2019-04-29
Variable values not passed to parallel execution


ghislainfourny:
for $rank in 1 to 4
return {
  "Rank" : $rank,
  "Count" : count(
    for $i in json-file("/tmp/input.json", 100)
    where $i.target eq $i.choices[[$rank]]
    return 1
  )
}

Error [err: XPDY0130 ] Runtime error retrieving variable $rank value

Another case:

let $guess := (for $i in json-file("confusion-2014-03-02/confusion-2014-03-02.json", 10)
							order by $i.guess
							count $c
							where $c<=1
							return $i.guess)

return count(for $i in json-file("confusion-2014-03-02/confusion-2014-03-02.json", 10)
							where $i.target = $guess
							return 1)

CanBerker:
The last query fails with a parser exception:

Error [err: XPST0003]LINE:4:COLUMN:0:Parser failed. 

return count(for $i in json-file("./src/main/resources/queries/conf-ex.json", 10)
where $i.target = $guess
return 1)

ghislainfourny:
> The last query fails with a parser exception:
> 
> Error [err: XPST0003]LINE:4:COLUMN:0:Parser failed.
> 
> return count(for $i in json-file("./src/main/resources/queries/conf-ex.json", 10)
> where $i.target = $guess
> return 1)

This is the second part of the upper query -- the query is

let $guess := (for $i in json-file("confusion-2014-03-02/confusion-2014-03-02.json", 10)
order by $i.guess
count $c
where $c<=1
return $i.guess)
return count(for $i in json-file("confusion-2014-03-02/confusion-2014-03-02.json", 10)
where $i.target = $guess
return 1)

CanBerker:
Thanks for the clarification

Issue #404, opened 2019-12-05
Undeterministic failure of count clause


ghislainfourny:
Sometimes, using the count clause fails with a null pointer exception because some partition IDs do not get assigned a partition offset in the underlying Spark SQL algorithm.

Reproduceable (not always) with:

for $i in json-file("confusion-2014-03-02/confusion-2014-03-02.json", 10)
    order by $i.date descending
    count $c
    where $c <= 1
    return $i.date

CanBerker:
Solved in #406

PR #406, opened 2019-12-05
Fix contexts for passing outer variables.

ghislainfourny:
Fixes https://github.com/RumbleDB/rumble/issues/244

### DIFF
--- a/src/main/java/sparksoniq/spark/DataFrameUtils.java
+++ b/src/main/java/sparksoniq/spark/DataFrameUtils.java
@@ -31,9 +31,25 @@ import org.apache.spark.sql.expressions.Window;
 import org.apache.spark.sql.types.DataTypes;
 import org.apache.spark.sql.types.StructType;
 import org.rumbledb.api.Item;
-
 import scala.collection.mutable.WrappedArray;
-import sparksoniq.jsoniq.item.*;
+import sparksoniq.jsoniq.item.ArrayItem;
+import sparksoniq.jsoniq.item.Base64BinaryItem;
+import sparksoniq.jsoniq.item.BooleanItem;
+import sparksoniq.jsoniq.item.DateItem;
+import sparksoniq.jsoniq.item.DateTimeItem;
+import sparksoniq.jsoniq.item.DayTimeDurationItem;
+import sparksoniq.jsoniq.item.DecimalItem;
+import sparksoniq.jsoniq.item.DoubleItem;
+import sparksoniq.jsoniq.item.DurationItem;
+import sparksoniq.jsoniq.item.FunctionItem;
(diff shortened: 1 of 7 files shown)


ghislainfourny:
@CanBerker this is a fix when outer variables were not bound correctly in some inner clauses (see tests). I fixed it for where and return, but I think there is the same issue for group/order/for and possibly predicates, which should be solvable with the same approach. It would be great if you could take it from there.

The persist() call is hard to test; it is an issue that happens indeterministically on very large data, because the partitions change every time otherwise, leading to null pointer exceptions.

CanBerker:
@ghislainfourny Thanks for catching this oversight with contexts! I will pick this up and apply the same idea to other iterators. 

> The persist() call is hard to test and is addressing an issue with count clauses; 
I understand the persist() call solves the said problem with count completely. Is this correct ?

CanBerker:
I have made a few cosmetic changes mostly converting variables with default access to private and using constructors for sets and maps instead of a consequent putAll operation

CanBerker:
@ghislainfourny I have just noticed one thing from the related issue and pointed it out [here](https://github.com/RumbleDB/rumble/issues/244#issuecomment-562526150)

ghislainfourny:
> @ghislainfourny Thanks for catching this oversight with contexts. I will pick this up and apply the same idea to other iterators.
> 
> > The persist() call is hard to test and is addressing an issue with count clauses;
> 
> I understand the persist() call solves the said problem with count completely. Is this correct ?
> Edit: If so shall we mark this PR's relation with #404 ?

It does. Yes, this fixes #404  .

ghislainfourny:
> @ghislainfourny I have just noticed one thing from the related issue and pointed it out [here](https://github.com/RumbleDB/rumble/issues/244#issuecomment-562526150) It might be good to verify that query example is fully functional as well.

Thank you for the thorough check. The three lines are the second part of the upper query; a query cannot start with a return clause. Does it address your concern?

ghislainfourny:
> I have made a few cosmetic changes mostly converting variables with default access to private and using constructors for sets and maps instead of a consequent putAll operation

I like these changes. Thanks a lot!

ghislainfourny:
> @ghislainfourny Thanks for catching this oversight with contexts. I will pick this up and apply the same idea to other iterators.

@CanBerker thanks a lot!

CanBerker:
> > @ghislainfourny I have just noticed one thing from the related issue and pointed it out [here](https://github.com/RumbleDB/rumble/issues/244#issuecomment-562526150) It might be good to verify that query example is fully functional as well.
> 
> Thank you for the thorough check. The three lines are the second part of the upper query; a query cannot start with a return clause. Does it address your concern?

Exactly, I was confused by that query due to that misunderstanding. It's all good now :)

Issue #403, opened 2019-12-05
Null pointer exception on cartesian product of two for clauses


ghislainfourny:
This query returns a null pointer exception:

for $i in parallelize(1) for $l in (for $i in parallelize(1) return 1)
return $i

If a for clause supports RDDs and does not consume any variables bound by previous clauses (which we can now easily check with the dependency machinery), then it can be supported with a Cartesian product in Spark. Otherwise, an error should be thrown to avoid the null pointer exception.

PR #407, opened 2019-12-06
Local and Spark Enabled Execution of Nested FLWOR Expressions

CanBerker:
Local implementation seems to handle the nested FLWOR expressions well.

However, there are some issues with Spark enabled execution of nested FLWOR execution as listed in #403 

I believe the root cause of this issue is the same concept with the discussion that can be found here:
https://stackoverflow.com/questions/47358177/caused-by-java-lang-nullpointerexception-at-org-apache-spark-sql-dataset

In the case of test case NestedFLWORSpark2.iq shown below, no such problem arises:
```
(:JIQS: ShouldRun; Output="1" :)
for $i in parallelize(1) for $l in (let $i := parallelize(1) return 1)
return $i
```

This is due to the fact that letClause is locally evaluated, causing no nesting issues.

In other test cases which are on .HOLD at the moment, I think the usage of DataFrame operations within the execution of a DF UDF causes NullPointerExceptions. More precisely, in the query below:
```
for $i in parallelize(1) for $l in (for $i in parallelize(1) return 1)
return $i
``` 
The execution of  **expression** "(for $i in parallelize(1) return 1)" is triggered within the UDF of the for clause "for $l in **expr**". As **expr** is a spark enabled FLWOR expression, DataFrame execution of **expr** nests within DataFrame execution of ForClauseUDF, causing problems with Spark execution. 

I will give a bit more thought on if and how we can solve this issue. I would also appreciate your feedback on the matter @ghislainfourny


ghislainfourny:
@CanBerker many thanks for looking into it.

The idea is that there are two possible cases when a non-initial for expression supports RDD evaluation:

Case 1 : the for expression does not consume any variables bound in the same FLWOR

This is the case for all tests I see here.

Example

for $i in json-file(..)
for $j in json-file(...)
return { left: $i, right: $j }

-> in this case, you can invoke, in getDataFrame(), getRDD(), passing the current dynamic context, convert it to a DataFrame as if it were an initial for (maybe even using the same code), and then compute the cartesian product of the child dataframe and of this new dataframe.

Case 2 : the for expression does consume any variables bound in the same FLWOR

Example

for $i in json-file(..)
for $j in json-file(...)[$j.foo eq $i.foo]
return { left: $i, right: $j }

This is the case that cannot be solved in any easy way, because this is a nested job. In this case, which is easy to detect, we should properly throw an error, which is more user-friendly than letting it go all the way to the null pointer expression. We can create a new Rumble error code for this.

At a latter time, there are a few special cases that we can optimize into a SQL join, but we can leave this for later.

Does it make sense?

src/main/resources/test_files/runtime/NestedFLWORLocal3.iq
```diff
@@ -0,0 +1,5 @@
+(:JIQS: ShouldRun; Output="(2, 2, 3, 2, 3, 4)" :)
```
ghislainfourny: Yes, local FLWORS definitely work. This is specific to DF FLWORS.

src/main/resources/test_files/runtime-spark/NestedFLWORSpark2.iq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="1" :)
```
ghislainfourny: I am actually surprised that this one works, because parallelize() should trigger an RDD, I will double-check. It may be that the implementation of the parallelize iterator bypasses the RDD in case of a local evaluation -- which is not necessarily bad but at least we should understand what is going on, and that this is specific to parallelize.

CanBerker: This query also works.
jiqs$ for $i in parallelize(1) for $l in (let $k := json-file("./src/main/resources/queries/conf-ex.json") return $k)
> return $l
>>> 
>>> 
{ "guess" : "Latvian", "target" : "Russian", "country" : "AU", "choices" : [ "Lao", "Latvian", "Russian", "Swahili" ], "sample" : "b7df3f9d67cef259fbcaa5abcad9d774", "date" : "2013-08-20" }
{ "guess" : "Russian", "target" : "Russian", "country" : "AU", "choices" : [ "Croatian", "Nepali", "Russian", "Slovenian" ], "sample" : "8a59d48e99e8a1df7e366c4648095e27", "date" : "2013-08-20" }
{ "guess" : "Czech", "target" : "Czech", "country" : "SE", "choices" : [ "Maori", "Czech", "Korean", "Turkish" ], "sample" : "1787b5c79a00b3513ce76847bc1f5b75", "date" : "2013-08-20" }
{ "guess" : "Greek", "target" : "Serbian", "country" : "SE", "choices" : [ "German", "Greek", "Kannada", "Serbian" ], "sample" : "0d5b697ebb326b5043ce7fa60a7b968d", "date" : "2013-08-20" }
{ "guess" : "Serbian", "target" : "Serbian", "country" : "AU", "choices" : [ "Dari", "Serbian", "Sinhalese", "Vietnamese" ], "sample" : "0d5b697ebb326b5043ce7fa60a7b968d", "date" : "2013-08-20" }

Maybe I can be of help in understanding this one. The reason is that we support storing RDDs without materialization in let clauses now, The inner FLWOR expression, that is (let $i := parallelize(1) return 1), does not require a nested DataFrame/UDF operation

CanBerker: With our recent changes let clause is evaluated locally. This triggers local evaluation for return clause as well. This local evaluation triggers getLocalValue from DynamicContext for the VariableReferenceIteratorExpression in the body of the return clause.

Since no nesting takes place, the query doesn't crash

ghislainfourny: @CanBerker you are correct of course. Indeed, there is no actual Spark job triggered for $k. This is unexpectedly wonderful, because this expands our possibilities.

CanBerker:
Dear @ghislainfourny,

I believe I have asked for a review prematurely yesterday, please excuse my oversight there. Regarding your earlier comment, I have added the following tests (can also be seen in tests) with regards to what you have described under "case 1". These queries are already functional and tests are passing.

#### For cartesian
-----
```
for $i in json-file("./src/main/resources/queries/conf-ex.json")
for $j in json-file("./src/main/resources/queries/conf-ex.json")
return { left: $i.guess, right: $j.guess }
```
#### For cartesian with outer flwor clause
-----
```
let $k := "foo"
return
for $i in json-file("./src/main/resources/queries/conf-ex.json")
for $j in json-file("./src/main/resources/queries/conf-ex.json")
return { left: $i.guess, right: $j.guess, "$k": $k }
```

#### FLWOR query with predicate using FLWOR variable
-----
I believe there is an issue for the queries of the following form where a predicate uses a variable defined in the FLWOR. I will create an issue and address this in a new PR.

```
let $k := "foo"
return
for $i in json-file(..)
for $j in json-file(...)[$j.bar eq $k]
return { left: $i, right: $j }
```

#### Re-visiting issue 403
-----
Lastly, I would like to re-iterate over the problem that is given in issue #403 and make sure we are on the same page.
```
for $i in parallelize(1) 
    for $l in 
        (for $i in parallelize(1) return 1)
return $i
```
Unless locally evaluated each FLWOR expression internally uses DFs for their execution. The parallelize in the first for clause ensures that we are in non-local execution. During the execution of the outmost FLWOR expression, second for clause(*for $l in*)'s getDataFrame method tries to execute the expression given to this clause. This expression is yet another FLWOR expression, that is *(for $i in parallelize(1) return 1)* As this FLWOR expression too is Spark-enabled due to the parallelize call within it, it resorts to using DataFrames as well. Here the issue arises:
> As we evaluate an expression within the getDataFrame() call which internally uses the forClauseUDF, if said expression is another FLWOR expression and it is Spark Enabled, this triggers a nested DataFrame calculation aka. nested UDF calling. I believe this is the root cause of the nullpointer exception.

I believe my explanation was not clear enough the first time so I hope this helps. Please let me know what you think.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -170,53 +169,45 @@ public void close() {
...
-            StructType schema = DataTypes.createStructType(fields);
-
-            JavaRDD<Row> rowRDD = initialRdd.map(new ForClauseSerializeClosure());
-
-            // apply the schema to row RDD
-            return SparkSessionManager.getInstance().getOrCreateSession().createDataFrame(rowRDD, schema);
+            return getDataFrameFromRDDExpression(context);
         }
 
         if (_child.isDataFrame()) {
```
ghislainfourny: The old code should still be kept, but with the additional test that the expression depends on a variable bound by the current FLWOR (and throw an error if _expression.isRDD())

The code on the right should also be kept, but with the additional test that the expression does not depend on a variable bound by the current FLWOR.

Testing whether the expression does or not depend on a variable bound by the current FLWOR should be straightforward, because there are a few methods in clauses and iterator classes that return the variable dependencies. It is just a test whether an intersection is empty or not.

ghislainfourny:
> EDIT2: QUESTION: The two tests I have added (and shown below) on "cartesian product calculation with for clauses" work on the terminal application and local tests. However they time out on travis for some reason. I'm thinking of putting these on hold. Do you have any suggestions on this ?

This may be related to my comment -- what are these two tests exactly (there are more than two in your text)?

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: I will take a look at this now

CanBerker:
@ghislainfourny I added ordering clauses to a majority of forClause and WhereClause tests to get deterministic output from these. Do you see any issues with this ?

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: >(and throw an error if _expression.isRDD())

I remember you mentioning a custom Sparksoniq error message for this. Did you have a specific message in mind ?

ghislainfourny: Yes, I suggest increasing (0003). However, I would suggest replacing SPRKIQ0001 and SPRKIQ0002 with RBST0001, RBST0002 calling the new one RBST0003 to stick with error code conventions from the W3C (2 letters for technology, 2 letters for static/dynamic/typed/etc and 4 digits). Does it help?

ghislainfourny:
> @ghislainfourny I added ordering clauses to a majority of forClause and WhereClause tests to get deterministic output from these. Do you see any issues with this ?

No, this is a good idea! Back then we did not have all clauses in place. Where there undeterministic issues here?

ghislainfourny:
> Both of these tests seem to be fine at the moment

Very good, let me know if there are issues again with them.

src/main/resources/test_files/runtime-spark/DataFrames/ForClause2.jq
```diff
@@ -1,2 +1,2 @@
 (:JIQS: ShouldRun; Output="({ "$i" : 1, "$j" : 4 }, { "$i" : 1, "$j" : 5 }, { "$i" : 1, "$j" : 6 }, { "$i" : 2, "$j" : 4 }, { "$i" : 2, "$j" : 5 }, { "$i" : 2, "$j" : 6 }, { "$i" : 3, "$j" : 4 }, { "$i" : 3, "$j" : 5 }, { "$i" : 3, "$j" : 6 })" :)
-for $i in parallelize((1,2,3)) for $j in (4,5,6) return {"$i": $i, "$j": $j}
+for $i in parallelize((1,2,3)) for $j in (4,5,6) order by $i, $j return {"$i": $i, "$j": $j}
```
ghislainfourny: The change of order may be triggered by the switch to cartesian product -- since normally the order is supposed to be deterministic, this actually indicates that we should not use the cartesian product if the new for clause is local, but stick to the old code. That is, cartesian product only if (i) child is a DF, (ii) for expression does not depend on variables bound in current FLWOR and (iii) for expression is an RDD. For this case, we can live with indeterministic outputs. Does it make sense?

CanBerker: Yes, this seems very sound now

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -170,31 +171,69 @@ public void close() {
...
-            // define a schema
-            List<StructField> fields = new ArrayList<>();
-            StructField field = DataTypes.createStructField(_variableName, DataTypes.BinaryType, true);
-            fields.add(field);
-            StructType schema = DataTypes.createStructType(fields);
+        if (_child.isDataFrame()) {
+            if (_expression.isRDD()) {
+                Set<String> intersection = new HashSet<>(_expression.getVariableDependencies().keySet());
+                intersection.retainAll(getVariablesBoundInCurrentFLWORExpression());
+                boolean expressionUsesVariablesOfCurrentFlwor = !intersection.isEmpty();
```
ghislainfourny: Wonderful.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -170,31 +171,69 @@ public void close() {
...
+
+                if (expressionUsesVariablesOfCurrentFlwor) {
+                    throw new InvalidForClauseExpressionException(
+                            "An RDD-enabled for clause expression cannot reference the variables defined in the encapsulating, RDD-enabled FLWOR expression.",
```
ghislainfourny: I suggest this message, because RDD-enabled expressions is implementation terminology that we want to shield users from (some of our users are not computer scientists and do not know the details of Spark). The terminology we typically use for the end user is the notion of "big sequence" of items.

"A for clause expression cannot produce a big sequence of items for a big number of tuples, as this would lead to a data flow explosion."

What do you think?

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -170,31 +171,69 @@ public void close() {
...
+                            getMetadata().getExpressionMetadata()
+                    );
+                }
 
-            JavaRDD<Row> rowRDD = initialRdd.map(new ForClauseSerializeClosure());
+                // since no variable dependency to the current FLWOR expression exists for the expression
+                // evaluate the DF with the parent context and calculate the cartesian product
+                Dataset<Row> expressionDF;
+                if (_expression.isRDD()) {
```
ghislainfourny: We already know that the expression is an RDD at this point, so that the test is superfluous and the else clause can go away.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -170,31 +171,69 @@ public void close() {
...
+                        String.format(
+                            "select %s from %s, %s",
+                            selectSQL,
+                            inputDFTableName,
+                            expressionDFTableName
+                        )
+                    );
+            }
 
+            Dataset<Row> df = this._child.getDataFrame(context, getProjection(parentProjection));
```
ghislainfourny: Can we add a comment // the expression is locally evaluated

src/main/java/sparksoniq/exceptions/InvalidForClauseExpressionException.java
```diff
@@ -0,0 +1,33 @@
...
+ * Authors: Stefan Irimescu, Can Berker Cikis
+ *
+ */
+
+package sparksoniq.exceptions;
+
+import sparksoniq.exceptions.codes.ErrorCodes;
+import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
+
+public class InvalidForClauseExpressionException extends SparksoniqRuntimeException {
```
ghislainfourny: JobWithinAJobException?

docs/Error codes.md
```diff
@@ -18,11 +18,18 @@ for an object or array selector.
...
 
-- [SPRKIQ0001] - CLI error. Raised when invalid parameters are supplied at launch.
+- [RBST0001] - CLI error. Raised when invalid parameters are supplied at launch.
 
-- [SPRKIQ0002] - CLI error. Unimplemented feature error.
+- [RBST0002] - Unimplemented feature error.
 Raised when a JSONiq feature that is not yet implemented in Rumble is used.
 
+- [RBST0003] - Invalid for clause expression error.
+Raised when an RDD enabled expression is located within the body of
```
ghislainfourny: Raised when an expression produces a different, big sequence of items for each binding within a big tuple, which would lead to a data flow explosion and to a nesting of jobs on the Spark cluster.

(see my comments on documentation and error messages that non-computer scientists can understand with no Spark knowledge)

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Sounds great. Thanks for the suggestion

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: I had missed this when moving the code around. Thanks for catching this.

src/main/java/sparksoniq/exceptions/InvalidForClauseExpressionException.java
CanBerker: Done

docs/Error codes.md
CanBerker: Done

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Done

CanBerker:
Thank you for your remarks on the aesthetics of error messages. I feel the final state is much better.

docs/Error codes.md
```diff
@@ -18,11 +18,18 @@ for an object or array selector.
...
-- [SPRKIQ0002] - CLI error. Unimplemented feature error.
+- [RBST0002] - Unimplemented feature error.
 Raised when a JSONiq feature that is not yet implemented in Rumble is used.
 
+- [RBST0003] - Invalid for clause expression error.
+Raised when an expression produces a different,
+big sequence of items for each binding within a big tuple,
+which would lead to a data flow explosion and to a nesting of jobs on the Spark cluster.
+
+which depends on the is used within an RDD enabled FLWOR expression
```
ghislainfourny: this line should probably go away

src/main/resources/test_files/runtime-spark/NestedFLWORSpark4.iq.HOLD
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="1" :)
+for $i in parallelize(1) for $l in (for $i in parallelize(1) return 1)
```
ghislainfourny: This looks like a duplicate of NestedFLWORSpark2.iq  that actually works?

ghislainfourny: This looks like a duplicate of NestedFLWORSpark2.jq that works?

Could we change to

for $l in (for $i in parallelize($i) return 1)

and check that it throws RBST0003? (This error code is untested yet, if I am correct).

CanBerker: If i may point out tiny difference:
NestedFLWORSpark2
for $i in parallelize(1) for $l in (_let_ $i := parallelize(1) return 1)
return $i

NestedFLWORSpark4.iq.HOLD 
for $i in parallelize(1) for $l in (_for_ $i in parallelize(1) return 1)
return $i

The local evaluation on the let clause was what made test2 work. When this is a query with 4, it woud cause nesting and fail with a NullPointerException.

CanBerker: One more thing: the following is not a valid query.
for $l in (for _$i_ in parallelize(_$i_) return 1)
Replacing $i with $$ also doesn't work as I believe there is no prior expression

docs/Error codes.md
CanBerker: Done

src/main/resources/test_files/runtime-spark/NestedFLWORSpark4.iq.HOLD
ghislainfourny: > One more thing: the following is not a valid query.
> for $l in (for _$i_ in parallelize(_$i_) return 1)
> Replacing $i with $$ also doesn't work as I believe there is no prior expression

Correct, that was only a subset. I mean this query:

for $i in parallelize(1)
for $l in (for $i in parallelize($i) return 1)
return $l

should return RBST0003.

ghislainfourny: > If i may point out tiny difference:
> NestedFLWORSpark2
> for $i in parallelize(1) for $l in (_let_ $i := parallelize(1) return 1)
> return $i
> 
> NestedFLWORSpark4.iq.HOLD
> for $i in parallelize(1) for $l in (_for_ $i in parallelize(1) return 1)
> return $i

Yes, I missed this difference. Thank you!

> The local evaluation on the let clause was what made test2 work. When this is a query with for clause, it woud cause nesting and fail with a NullPointerException.

The expression in the second for does not depend on the outer $i: should it not now work, based on what you have just implemented? (i.e., we can simply activate this test, and create the failing one RBST0003 separately) If it does not work, could it be because the variable "hiding" ($i hides $i) that is not working properly?

CanBerker: The test works both with and without variable hiding. You can see it under NestedFLWORSpark4 and NestedFLWORSpark5

CanBerker: I have also added a negative test that triggers RBST0003 in file NestedFLWORSparkError1.jq

CanBerker: Also added the test case you have given above:
for $i in parallelize(1)
for $l in (for $k in parallelize($i) return $k)
return $l

CanBerker:
In the current state, as you mentioned, the test case with the let clause remains disabled:
for $i in parallelize(1) let $l := (for $i in parallelize(1) return 1)
return $i

This will be addressed in a separate PR in the future.

src/main/resources/test_files/runtime-spark/NestedFLWORSparkError1.iq
```diff
@@ -0,0 +1,5 @@
+(:JIQS: ShouldCrash; ErrorCode="RBS0003"; ErrorMetadata="LINE:2:COLUMN:0:" :)
```
ghislainfourny: Here it says RBS0003, but the error is RBST0003 -- it would be nice to double-check that there is no bug in the test machinery, do make sure that this is not a false positive?

ghislainfourny:
Very good, thank you!

PR #408, opened 2019-12-06
Fix count optimization with let group

ghislainfourny:
@CanBerker this is a bug found in class. It would be nice if we can review/merge today and I will do an interim release also with the bugs solved last week. Thanks!

This partially solves https://github.com/RumbleDB/rumble/issues/410 for the let clause (and potentially group clauses, untested). Other clauses can be fixed later, with all needed tests.

Also, this opens the way to the treatment of FLWOR dataframes with other types than binaries in UDFs. I am afraid we have no choice but to pass an array for each type to the UDFs (and will need to extend this to all types), except if we discover that Spark SQL supports polymorphic UDFs in some way...

### DIFF
--- a/src/main/java/sparksoniq/spark/DataFrameUtils.java
+++ b/src/main/java/sparksoniq/spark/DataFrameUtils.java
@@ -72,19 +72,9 @@ import static org.apache.spark.sql.functions.udf;
 
 public class DataFrameUtils {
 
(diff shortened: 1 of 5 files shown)


ghislainfourny:
Thanks a lot, @CanBerker the cleanup looks very good.

PR #411, opened 2019-12-09
Fix context chaining for FLWOR expression evaluation (DF-enabled nested in local)

CanBerker:
This is the continuation of #406 
The predicate looked fine so no improvement is made there.

GroupClauseDetermineTypeUDF.java is removed as it is unused.

- Fixed let clause causing [err: XPDY0130]LINE:4;COLUMN:10;Runtime error retrieving variable $i value

- Fixed order clause causing [err: XPDY0130]LINE:4;COLUMN:10;Runtime error retrieving variable $i value

- Fixed group clause causing [err: XQST0094]LINE:4:COLUMN:9:Variable $i cannot be used in group clause


src/main/java/sparksoniq/spark/udf/GroupClauseCreateColumnsUDF.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.spark.sql.RowFactory;
 import org.apache.spark.sql.api.java.UDF1;
 import org.joda.time.Instant;
+import org.omg.DynamicAny.DynAny;
```
ghislainfourny: This looks interesting, what does this do?

CanBerker: Nothing actually :-) It was merely a typo that somehow survived the import cleanups. Thanks for pointing it out !

CanBerker:
These tests are exactly what I needed. Thank you. I will add them right away

PR #413, opened 2019-12-13
Throw appropriate errors for job within a job for all other clauses

ghislainfourny:
Depends on the nextedFLWOR branch.

Now it cleanly throws an error for all other clauses when an RDD is encountered and the child clause has a DF, because it will anyway always cause a null pointer exception because of a job within a job in the current state. A clean error will be more helpful to users and we can still relax if we detect a few special cases that would still work, such as unused variables.

Later on, we can check for variable dependencies and make more cases work.

I had to change initIsRDD() to always false for dynamic function calls -- this is actually an undecidable problem as we established that a dynamic resolution is not possible, so that we need to consider it false by default (we can talk about this when we discuss the static context isRDD resolution). In the previous state, null pointer exceptions were regularly thrown on dynamic function calls.


ghislainfourny:
@CanBerker you can take a look after merging the other branch (then the diff will be smaller).

PR #414, opened 2019-12-17
Support count optimization in other clauses.

ghislainfourny:
While extending to other clauses, I made it forward compatible with future support of optimized DataFrames with native types.


ghislainfourny:
@CanBerker thank you!

PR #415, opened 2019-12-18
Static is rdd

ghislainfourny:
Depends on #452

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
@@ -20,7 +20,6 @@
 
 package sparksoniq.jsoniq.compiler.translator.expr.flowr;
 
-import sparksoniq.exceptions.OurBadException;
 import sparksoniq.jsoniq.compiler.translator.expr.ExpressionOrClause;
 import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
 import sparksoniq.semantics.visitor.AbstractExpressionOrClauseVisitor;
@@ -58,11 +57,6 @@ public class FlworClause extends ExpressionOrClause {
     @Override
     protected void initHighestExecutionMode() {
         this._highestExecutionMode = previousClause.getHighestExecutionMode();
-    }
-
-    @Override
-    public boolean isRDD() {
-        throw new OurBadException("FLWOR clauses(except return) do not use RDDs. Use either local or DataFrame API");
     }
 
     @Override
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
@@ -55,7 +55,7 @@ public class FlworClause extends ExpressionOrClause {
     }
 
     @Override
-    protected void initHighestExecutionMode() {
+    public void initHighestExecutionMode() {
         this._highestExecutionMode = previousClause.getHighestExecutionMode();
     }
 
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
@@ -55,7 +55,8 @@ public class FlworClause extends ExpressionOrClause {
         this.previousClause = previousClause;
     }
 
-    protected void initIsRDD() {
+    @Override
+    protected void initIsRDDAndIsDataFrame() {
         this.isDataFrame = previousClause.isDataFrame();
     }
 
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
@@ -21,7 +21,6 @@
 package sparksoniq.jsoniq.compiler.translator.expr.flowr;
 
 import sparksoniq.exceptions.OurBadException;
-import sparksoniq.exceptions.SparksoniqRuntimeException;
 import sparksoniq.jsoniq.compiler.translator.expr.ExpressionOrClause;
 import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
 import sparksoniq.semantics.visitor.AbstractExpressionOrClauseVisitor;
@@ -58,6 +57,16 @@ public class FlworClause extends ExpressionOrClause {
 
     protected void initIsRDD() {
         this.isDataFrame = previousClause.isDataFrame();
+    }
+
+    @Override
+    public boolean isRDD() {
+        throw new OurBadException("FLWOR clauses(except return) do not use RDDs. Use either local or DataFrame API");
+    }
+
+    @Override
+    public void setIsRDD(boolean isRDD) {
+        throw new OurBadException("FLWOR clauses(except return) do not use RDDs. Use either local or DataFrame API");
     }
 
     @Override
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
@@ -54,6 +54,9 @@ public class FlworClause extends ExpressionOrClause {
         this.previousClause = previousClause;
     }
 
+    /**
+     * This method is overridden in clauses such as for and let for special behavior
+     */
     @Override
     public void initHighestExecutionMode() {
         this._highestExecutionMode = previousClause.getHighestExecutionMode();
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
@@ -56,17 +56,12 @@ public class FlworClause extends ExpressionOrClause {
     }
 
     @Override
-    protected void initIsRDDAndIsDataFrame() {
-        this.isDataFrame = previousClause.isDataFrame();
+    protected void initHighestExecutionMode() {
+        this._highestExecutionMode = previousClause.getHighestExecutionMode();
     }
 
     @Override
     public boolean isRDD() {
-        throw new OurBadException("FLWOR clauses(except return) do not use RDDs. Use either local or DataFrame API");
-    }
-
-    @Override
-    public void setIsRDD(boolean isRDD) {
         throw new OurBadException("FLWOR clauses(except return) do not use RDDs. Use either local or DataFrame API");
     }
 
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
@@ -56,17 +56,6 @@ public class FlworClause extends ExpressionOrClause {
         this.previousClause = previousClause;
     }
 
-    @Override
-    public boolean isRDD() {
-        throw new OurBadException("FLWOR clauses do not use RDDs. Use either local or DataFrame API");
-    }
-
-    @Override
-    public void setIsRDD(boolean isRDD) {
-        throw new OurBadException("FLWOR clauses do not use RDDs. Use either local or DataFrame API");
-    }
-
-    @Override
     protected void initIsRDD() {
         this.isDataFrame = previousClause.isDataFrame();
     }
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
@@ -21,6 +21,7 @@
 package sparksoniq.jsoniq.compiler.translator.expr.flowr;
 
 import sparksoniq.exceptions.SemanticException;
+import sparksoniq.exceptions.SparksoniqRuntimeException;
 import sparksoniq.jsoniq.compiler.translator.expr.Expression;
 import sparksoniq.jsoniq.compiler.translator.expr.ExpressionOrClause;
 import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
@@ -76,9 +77,23 @@ public class FlworExpression extends Expression {
     }
 
     @Override
-    protected void initIsRDD() {
-        this.isRDD = _returnClause.isDataFrame();
-        this.isDataFrame = false;
+    public boolean isRDD() {
+        return this._returnClause.isRDD();
+    }
+
+    @Override
+    public void setIsRDD(boolean isRDD) {
+        throw new SparksoniqRuntimeException("The FLWOR Expression class does not implement this", getMetadata());
+    }
+
+    @Override
+    public boolean isDataFrame() {
+        return this._returnClause.isDataFrame();
+    }
+
+    @Override
+    public void setIsDataFrame(boolean isRDD) {
+        throw new SparksoniqRuntimeException("The FLWOR Expression class does not implement this", getMetadata());
     }
 
     @Override
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
@@ -21,7 +21,6 @@
 package sparksoniq.jsoniq.compiler.translator.expr.flowr;
 
 import sparksoniq.exceptions.SemanticException;
-import sparksoniq.exceptions.SparksoniqRuntimeException;
 import sparksoniq.jsoniq.compiler.translator.expr.Expression;
 import sparksoniq.jsoniq.compiler.translator.expr.ExpressionOrClause;
 import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
@@ -77,6 +77,12 @@ public class FlworExpression extends Expression {
     }
 
     @Override
+    public void initHighestExecutionMode() {
+        // overall flwor expression's execution mode is never used and remains unset
+        this._highestExecutionMode = ExecutionMode.UNSET;
+    }
+
+    @Override
     public ExecutionMode getHighestExecutionMode() {
         // overall flwor expression's execution mode is stored in the return clause
         return this._returnClause.getHighestExecutionMode();
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
@@ -80,22 +80,10 @@ public class FlworExpression extends Expression {
         return this._returnClause.isRDD();
     }
 
-    @Override
-    public void setIsRDD(boolean isRDD) {
-        throw new SparksoniqRuntimeException("The FLWOR Expression class does not implement this", getMetadata());
-    }
-
-    @Override
     public boolean isDataFrame() {
         return this._returnClause.isDataFrame();
     }
 
-    @Override
-    public void setIsDataFrame(boolean isRDD) {
-        throw new SparksoniqRuntimeException("The FLWOR Expression class does not implement this", getMetadata());
-    }
-
-    @Override
     public List<ExpressionOrClause> getDescendants(boolean depthSearch) {
         List<ExpressionOrClause> result = new ArrayList<>();
         result.add(_startClause);
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
@@ -21,6 +21,7 @@
 package sparksoniq.jsoniq.compiler.translator.expr.flowr;
 
 import sparksoniq.exceptions.SemanticException;
+import sparksoniq.jsoniq.ExecutionMode;
 import sparksoniq.jsoniq.compiler.translator.expr.Expression;
 import sparksoniq.jsoniq.compiler.translator.expr.ExpressionOrClause;
 import sparksoniq.jsoniq.compiler.translator.metadata.ExpressionMetadata;
@@ -76,12 +77,9 @@ public class FlworExpression extends Expression {
     }
 
     @Override
-    public boolean isRDD() {
-        return this._returnClause.isRDD();
-    }
-
-    public boolean isDataFrame() {
-        return this._returnClause.isDataFrame();
+    public ExecutionMode getHighestExecutionMode() {
+        // overall flwor expression's execution mode is stored in the return clause
+        return this._returnClause.getHighestExecutionMode();
     }
 
     public List<ExpressionOrClause> getDescendants(boolean depthSearch) {
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
@@ -59,10 +59,10 @@ public class ForClause extends FlworClause {
 
     @Override
     protected void initHighestExecutionMode() {
-        // call isDF on the last forVariable
+        // call isDataFrame on the last forVariable
         this._highestExecutionMode =
-            forVariables.get(forVariables.size() - 1).getHighestExecutionMode().isDF()
-                ? ExecutionMode.DF
+            forVariables.get(forVariables.size() - 1).getHighestExecutionMode().isDataFrame()
+                ? ExecutionMode.DATAFRAME
                 : ExecutionMode.LOCAL;
     }
 
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
@@ -57,7 +57,7 @@ public class ForClause extends FlworClause {
     }
 
     @Override
-    protected void initIsRDD() {
+    protected void initIsRDDAndIsDataFrame() {
         // call isDF on the last forVariable
         this.isDataFrame = forVariables.get(forVariables.size() - 1).isDataFrame();
     }
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
@@ -59,11 +59,11 @@ public class ForClause extends FlworClause {
     @Override
     protected void initIsRDD() {
         // chain forVariables with previousClause relationship
-        for (int varIndex = forVariables.size()-1; varIndex > 0; varIndex--) {
-            forVariables.get(varIndex).setPreviousClause(forVariables.get(varIndex-1));
+        for (int varIndex = forVariables.size() - 1; varIndex > 0; varIndex--) {
+            forVariables.get(varIndex).setPreviousClause(forVariables.get(varIndex - 1));
         }
         // call isDF on the last forVariable
-        this.isDataFrame = forVariables.get(forVariables.size()-1).isDataFrame();
+        this.isDataFrame = forVariables.get(forVariables.size() - 1).isDataFrame();
     }
 
     @Override
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
@@ -58,10 +58,6 @@ public class ForClause extends FlworClause {
 
     @Override
     protected void initIsRDD() {
-        // chain forVariables with previousClause relationship
-        for (int varIndex = forVariables.size() - 1; varIndex > 0; varIndex--) {
-            forVariables.get(varIndex).setPreviousClause(forVariables.get(varIndex - 1));
-        }
         // call isDF on the last forVariable
         this.isDataFrame = forVariables.get(forVariables.size() - 1).isDataFrame();
     }
--- a/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
+++ b/src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
@@ -61,7 +61,7 @@ public class ForClause extends FlworClause {
     protected void initHighestExecutionMode() {
         // call isDF on the last forVariable
         this._highestExecutionMode =
-            forVariables.get(forVariables.size() - 1).isDataFrame()
+            forVariables.get(forVariables.size() - 1).getHighestExecutionMode().isDF()
                 ? ExecutionMode.DF
                 : ExecutionMode.LOCAL;
     }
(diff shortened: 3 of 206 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/BuiltinFunction.java
```diff
@@ -6,16 +6,16 @@
 
     private FunctionIdentifier identifier;
     private final FunctionSignature signature;
-    private final Class<? extends RuntimeIterator> functionIterator;
+    private final Class<? extends RuntimeIterator> functionIteratorClass;
```
CanBerker: Simple rename for clarity

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ArrayRuntimeIterator.java
```diff
@@ -46,7 +45,7 @@ public ArrayRuntimeIterator(RuntimeIterator arrayItems, IteratorMetadata iterato
     public Item next() {
         if (this._hasNext) {
             List<Item> result = this.runChildrenIterators(this._currentDynamicContext);
-            this._item = ItemFactory.getInstance().createArrayItem(result);
+            Item _item = ItemFactory.getInstance().createArrayItem(result);
```
CanBerker: Variable made local

src/main/java/sparksoniq/jsoniq/runtime/iterator/RuntimeIterator.java
```diff
@@ -191,13 +198,37 @@ public IteratorMetadata getMetadata() {
         return metadata;
     }
 
-    public abstract boolean isRDD();
+    public boolean isRDD() {
+        if (!isRDDIsSet) {
+            throw new RuntimeException("isRDD field accessed in iterator without being set.");
+        }
+        return isRDD;
```
CanBerker: These exceptions are put in place as safeguards to prevent unnoticed bugs. They can be potentially be removed after the implementation is checked to be sound.

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/FunctionDeclaration.java
```diff
@@ -74,6 +76,20 @@ public Expression get_body() {
...
+        this.isDataFrame = false;
+
+        FunctionIdentifier identifier = new FunctionIdentifier(this._name, this._params.size());
+        // if named(static) function declaration
+        if (!this._name.equals("")) {
+            // TODO: find a way to set isRDD and isDF from this._body
+            Functions.addUserDefinedFunctionIsRDD(identifier, false, this.getMetadata());
+            Functions.addUserDefinedFunctionIsDataFrame(identifier, false, this.getMetadata());
+        }
+    }
```
CanBerker: Note: This is still not addressed

src/main/java/sparksoniq/jsoniq/runtime/iterator/DataFrameRuntimeIterator.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+
+package sparksoniq.jsoniq.runtime.iterator;
+
+import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
+
+import java.util.List;
+
+public abstract class DataFrameRuntimeIterator extends RDDRuntimeIterator {
```
CanBerker: Class defined for DataFrame only Iterators. eg. Parquet file and Structured Json File.

In the old implementation this would have a method isDataFrame which would always return true.

New implementation uses this class while checking if a function's iterator is a subclass of this.

src/main/resources/test_files/runtime-spark/Pushdown/KeysDF1.jq
```diff
@@ -1,2 +1,2 @@
-(:JIQS: ShouldRun; Output="(choices, country, date, guess, sample, target)" :)
+(:JIQS: ShouldRun; Output="(country, choices, guess, target, date, sample)" :)
 keys(structured-json-file("./src/main/resources/queries/conf-ex.json"))
```
CanBerker: This test is executed locally on the master which I believe is not correct. It is now executed with RDD API on the keys iterator which gives the new results.

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
```diff
@@ -38,12 +38,28 @@ public ForClause(List<ForClauseVar> vars, ExpressionMetadata metadataFromContext
...
         return forVariables;
     }
 
+    @Override
+    public void setPreviousClause(FlworClause previousClause) {
+        super.setPreviousClause(previousClause);
+    }
+
+    @Override
+    protected void initIsRDD() {
```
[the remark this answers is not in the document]
CanBerker: Done

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
```diff
@@ -38,12 +38,28 @@ public ForClause(List<ForClauseVar> vars, ExpressionMetadata metadataFromContext
...
+        for (int varIndex = forVariables.size() - 1; varIndex > 0; varIndex--) {
+            forVariables.get(varIndex).setPreviousClause(forVariables.get(varIndex - 1));
+        }
     }
 
     public List<ForClauseVar> getForVariables() {
         return forVariables;
     }
 
+    @Override
```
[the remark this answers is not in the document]
CanBerker: I believe this was an oversight. I have updated the behavior similar to the letClause as both of these clauses can have a multitude of variable definitions, which must be chained together for correctness.

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/VariableReference.java
```diff
@@ -38,6 +38,12 @@
     public VariableReference(String _name, ExpressionMetadata metadata) {
         super(metadata);
         this.name = _name;
+        // TODO: can we eliminate this assignment by properly visiting VariableReferences in staticContextVisitor
+        _highestExecutionMode = ExecutionMode.LOCAL;
```
CanBerker: Removing this line causes a multitude of runtime exceptions as VariableReferenceIterators(created from variableReferences that inherit from Expression) have their execution mode unset. This is caused by them not being properly visited by the staticContextVisitor which sets the execution mode by looking up the variable in the context. I think this will require a bit of work on the staticContextVisitor if we wish to address this.

src/main/java/sparksoniq/semantics/visitor/StaticContextVisitor.java
```diff
@@ -48,76 +49,109 @@
...
+        // define a static context for the function body, add params to the context and visit the body expression
+        StaticContext functionDeclarationContext = new StaticContext(argument);
+        expression.get_params()
+            .forEach(
+                (paramName, flworVarSequenceType) -> functionDeclarationContext.addVariable(
+                    paramName,
+                    flworVarSequenceType.getSequence(),
+                    expression.getMetadata(),
+                    ExecutionMode.LOCAL // static udf currently supports materialized(local) params, not RDDs or DFs
+                )
```
CanBerker: At the moment, static function declarations allow parametrized variables to be in the materialized form. In other words, (and if I'm not mistaken), user defined functions would not work properly when a variable that is a DF or an RDD is passed as a parameter.

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -54,30 +54,25 @@
 
     private static final long serialVersionUID = 1L;
     private String _variableName; // for efficient use in local iteration
-    private RuntimeIterator _expression;
+    private RuntimeIterator _assignmentIterator;
```
CanBerker: Simple renaming for the iterators used in for and let clause iterators

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
```diff
@@ -54,30 +54,25 @@
...
     private Map<String, DynamicContext.VariableDependency> _dependencies;
     private DynamicContext _tupleContext; // re-use same DynamicContext object for efficiency
     private FlworTuple _nextLocalTupleResult;
     private FlworTuple _inputTuple; // tuple received from child, used for tuple creation
 
     public ForClauseSparkIterator(
             RuntimeTupleIterator child,
-            VariableReferenceIterator variableReference,
-            RuntimeIterator assignmentExpression,
+            String variableName,
```
CanBerker: Building an iterator just to get the name of the variable was redundant for 'for' and 'let' clauses. We simply pass in the variable name now

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
```diff
@@ -44,19 +45,12 @@
     private List<Item> _items = null;
     private int _currentIndex = 0;
 
-    public VariableReferenceIterator(String variableName, SequenceType seq, IteratorMetadata iteratorMetadata) {
-        super(null, iteratorMetadata);
-        _variableName = "$" + variableName;
+    public VariableReferenceIterator(String variableName, SequenceType seq, ExecutionMode executionMode, IteratorMetadata iteratorMetadata) {
+        super(null, executionMode, iteratorMetadata);
+        _variableName = variableName;
```
CanBerker: In order to be consistent across the application, I have standardized the variable names in wherever they are used. Variables are now stored and referenced in the dynamic context without the $ prefix. They are also mentioned in error messages without the prefix. I have opted for this as the behavior now matches Zorba.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
```diff
@@ -522,44 +524,66 @@ public static boolean checkBuiltInFunctionExists(FunctionIdentifier identifier)
...
+                                IteratorMetadata.class
+                        );
+                functionCallIterator = constructor.newInstance(arguments, executionMode, metadata);
+            } catch (ReflectiveOperationException ex) {
+                throw new UnknownFunctionCallException(
+                        identifier.getName(),
+                        arguments.size(),
+                        metadata.getExpressionMetadata()
+                );
+            }
```
CanBerker: I'd like to have your feedback on this one @ghislainfourny. To minimize the refactoring, I have initially opted for adding the execution mode as a parameter to the HybridRuntimeIterator and its descendants. LocalRuntimeIterators simply get this parameter fixed to LOCAL. This resulted in hybrid enabled built-in functions having constructors with 3 parameters while Local built-in functions have constructors with 2 parameters. In order to handle this and get the tests running. I have implemented the above fix which is quick and dirty.

One potential solution that would still be somewhat dirty but more performant. We could eliminate the exceptions by performing a certain check on the constructor to avoid performance loss due to exception handling.

Alternative feels more neat however will require a bit more refactoring. We can simply add ExecutionMode to all the constructors of all iterators and enforce that localRuntimeIterators can only be assigned ExecutionMode.LOCAL. A similar implementation can be found in DataFrameRuntimeIterator or RDDRuntimeIterator. 

What is your take on this? Do you see any preferable alternatives/suggestions?

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/FunctionDeclaration.java
CanBerker: Outdated

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Outdated

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
```diff
@@ -50,8 +54,9 @@ public void setPreviousClause(FlworClause previousClause) {
         this.previousClause = previousClause;
     }
 
-    public FLWOR_CLAUSES getClauseType() {
-        return clauseType;
+    @Override
+    public void initHighestExecutionMode() {
+        this._highestExecutionMode = previousClause.getHighestExecutionMode();
```
ghislainfourny: Could we add a comment that this is meant to be overridden in some cases? Thanks!

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
```diff
@@ -77,6 +77,12 @@ private void set_returnClause(ReturnClause returnClause) {
     }
 
     @Override
+    public ExecutionMode getHighestExecutionMode() {
+        this._highestExecutionMode = ExecutionMode.UNSET;
```
ghislainfourny: A side effect in a get function raises red flags. Is the default not unset anyway (because of init), and do we not just ignore it?

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
```diff
@@ -56,6 +57,15 @@ public void setPreviousClause(FlworClause previousClause) {
         forVariables.get(0).previousClause = this.previousClause;
     }
 
+    @Override
+    public void initHighestExecutionMode() {
+        // call isDataFrame on the last forVariable
+        this._highestExecutionMode =
+            forVariables.get(forVariables.size() - 1).getHighestExecutionMode().isDataFrame()
+                ? ExecutionMode.DATAFRAME
+                : ExecutionMode.LOCAL;
```
ghislainfourny: Is there a reason why not just

this._highestExecutionMode = forVariables.get(forVariables.size() - 1).getHighestExecutionMode()

?

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/LetClause.java
```diff
@@ -55,6 +56,15 @@ public void setPreviousClause(FlworClause previousClause) {
         letVars.get(0).previousClause = this.previousClause;
     }
 
+    @Override
+    public void initHighestExecutionMode() {
+        // call isDataFrame on the last letVariable
+        this._highestExecutionMode =
+            letVars.get(letVars.size() - 1).getHighestExecutionMode().isDataFrame()
+                ? ExecutionMode.DATAFRAME
```
ghislainfourny: The same applies here: why not

this._highestExecutionMode = letVars.get(letVars.size() - 1).getHighestExecutionMode()

?

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/OrderByClauseExpr.java
```diff
@@ -48,6 +49,12 @@ public OrderByClauseExpr(
         this._emptyOrder = empty_order;
     }
 
+    @Override
+    public void initHighestExecutionMode() {
+        // OrderByClauseExpr's execution mode is not used. Leave it unset
+        this._highestExecutionMode = ExecutionMode.UNSET;
```
ghislainfourny: Why is it not used, i.e., why is this clause treated differently?

src/main/java/sparksoniq/jsoniq/tuple/FlworTuple.java
```diff
@@ -104,7 +102,7 @@ public boolean isRDD(String key, IteratorMetadata metadata) {
             || dfVariables.containsKey(key);
     }
 
-    public boolean isDF(String key, IteratorMetadata metadata) {
+    public boolean isDataFrame(String key, IteratorMetadata metadata) {
```
ghislainfourny: This can be moved to a separate PR, reviewed and merged.

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/postfix/extensions/DynamicFunctionCallExtension.java
```diff
@@ -48,6 +49,11 @@ public DynamicFunctionCallExtension(List<Expression> arguments, ExpressionMetada
         return getDescendantsFromChildren(result, depthSearch);
     }
 
+    @Override
+    public void initHighestExecutionMode() {
+        this._highestExecutionMode = ExecutionMode.LOCAL;
```
ghislainfourny: Is it not the default already, i.e., is this not superfluous?

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
CanBerker: You're absolutely right. That was simply an oversight on my part. That line was meant for the init method.

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
CanBerker: No specific reason, just an oversight :) Good catch!

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/LetClause.java
CanBerker: Also fixed here

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/postfix/extensions/DynamicFunctionCallExtension.java
CanBerker: Postfix extension (parent class of DynamicFunctionCallExtension) overrides init method to the behavior which is shared by all postfix extensions except dynamicFunctionCallExtension. This is always local, which is the behavior identical to the grandparent.

src/main/java/sparksoniq/jsoniq/tuple/FlworTuple.java
CanBerker: Done. Please review at #446

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
CanBerker: Added

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/postfix/extensions/DynamicFunctionCallExtension.java
CanBerker: I have added a comment to the code as well

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/OrderByClauseExpr.java
CanBerker: This again due to the current nature of how we are creating and using our iterators. These orderbyClauseExpressions are the individual expressions within an order by expression (ie. order by $x asc, $y dec.... ). The execution mode of such individual expressions are not used in the current state of the program. 

Overall orderByClause gets its execution mode set based on the previousClause and this expression's execution mode sets the execution mode of the OrderByClauseSparkIterator. 

The orderByClauseExpr (the subexpressions) contain the actual expression along with the ascending/descending and empty handling information. This actual expression is what the iterator is generated from and it sets the execution mode directly for the iterator. As orderByClauseExpr does not partake in generation and execution mode of an iterator, its execution mode is unused in the current implementation.

I generally opt for setting these UNSET values explicitly in the initialization for showing the deliberate nature of this value, especially for the case if the implementation changes in the future and orderByClauseExpr plays a more direct role for a RuntimeIterator.

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/VariableReferenceIterator.java
CanBerker: Outdated

src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
```diff
@@ -40,26 +40,25 @@
...
+        return false;
     }
 
-    @Override
-    public boolean isDataFrame() {
-        return false;
+    protected void fallbackToRDDIfDFNotImplemented(ExecutionMode executionMode) {
+        if (executionMode == ExecutionMode.DATAFRAME && !this.implementsDataFrames()) {
+            this._highestExecutionMode = ExecutionMode.RDD;
+        }
```
CanBerker: I will investigate again why this function was needed in the first place, it may be masking some other correctness issue

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworClause.java
ghislainfourny: Thanks!

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/FlworExpression.java
ghislainfourny: Thanks!

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/FunctionCall.java
```diff
@@ -52,11 +62,64 @@ public String getFunctionName() {
...
     }
 
+    @Override
+    public final void initHighestExecutionMode() {
+        throw new OurBadException("Function call expressions do not use the highestExecutionMode initializer");
+    }
+
+    public void initFunctionCallHighestExecutionMode(boolean ignoreMissingFunctionError) {
+        FunctionIdentifier identifier = new FunctionIdentifier(this._functionName, this._arguments.size());
+        if (Functions.checkBuiltInFunctionExists(identifier)) {
```
ghislainfourny: My intuition tells me that this code could be simplified by adding a return statement after each _highestExecutionMode assignments. Then all elses would go away and the structure would become flatter and easier to read.

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
```diff
@@ -0,0 +1,19 @@
...
+import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
+import sparksoniq.semantics.StaticContext;
+
+public class VisitorHelpers {
+
+    public static RuntimeIterator generateRuntimeIterator(ExpressionOrClause expression, RuntimeIterator argument) {
+        return new RuntimeIteratorVisitor().visit(expression, argument);
+    }
+
+    public static StaticContext generateStaticContextDoublePass(ExpressionOrClause expression, StaticContext argument) {
```
ghislainfourny: public static void populateStaticContext(ExpressionOrClause expression)

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
```diff
@@ -0,0 +1,19 @@
+package sparksoniq.semantics.visitor;
+
+import sparksoniq.jsoniq.compiler.translator.expr.ExpressionOrClause;
+import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
+import sparksoniq.semantics.StaticContext;
+
+public class VisitorHelpers {
+
+    public static RuntimeIterator generateRuntimeIterator(ExpressionOrClause expression, RuntimeIterator argument) {
```
ghislainfourny: public static RuntimeIterator generateRuntimeIterator(ExpressionOrClause expression)

 (argument is null always)

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
```diff
@@ -0,0 +1,19 @@
...
+        return new RuntimeIteratorVisitor().visit(expression, argument);
+    }
+
+    public static StaticContext generateStaticContextDoublePass(ExpressionOrClause expression, StaticContext argument) {
+        StaticContextVisitor visitor = new StaticContextVisitor();
+        visitor.visit(expression, argument);
```
ghislainfourny: argument is null always

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
```diff
@@ -0,0 +1,19 @@
...
+        visitor.setConfigForConsequentPasses();
+        return visitor.visit(expression, argument);
```
ghislainfourny: argument is null always

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/FunctionCall.java
ghislainfourny: Offline discussion: function catalogue maps each builtin function to execution mode, and here we only look it up.

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
CanBerker: Done

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
CanBerker: Done

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
CanBerker: Done

src/main/java/sparksoniq/semantics/visitor/VisitorHelpers.java
CanBerker: Done

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
CanBerker: TODO - Offline discussion: expose executionMode parameters from all iterator constructors, remove the exception handling Above

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/FunctionCall.java
CanBerker: I agree with the simplification using returns and I will implement this.

Regarding the second point: for the case of LocalRuntimeIterator, RDDRuntimeIterators or DataFrameRuntimeIterators where execution mode is always a fixed value, the catalogue approach would work. However in the case of HybridRuntimeIterators we need some custom logic in place that determines the execution mode based on the the child expression. But even this logic is not shared by all hybrid iterators due to peculiarities of individual iterators such as object 'keys', where we resort to local execution also when child is a DataFrame. 

Given this information that not all built-in functions have a static executionMode that can be accessed with a simple lookup, how would you like to proceed?

CanBerker: The return oriented implementation is in place. I have also attempted to tidy up the code a bit to increase readability. 

I am looking forward to your feedback for further improvements or alternative solutions to the mentioned issue.

ghislainfourny: Many thanks for the clean return-based implementation.

I am optimistic on the hybrid function iterators, especially looking at the cleaned code below, because there are not so many cases and we can come up with an larger list of function-specific execution modes (local, RDD, DF, but also same-as-first-parameter, etc). Then we just need to map this larger list (stored and looked up in the catalogue) to our "regular" execution mode list looking at the parameter iterators.

Could you please prepare a list of all functions that are based on a hybrid iterator? Then we'll take a closer look at them.

Thanks!

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
```diff
@@ -123,6 +125,7 @@ private void setFunctionItemAndIteratorWithCurrentContext() {
         }
         _functionCallIterator = Functions.buildUserDefinedFunctionCallIterator(
             _functionItem,
+            _functionItem.getBodyIterator().getHighestExecutionMode(),
```
ghislainfourny: For dynamic call, I think this is always local no matter what.

(only static UDF calls are able to execute the function body as an RDD).

ghislainfourny:
@CanBerker this is going in the right direction. I only have the two visitors as well as the function call mechanism left to review and we can fine-tune the mechanism there (see comment).

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/base/Functions.java
CanBerker: Done

src/main/java/sparksoniq/spark/iterator/flowr/ForClauseSparkIterator.java
CanBerker: Outdated

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/VariableReference.java
CanBerker: This is addressed

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/FunctionCall.java
CanBerker: These are the Hybrid function iterators:
- keys
- values
- distinct-values

- json-file
- parallelize
- text-file

Do you wish to perform this refactoring within this PR or shall we address introduction of function executionMode catalogue and functionIterator-only execution modes in a separate PR?

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/DynamicFunctionCallIterator.java
CanBerker: DynamicFunctionCallRuntimeIterator always evaluates locally indeed. However if a UDF is RDD/DF enabled, I think it could be evaluated in that manner. In this scope of this PR all UDFs are locally evaluated indeed. But #451 will be bringing the support for RDD/DF enabled UDFs.
For the example below,  statically detected postFixExtension of dynamicFunctionCall will be locally executed. However the function invoked by this dynamicFunctionCall could be evaluated with RDDs. Please correct me if there is something that I am missing.

declare function test() as integer{
    parallelize((1,2,3))
};
test#0()

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/primary/FunctionCall.java
ghislainfourny: Is it not just keys, values and distinct-values?

Among these: json-file, parallelize and text-file are anyway always RDDs (RDDRuntimeIterator), correct?

So if it is just keys, values and distinct-values that are "truly" hybrid and could be either local or RDD, then it should be quite easy as we only need one more case: when the highestExecutionMode is taken from the first parameter's highestExecutionMode. All three functions work that way.

This can be done with a separate, function-specific enum.

CanBerker: Yes absolutely, that was a mistake. It's only the three that you have mentioned.

The cases are accurate. I will implement these next.

src/main/java/sparksoniq/semantics/visitor/StaticContextVisitor.java
ghislainfourny: I agree with always passing materialized parameters to UDFs. This makes the higherExecutionMode of the expressions in the function body not depend on the parameters, and "truly static": indeed, if the same function is called at multiple places, there is no guarantee the passed parameters will always be local or always be RDDs.

If we relax this later, this will require significantly more work to get right; now would be too premature.

ghislainfourny:
@CanBerker I have now looked at the visitors and it looks fine too.

Thus, to the best of my knowledge, the builtin function "extended execution mode map" (if we call it that way) is the last issue to handle :)

src/main/java/sparksoniq/jsoniq/runtime/iterator/HybridRuntimeIterator.java
ghislainfourny: I think that's fine for now -- in the future, it would be desirable to have a perfect match between what the static analysis says and what actually happens at runtime, but we can fine-tune later.

ghislainfourny:
Thanks @CanBerker also for the URL validation. That's good to go! (I can't approve the PR as I created it).

PR #417, opened 2020-01-06
Fix bug and trigger clear message asking for bug-reports when it's our fault

ghislainfourny:
Fixes https://github.com/RumbleDB/rumble/issues/418

Also:
- adds undocumented parameter --error-info to display stack trace when other issues like this happen in the future.
- adds OurBadException to clearly distinguish those cases that are our fault and must trigger a clear message asking for a bug report. Many runtime exceptions were incorrectly displaying an error message as if it were the user's fault (with a clean dynamic error code).

### DIFF
--- a/src/main/java/org/rumbledb/cli/Main.java
+++ b/src/main/java/org/rumbledb/cli/Main.java
@@ -77,18 +77,18 @@ public class Main {
                 );
             }
         } catch (Exception ex) {
-            handleException(ex, sparksoniqConf.getErrorInfo());
+            handleException(ex, sparksoniqConf.getShowErrorInfo());
         }
     }
 
-    private static void handleException(Throwable ex, boolean errorInfo) {
+    private static void handleException(Throwable ex, boolean showErrorInfo) {
         if (ex != null) {
             if (ex instanceof SparkException) {
                 Throwable sparkExceptionCause = ex.getCause();
-                handleException(sparkExceptionCause, errorInfo);
+                handleException(sparkExceptionCause, showErrorInfo);
             } else if (ex instanceof SparksoniqRuntimeException && !(ex instanceof OurBadException)) {
                 System.err.println("⚠️  ️" + ex.getMessage());
-                if (errorInfo) {
+                if (showErrorInfo) {
                     ex.printStackTrace();
                 }
             } else {
@@ -97,7 +97,7 @@ public class Main {
                     "We should investigate this 🙈. Please contact us or file an issue on GitHub with your query."
                 );
                 System.out.println("Link: https://github.com/RumbleDB/rumble/issues");
-                if (errorInfo) {
+                if (showErrorInfo) {
                     ex.printStackTrace();
                 }
             }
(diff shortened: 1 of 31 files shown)


src/main/java/org/rumbledb/config/SparksoniqRuntimeConfiguration.java
```diff
@@ -62,6 +62,13 @@ public boolean getOverwrite() {
             return false;
     }
 
+    public boolean getErrorInfo() {
+        if (this._arguments.containsKey("error-info"))
+            return this._arguments.get("error-info").equals("yes");
+        else
+            return false;
+    }
```
CanBerker: How would you feel about the name 'show-error-info' for flag/configuration. The getter method naming is a bit cryptic as it is. The current use case of the configuration and its naming in Main.java (I will also tag this), felt misleading to me

src/main/java/org/rumbledb/cli/Main.java
```diff
@@ -74,24 +77,29 @@ public static void main(String[] args) throws IOException {
                 );
             }
         } catch (Exception ex) {
-            handleException(ex);
+            handleException(ex, sparksoniqConf.getErrorInfo());
         }
     }
 
-    private static void handleException(Throwable ex) {
+    private static void handleException(Throwable ex, boolean errorInfo) {
```
CanBerker: As discussed in SparksoniqRuntimeConfiguration,  in my opinion, 'showErrorInfo' would serve as a much better boolean name. I would expect a method call such as handleException(...., errorInfo) to pass in the actual information, rather than the flag indicating whether info should be displayed or not.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/general/SubsequenceFunctionIterator.java
```diff
@@ -179,7 +180,7 @@ public void setNextResult() {
                 } else if (_length == -1) { // _length not specified -> take all items until the end
                     _nextResult = _sequenceIterator.next();
                 } else {
-                    throw new SparksoniqRuntimeException(
+                    throw new OurBadException(
                             "Unexpected length value found. Please report the bug with subsequence function iterator."
```
CanBerker: I feel that this error message duplicates the 'please report a bug' sentiment. Shall we change it to something along the lines of 'Unexpected length value found in subsequence iterator'?

src/main/java/org/rumbledb/config/SparksoniqRuntimeConfiguration.java
ghislainfourny: Yes, I agree. This is done.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/sequences/general/SubsequenceFunctionIterator.java
ghislainfourny: I agree too.

src/main/java/org/rumbledb/cli/Main.java
ghislainfourny: Also agreed

ghislainfourny:
Thanks a lot @CanBerker I implemented all your feedback. I also copied over the changed handleException code to the Shell's handleException for the same behavior.

Issue #421, opened 2020-01-07
Error when then clause of a conditional express returns the empty sequence


ghislainfourny:
if(true)
then ()
else ()

PR #422, opened 2020-01-08
Empty sequence if

ghislainfourny:
Solves #421 

The if expression iterator was incorrectly materializing the data. Now it is implemented in a streaming fashion that also treats empty sequences correctly.


ghislainfourny:
Thanks @CanBerker

PR #425, opened 2020-01-17
Cherry pick 2.

ghislainfourny:
Another cherry pick


CanBerker:
This renaming was directly related to cleaning up of redundant _currentDynamicContext usage in getRDD and getDF methods. (probably in the next or previous commit) I think it would be the best if we cherry picked those related commits together

ghislainfourny:
> This renaming was directly related to cleaning up of redundant _currentDynamicContext usage in getRDD and getDF methods. (probably in the next or previous commit) I think it would be the best if we cherry picked those related commits together

Yes, this makes sense. Could you please do so? (only the related commits, the goal is to only have what is strictly related). Thanks!

CanBerker:
I will address this in a new branch & PR

PR #430, opened 2020-01-18
4-Cleanup function item call iterator

CanBerker:


ghislainfourny:
> _No description provided._

@CanBerker could you explain this one to me? Thanks!

CanBerker:
FunctionItemCallIterator has 2 execution paths: either a partial application or a regular function call.
These two execution paths were extracted into private methods to add clarity to the execution. Additionally, reading and verification of the arguments that are passed in is also extracted to a function. This helper method is called within the bodies of the two earlier methods.

RDD execution does not make use of class field _currentDynamicContextForLocalExecution. To this end, createDynamicContextWithArguments takes a dynamic context as parameter to be used in execution. 

PS: I have noticed that generatePartiallyAppliedFunction should also support dynamic context passed in as parameter. This is implemented later in #437, however I can try to add a temporary implementation to this PR for overall correctness.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
```diff
@@ -113,64 +119,75 @@ private void processArguments() {
...
-            // place argument values into dynamic context
-            _currentDynamicContext = new DynamicContext(_currentDynamicContext);
-            for (Map.Entry<String, List<Item>> argumentEntry : argumentValues.entrySet()) {
-                _currentDynamicContext.addVariableValue(
-                    argumentEntry.getKey(),
-                    argumentEntry.getValue()
-                );
-            }
-            _functionBodyIterator = _functionItem.getBodyIterator();
+            List<Item> argValue = getItemsFromIteratorWithCurrentContext(argIterator);
```
ghislainfourny: shouldn't context (the parameter) be used as well here?

CanBerker: I was investigating this as well, as it seemed like a potential bug that required testing. However, I just noticed that if the function is partially executed, the execution mode is set to Local. Hence no problems came up previously. I will try to clarify this in the code by adding throwing an exception if execution reaches RDD evaluation during a partial application.

CanBerker: Could you take a look at the latest commit now

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
```diff
@@ -225,8 +243,15 @@ public void setNextResult() {
 
     @Override
     public JavaRDD<Item> getRDDAux(DynamicContext dynamicContext) {
-        processArguments();
-        return _functionBodyIterator.getRDD(_currentDynamicContext);
+        if (_isPartialApplication) {
+            throw new OurBadException(
```
ghislainfourny: Yes, exactly -- in the static analysis, you can simply make sure that all partial function evaluations are locally evaluated.

This makes sense, since to the best of my understand, only one item is then returned anyway: using RDDs would be an unnecessary overhead.

src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
ghislainfourny: I did and it looks good. I still have a concern with the line above: 

List<Item> argValue = getItemsFromIteratorWithCurrentContext(argIterator);

Indeed, this method createNewDynamicContextWithArguments is also called in getRDDAux, so that it might make more sense to past the dynamic context supplied as a parameter (context).

A query like this might be a way to test:

declare function f($i) { parallelize($i) }; let $a := 3 return f($a)

ghislainfourny:
Comment addressed in subsequent PR, so approve.

PR #432, opened 2020-01-18
6-Add previous clause to flwor expressions

CanBerker:


src/main/java/sparksoniq/jsoniq/compiler/translator/expr/flowr/ForClause.java
```diff
@@ -38,12 +38,24 @@ public ForClause(List<ForClauseVar> vars, ExpressionMetadata metadataFromContext
         if (vars == null || vars.isEmpty())
             throw new SemanticException("For clause must have at least one variable", metadataFromContext);
         this.forVariables = vars;
+
+        // chain forVariables with previousClause relationship
+        for (int varIndex = forVariables.size() - 1; varIndex > 0; varIndex--) {
```
ghislainfourny: This may need disentanglement later -- there might be several variables in the same clause, but that is still the same clause.

In the current implementation, these are treated as several clauses. In the long-term we will need to address this (you can simply raise an issue).

src/main/java/sparksoniq/jsoniq/compiler/translator/expr/postfix/extensions/PostfixExtension.java
```diff
@@ -45,6 +46,14 @@ public void setParent(PostFixExpression _parent) {
         parent = _parent;
     }
 
+    public PostfixExtension getPrevious() {
```
ghislainfourny: Why this special treatment?

PR #435, opened 2020-01-20
10 - Cosmetic improvements for StaticIsRDD

CanBerker:

PR #437, opened 2020-01-20
11-Materialize function

CanBerker:
Based on #435 

Implement materialize method in the RuntimeIterator. This replaces getItemsFromIteratorWithCurrentContext method.

### DIFF
--- a/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
+++ b/src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
@@ -76,7 +76,7 @@ public class FunctionItemCallIterator extends HybridRuntimeIterator {
     @Override
     public void openLocal() {
         if (_isPartialApplication) {
-            _functionBodyIterator = generatePartiallyAppliedFunction();
+            _functionBodyIterator = generatePartiallyAppliedFunction(_currentDynamicContextForLocalExecution);
         } else {
             _functionBodyIterator = _functionItem.getBodyIterator();
             _currentDynamicContextForLocalExecution = this.createNewDynamicContextWithArguments(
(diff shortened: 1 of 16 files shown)


src/main/java/sparksoniq/jsoniq/runtime/iterator/functions/FunctionItemCallIterator.java
```diff
@@ -149,7 +149,7 @@ private FunctionRuntimeIterator generatePartiallyAppliedFunction() {
                 partialAppParamNames.add(argName);
                 partialAppParamTypes.add(_functionItem.getSignature().getParameterTypes().get(i));
             } else {
-                List<Item> argValue = getItemsFromIteratorWithCurrentContext(argIterator);
+                List<Item> argValue = argIterator.materialize(context);
```
ghislainfourny: Yes, I like that. This may even address my concern in the other PR.

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ArrayRuntimeIterator.java
```diff
@@ -32,23 +32,27 @@
 
 public class ArrayRuntimeIterator extends LocalRuntimeIterator {
 
-
     private static final long serialVersionUID = 1L;
 
     public ArrayRuntimeIterator(RuntimeIterator arrayItems, IteratorMetadata iteratorMetadata) {
-        super(new ArrayList<>(), iteratorMetadata);
-        if (arrayItems != null)
+        super(null, iteratorMetadata);
```
ghislainfourny: Are you positive this won't lead to null pointer exceptions?

src/main/java/sparksoniq/jsoniq/runtime/iterator/primary/ArrayRuntimeIterator.java
```diff
@@ -32,23 +32,27 @@
...
+        }
     }
 
     @Override
     public Item next() {
         if (this._hasNext) {
-            List<Item> result = this.runChildrenIterators(this._currentDynamicContextForLocalExecution);
+            List<Item> result = new ArrayList<>();
+            if (!this._children.isEmpty()) {
+                result.addAll(this._children.get(0).materialize(_currentDynamicContextForLocalExecution));
```
ghislainfourny: Don't we need to iterator over the children here to keep the same behavior?

This may also be achieved with a materializeAllChildrenIteratorsIntoConcatenatedSequence() method (or a name like this).

Except of, of course, we know for sure that there is only child iterator anyway...

ghislainfourny: Checked offline: only one child

PR #438, opened 2020-01-20
12-Support RDD and DF in functionItem closures

CanBerker:


src/main/java/sparksoniq/jsoniq/item/FunctionItem.java
```diff
@@ -52,24 +55,45 @@
     // signature contains type information for all parameters and the return value
     private FunctionSignature signature;
     private RuntimeIterator bodyIterator;
-    private Map<String, List<Item>> variablesInClosure;
+    private Map<String, List<Item>> _localVariablesInClosure;
```
ghislainfourny: Could we encapsulate these three maps in a reusable class, that both function items and dynamic contexts can use? That may simplify the APIs.

ghislainfourny: TODO add an issue but merge this

CanBerker: @ghislainfourny Could you approve this PR?

PR #439, opened 2020-01-20
13-Implement materializeFirstItemOrNull

CanBerker:


src/main/java/sparksoniq/jsoniq/runtime/iterator/control/SwitchRuntimeIterator.java
```diff
@@ -89,38 +89,38 @@ private void initializeIterator(
...
-            throw new NonAtomicKeyException(
-                    "Invalid args. Switch condition can't be an array type",
-                    getMetadata().getExpressionMetadata()
-            );
-        } else if (testValue != null && testValue.isObject()) {
-            throw new NonAtomicKeyException(
-                    "Invalid args. Switch condition  can't be an object type",
-                    getMetadata().getExpressionMetadata()
-            );
-        }
```
CanBerker: testValue != null checks are unified in an outer if for both testValue and caseValue variables

src/main/java/sparksoniq/jsoniq/runtime/iterator/control/SwitchRuntimeIterator.java
```diff
@@ -89,38 +89,38 @@ private void initializeIterator(
...
-
-        for (RuntimeIterator caseKey : cases.keySet()) {
-            Item caseValue = getSingleItemFromIterator(
-                caseKey
-            );
+        Item testValue = test.materializeFirstItemOrNull(_currentDynamicContextForLocalExecution);
```
ghislainfourny: Could we check that 

switch((1, 2)) case 1 return 0 default return 2

returns an error?

In other words, do we check there are no extra items?

In a nutshell, we might need to have two rather than one methods depending on whether we check or not:

materializeOneItemOrNull (checks that there is no second one, i.e., hasNext = false after that)

materializeFirstItemOrNull (does NOT check that there is no second one, if we don't care)

ghislainfourny: TODO for later when singleton method is in place

PR #440, opened 2020-01-21
11-2-Cleanup static context visitor

CanBerker:
Based on #435

### DIFF
--- a/src/main/java/sparksoniq/semantics/visitor/StaticContextVisitor.java
+++ b/src/main/java/sparksoniq/semantics/visitor/StaticContextVisitor.java
@@ -99,9 +99,8 @@ public class StaticContextVisitor extends AbstractExpressionOrClauseVisitor<Stat
 
     @Override
     public StaticContext visitLetClauseVar(LetClauseVar expression, StaticContext argument) {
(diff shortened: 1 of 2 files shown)


src/main/java/sparksoniq/semantics/visitor/StaticContextVisitor.java
```diff
@@ -124,29 +91,24 @@ public StaticContext visitForClauseVar(ForClauseVar expression, StaticContext ar
...
         }
-        // TODO visit at...
-        this.visit(expression.getExpression(), argument);
         return result;
     }
 
     @Override
     public StaticContext visitLetClauseVar(LetClauseVar expression, StaticContext argument) {
-        StaticContext result = visitFlowrVarDeclaration(expression, argument);
         this.visit(expression.getExpression(), argument);
```
ghislainfourny: The resulting sc is not used? It is because this function has the side effect of storing the static context in the expression?

ghislainfourny: clarified offline (visiting checks for variables)

CanBerker: @ghislainfourny this pr needs approval as well

PR #441, opened 2020-01-21
12-2-Add static type checking for typeswitch default case

CanBerker:
This is based on #440 
Original implementation did not perform static checks on the condition and default case of the typeswitch expression. I have added these in this PR.


CanBerker:
@ghislainfourny waiting approval

PR #446, opened 2020-01-23
Rename DF to dataFrame in Flwor tuple and dynamic context

CanBerker:


src/main/java/sparksoniq/semantics/DynamicContext.java
```diff
@@ -103,14 +103,14 @@ public void setBindingsFromTuple(FlworTuple tuple, IteratorMetadata metadata) {
         return _rddVariableValues.keySet();
     }
 
-    public Set<String> getDFKeys() {
-        return _dfVariableValues.keySet();
+    public Set<String> getDataFrameKeys() {
```
ghislainfourny: getDataFrameVariableNames (key is a low-level detail).

CanBerker: Great suggestion

PR #447, opened 2020-01-23
Improve naming of orderbyClauseSparkIteratorExpression and its usage

CanBerker:
This one may be a matter of taste but I tried to improve the naming of orderbyClauseSparkIteratorExpression as it did not mean much to me in its original state. This PR does not aim to change the functionality. It merely renames and extracts a runtimeIterator to a variable in 2 UDFs

Basically this class consists of an orderByClause's subExpression (called orderByClauseExpr), and the iterator generated from the said expression. 

As a general remark, especially in the past few months, I have found that the 'expression' keyword to be used ambiguous ways in multiple contexts. While the meaning becomes clear once you are familiar with the context, for future maintainability of the project and onboarding of new developers, it may be very helpful to disentangle such meanings with hopefully better keyword selections. A low priority refactoring effort may be beneficial for minimizing the potential technical debt of the project in the future.

There are two sources that I identify and while the first one is easy and quick to address, the second one will most likely be a challenging effort with design & naming:
- Using expression keyword as a variable name for RuntimeIterators even though they are separate classes living in the same project
- Ambiguity between what you would describe as an expression from a language's lexical perspective, and what is defined as an expression in the semantic perspective of XQuery specification. More specifically, every language construct we have in Rumble inherits from ExpressionOrClause type, however, in the context of XQuery specs which we adhere to, they are not necessarily expressions. 'MainModule' would be the simplest example of the lying at the top of our grammar. Furthermore, adding FLWOR clauses (not expressions semantically) to the picture with their lexical expressions and sub-expressions (such as components of order by clause or multiple variable definitions in for or let clauses) this ambiguity quickly grows. 

@ghislainfourny What are you thoughts on this issue? If you find it valuable as well, we can extract this to an issue.

### DIFF
--- a/src/main/java/sparksoniq/spark/iterator/flowr/expression/OrderByClauseExprWithIterator.java
+++ b/src/main/java/sparksoniq/spark/iterator/flowr/expression/OrderByClauseExprWithIterator.java
@@ -22,7 +22,6 @@ package sparksoniq.spark.iterator.flowr.expression;
 
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.OrderByClauseExpr;
 import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
-import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
 
 import java.io.Serializable;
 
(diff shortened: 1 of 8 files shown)


ghislainfourny:
Yes, a cleanup is good. Overall we need consistency: classes ending with Expression must be those in the expression tree, i.e., derived from ExpressionOrClause. For everything else, it is good to find another naming. We can discuss more in our next meeting.

src/main/java/sparksoniq/spark/iterator/flowr/expression/OrderByClauseExprWithIterator.java
```diff
@@ -22,36 +22,32 @@
 
 import sparksoniq.jsoniq.compiler.translator.expr.flowr.OrderByClauseExpr;
 import sparksoniq.jsoniq.runtime.iterator.RuntimeIterator;
-import sparksoniq.jsoniq.runtime.metadata.IteratorMetadata;
 
 import java.io.Serializable;
 
-public class OrderByClauseSparkIteratorExpression implements Serializable {
+public class OrderByClauseExprWithIterator implements Serializable {
```
ghislainfourny: How about OrderByClauseAnnotatedChildIterator? (to completely drop expr)

CanBerker: I like this naming, updated as such

PR #448, opened 2020-01-23
Simplify typeSwitchExpression by removing redundant variableReferenceIterators

CanBerker:
Original typeSwitchExpression implementation created variableReferenceIterators only to access their names. I have eliminated the iterator generation and passed the names in directly instead. This not only improves performance by eliminating redundant object creation, but also improves correctness. The use of variableReference in the case of $a in the example below is to simply define the variable, not look it up from current (static or dynamic) context. VariableReferenceIterator should be used **only** for the return clauses of 2nd and 3rd cases where an actual variable referencing should be performed. 

typeswitch(4.234243e3)
case $a as hexBinary+ return "not this"
case $b as boolean | double? | base64Binary? return ($b cast as string) || " correct"
case $c as string | decimal? return $c
default return "DEF"

@ghislainfourny We can also include Andrea as reviewer if you think this would be relevant to his work/report


ghislainfourny:
@AndreaRinaldi1 @CanBerker is working on a high-level change regarding the static context for the needs of his thesis -- this is a suggestion for a performance improvement on the typeswitch operator.

PR #450, opened 2020-01-26
ML estimator

CanBerker:
Based on #449


ghislainfourny:
@CanBerker Is there documentation somewhere on libsvm support in Spark, with a more advanced level of details?

CanBerker:
Not that I know of at the moment. So far, my focus has been more on getting the initial prototype working. But for further developments I had a feeling that we would need a better understanding of libsvm and deep/sparse vectors fit into SparkML. 

Next, I will try investigating more in depth. I would also appreciate any further insights or feedback regarding these novel components.

PR #452, opened 2020-02-02
Improved exception handling for file missing exceptions

CanBerker:
Exceptions triggered by missing files specified by user queries where wrongfully triggering OurBadException prompting user to create an issue for rumble. 

For DataFrame enabled file-reading operations, the solution was simple as AnalysisException triggered by reading non-existing files is caught and converted into a proper Rumble exception.

For RDD enabled file-reading operations, the problem still awaits a clean solution as it's not as straightforwad as the previous case. The InvalidInputException is triggered by calling the Spark's textFile() in text-file and json-file use cases. This trigger occurs in JsoniqQueryExecutor when a collection on the RDD is performed(with take or collect), which is outside the iterator's scope that originally formulates the read operation in the creation of the RDD.

One solution I could think of was the validate the existence of the file at given URL before calling the textFile() using the URL. However since our files can be located on both local and HDFS filesystem this required a bit more work. I have implemented a helper class to check whether a given URL is valid for operation to proceed.


src/main/java/sparksoniq/spark/iterator/function/UrlValidator.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.File;
+import java.io.IOException;
+
+public class UrlValidator {
+
+    public static boolean isValid(String url) {
+        if (url.startsWith("file://") || url.startsWith("/") || url.startsWith("./")) {
+            File f = new File(url);
+            return f.exists();
+        }
```
CanBerker: @ghislainfourny what is our current status regarding workspace definition and using file names without any prefix such as ./ or // etc.  ie. referring as "file-in-workspace.txt". This may require an extra touch here

PR #456, opened 2020-02-04
Rename package for parser

ghislainfourny:
Also just renames.

(And fixing a typo in the doc on the ant install).

Some of the diff is due to regenerating with ant + applying spotless.


docs/install.md
```diff
@@ -50,7 +50,7 @@ For convenience, we have included the ANTLRv4 files for 4.5.3 and 4.7 in the lib
 
 From the root directory of the rumble local checkout, you first need to build the parser:
 
-    $ ant -buildfile build_antlr_parser.xml generate-parser -Dantlr.jar=/lib/antlr-4.7-complete.jar
+    $ ant -buildfile build_antlr_parser.xml generate-parser -Dantlr.jar=lib/antlr-4.7-complete.jar
```
CanBerker: Nice catch! Took me a few seconds to even notice it in the diff :)

src/main/java/org/rumbledb/parser/JsoniqLexer.java
```diff
@@ -1,18 +1,16 @@
...
-import org.antlr.v4.runtime.atn.ATNDeserializer;
-import org.antlr.v4.runtime.atn.LexerATNSimulator;
-import org.antlr.v4.runtime.atn.PredictionContextCache;
+import org.antlr.v4.runtime.CharStream;
+import org.antlr.v4.runtime.Token;
+import org.antlr.v4.runtime.TokenStream;
+import org.antlr.v4.runtime.*;
+import org.antlr.v4.runtime.atn.*;
 import org.antlr.v4.runtime.dfa.DFA;
+import org.antlr.v4.runtime.misc.*;
```
CanBerker: I have a tiny remark here. I noticed that the wildcard imports came back with these changes. I am wondering if these files are generated with the wildcard format by ANTLR or if your IDE of choice is collecting them together. 

Furthermore Google Java Style Guide, which I believed we were following with spotless (I may be wrong), seems to [not permit wildcard imports](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements).

In any case, since the files are auto generated this is definitely a minor issue but it just caught my attention.

ghislainfourny: This directly came from ANTLR, followed by applying spotless.

Are you aware of an ANTLR configuration parameter that I should set to prevent wildcards?

Thanks!

PR #457, opened 2020-02-05
Metadata

ghislainfourny:
Refactoring exception metadata as we had three classes doing the same thing.

Even though it is long, almost everything is just renaming and removing useless unwrapping calls.


src/main/java/org/rumbledb/exceptions/ExceptionMetadata.java
```diff
@@ -18,29 +18,50 @@
...
+ * @author Stefan Irimescu, Ghislain Fourny
+ *
+ */
+public class ExceptionMetadata implements Serializable {
 
     private static final long serialVersionUID = 1L;
-    private final int tokenLineNumber;
-    private final int tokenColumnNumber;
+    private final int _tokenLineNumber;
+    private final int _tokenColumnNumber;
```
CanBerker: Seeing the variable name here prompted me for a discussion I had on my mind for some time. Disclaimer: This is orthogonal to the scope of this PR.

I would like to inquire what is the motivation behind prefixing class variables with _ . I previously had the notion that _ prefixing is sometimes used as a means of identifying private variables especially in dynamically typed languages. However, I am not sure how much value this brings here in terms of readability or some other positive trait, at the cost of it incurring another rule to adhere to.

In any case, I also wanted to ask if we should work on this sometime soon to achieve consistency across the project. In the current state, I believe that there is very little consistency and this only grows with added code.

ghislainfourny: This is a common convention (for example also in some styleguides such as Google's) to distinguish between class variables on the one hand, and local variables and method parameters on the other hand. This makes methods more readable because it makes it easier to see what modifies the state of the object itself, as opposed to the logics of the method.

I notice that you (and others) make this distinction by prefixing object fields with "this.". The compiler will, however, not complain if this is not done.

I am open to other proposals and it is a good idea to have a convention through the repository. I do not feel strong about any specific approach more than another. If you want, you can try to look around for styleguides and investigate the range of solutions found in practice?

CanBerker: Thanks a lot for the detailed information it was very helpful. Indeed for C++ style guide of Google I was able to find the convention of using underscore as a postfix [here](https://google.github.io/styleguide/cppguide.html#Variable_Names). In Java style guide however, I could not find a similar recommendation in Google Java Style guide [here](https://google.github.io/styleguide/javaguide.html). Yet I believe this underscore convention may serve its purpose in many development environments, especially if we cannot find a viable alternative.

After some digging, I have managed to find an Eclipse JDT configuration that is capable of throwing an error or a warning when an [unqualified access to a member variable takes place](https://help.eclipse.org/2019-12/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fjdt%2Fcore%2FJavaCore.html&anchor=COMPILER_PB_UNQUALIFIED_FIELD_ACCESS)

This can be particularly useful if we can add this to our spotless customization as it already uses these eclipse JDT configurations but I have not managed to get it to work yet. I will keep trying but I was wondering whether @wscsprint3r could provided assistance in this area.

PR #458, opened 2020-02-05
Rename packages.

ghislainfourny:
Continuation of the tidying up. This is just a package rename, plus the javadoc for the Expression class.


src/main/java/org/rumbledb/expressions/Expression.java
```diff
@@ -18,21 +18,31 @@
...
+/**
+ * An expression is the first-class citizen in JSONiq syntax. Any expression
+ * returns a sequence of items.
+ * 
+ * Expressions form a tree, but this tree may contain other nodes, such as clauses
+ * and function declarations.
+ * 
+ * An expression is associated with a static context containing information such as
+ * the in-scope variables.
+ */
```
CanBerker: Very nice

PR #461, opened 2020-02-07
Simplify descendant computation.

ghislainfourny:
The current implementation of getChildren() and getDescendants() was a bit spaghetti-y and redundant.

Now, each node only needs to declare its children. The descendant gathering is done at a single place in the Node class, and it also ensures the depth-first search semantics. This function is actually only used in the FrontEnd tests.

Other cleanups:
- removing some unused methods
- making some classes that shouldn't be instantiated abstract


src/main/java/org/rumbledb/expressions/Expression.java
```diff
@@ -46,11 +46,6 @@ protected Expression(ExceptionMetadata metadata) {
         super(metadata);
     }
 
-    @Override
```
ghislainfourny: This implementation is not actually used; subclasses that do not override it have no children or should not be expressions.

src/main/java/org/rumbledb/expressions/flowr/FlworVarSequenceType.java
```diff
@@ -112,10 +113,17 @@ public SequenceType getSequence() {
...
 
     @Override
-    public List<Node> getDescendants(boolean depthSearch) {
+    public List<Node> getChildren() {
         return new ArrayList<>();
     }
 
+    @Override
+    public <T> T accept(AbstractNodeVisitor<T> visitor, T argument) {
+        // TO this class should not be an expression.
```
CanBerker: Is TO - TODO here ? Shall we create an issue for the mentioned refactoring?

src/main/java/org/rumbledb/expressions/flowr/FlworVarSingleType.java
```diff
@@ -79,10 +80,17 @@ public SingleType getSingleType() {
...
 
     @Override
-    public List<Node> getDescendants(boolean depthSearch) {
+    public List<Node> getChildren() {
         return new ArrayList<>();
     }
 
+    @Override
+    public <T> T accept(AbstractNodeVisitor<T> visitor, T argument) {
+        // TO this class should not be an expression.
```
CanBerker: Same comment as above

src/main/java/org/rumbledb/expressions/primary/PrimaryExpression.java
```diff
@@ -43,11 +43,13 @@ protected PrimaryExpression(ExceptionMetadata metadata) {
 
     @Override
     public <T> T accept(AbstractNodeVisitor<T> visitor, T argument) {
-        return visitor.visitDescendants(this, argument);
+        // TO this class should not be an expression.
```
CanBerker: Same comment as above

PR #464, opened 2020-02-08
Improve vectorUDT to rumbleItem mapping

CanBerker:
As we have discussed in the meeting I have initially implemented SparseVector to ObjectItem mapping.

As per our discussion, I have investigated whether dense and sparse vectors can co-exist within the same dataframe. And this is permitted not only across different columns, but within the same column as well (this can be observed [here](https://spark.apache.org/docs/latest/ml-features.html#pca). 

For this reason, I have extended the implementation to map sparseVectors to objectItems and denseVectors to arrayItems. Do you see any problems with this approach?


ghislainfourny:
Many thanks for looking into it.

Let us tentatively merge this PR, but keeping in mind that we can, and maybe even should, reconsider this design later. I could also imagine mapping both to objects -- the main drawback is inefficiency of dense array storage -- or both to a memory-efficient sparse array structure -- the main drawback is then serialization explosion for sparse arrays. Let's give us more time to refine.

PR #470, opened 2020-02-11
Tidy primary expression tree

ghislainfourny:
Clean up of the primary expression tree.

Two classes were not expressions (i.e., they do not return sequences of items):
- placeholder: removed, and the convention (local to function call expressions) is that null instead of an Expression is a ?
- pair constructor: this is not an expression, it was only used to "package" two expressions to pass them up to the object constructor expression in the visitor. I made this explicit in the visitor and this class is deleted.

All other classes were renamed to end with Expression. FunctionDeclaration is now InlineFunctionExpression (even though I understand it is also used for the prolog functions, but I'll follow up on that later).


src/main/java/org/rumbledb/expressions/module/Prolog.java
```diff
@@ -24,22 +24,22 @@
...
-import org.rumbledb.expressions.primary.FunctionDeclaration;
+import org.rumbledb.expressions.primary.InlineFunctionExpression;
 import sparksoniq.semantics.visitor.AbstractNodeVisitor;
 
 import java.util.ArrayList;
 import java.util.List;
 
 public class Prolog extends Expression {
 
-    private final List<FunctionDeclaration> functionDeclaration;
```
CanBerker: This is a probably a typo from earlier. Should we rename this to functionDeclaration(s)

src/main/java/org/rumbledb/expressions/control/TypeSwitchCaseExpression.java
```diff
@@ -5,20 +5,20 @@
...
+import org.rumbledb.expressions.primary.VariableReferenceExpression;
 import sparksoniq.semantics.visitor.AbstractNodeVisitor;
 
 import java.util.ArrayList;
 import java.util.List;
 
 public class TypeSwitchCaseExpression extends Expression {
 
-    private VariableReference variableReferenceNode;
+    private VariableReferenceExpression variableReferenceNode;
```
CanBerker: Cosmetic suggestion: shall we remove 'node' keyword from this variable's name? (adding expression or not is optional)

src/main/java/org/rumbledb/expressions/control/TypeSwitchCaseExpression.java
```diff
@@ -5,20 +5,20 @@
...
     private List<FlworVarSequenceType> union = new ArrayList<>();
     private final Expression returnExpression;
 
     public TypeSwitchCaseExpression(
-            VariableReference var,
+            VariableReferenceExpression var,
```
CanBerker: cosmetic: shall we use a fullname here instead of abbreviation, for compliance with rest of the project?

src/main/java/org/rumbledb/expressions/control/TypeSwitchExpression.java
```diff
@@ -15,13 +15,13 @@
     private final Expression testCondition;
     private final List<TypeSwitchCaseExpression> cases;
     private final Expression defaultExpression;
-    private final VariableReference varRefDefault;
+    private final VariableReferenceExpression varRefDefault;
```
CanBerker: While touching here: using full names again instead of abbreviation. 

We can also be more explicit with 'default' here as it relates to other 'cases'. How about defaultCaseExpression and defaultCaseVariableReference as variable names?

src/main/java/org/rumbledb/expressions/control/TypeSwitchExpression.java
```diff
@@ -15,13 +15,13 @@
...
 
     public TypeSwitchExpression(
             Expression testCondition,
             List<TypeSwitchCaseExpression> cases,
             Expression defaultExpression,
-            VariableReference varRefDefault,
+            VariableReferenceExpression varRefDefault,
```
CanBerker: full name as above while matching the final naming of your choice of the field

src/main/java/org/rumbledb/expressions/flowr/FlworVarDecl.java
```diff
@@ -48,7 +48,7 @@ private FlworVarDecl(FLWOR_CLAUSES clauseType, ExceptionMetadata metadata) {
 
     public FlworVarDecl(
             FLWOR_CLAUSES forVar,
-            VariableReference varRef,
+            VariableReferenceExpression varRef,
             FlworVarSequenceType seq,
```
CanBerker: fullnames

src/main/java/org/rumbledb/expressions/flowr/ForClauseVar.java
```diff
@@ -35,13 +35,13 @@
...
-    private final VariableReference positionalVariableReference;
+    private final VariableReferenceExpression positionalVariableReference;
 
     public ForClauseVar(
-            VariableReference varRef,
+            VariableReferenceExpression varRef,
             FlworVarSequenceType seq,
             boolean emptyFlag,
-            VariableReference atVarRef,
+            VariableReferenceExpression atVarRef,
```
CanBerker: full names.

Shall we rename atVarRef here to positionalVariableReference for clarity?

PR #471, opened 2020-02-11
Implement (annotate(object*, object) as object*) function: Local/RDD -> DataFrame conversion with schema

CanBerker:
~@ghislainfourny I have recently been working on item to native java type conversions. We need this to outsource the computation to the SparkML framework, as it would not be able to operate on rumble items. The conversion of common atomic items is already in place. I was hoping I could get your feedback in this week's meeting before moving on to the complex types.~

Edit: annotate(object*, object) as object* is designed to facilitate passing items to SparkML components.
Given sequence of items are converted to a DataFrame of items using the given schema. This underlying implementation detail needs to be exposed for the time being until full schema support is added to Rumble

### DIFF
--- a/src/main/java/sparksoniq/spark/ml/AnnotateFunctionIterator.java
+++ b/src/main/java/sparksoniq/spark/ml/AnnotateFunctionIterator.java
@@ -1,0 +1,67 @@
+package sparksoniq.spark.ml;
+
+import org.apache.spark.api.java.JavaRDD;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.rumbledb.api.Item;
+import org.rumbledb.exceptions.ExceptionMetadata;
+import org.rumbledb.exceptions.UnexpectedTypeException;
+import org.rumbledb.items.ObjectItem;
+import org.rumbledb.runtime.DataFrameRuntimeIterator;
+import org.rumbledb.runtime.RuntimeIterator;
+import sparksoniq.jsoniq.ExecutionMode;
+import sparksoniq.semantics.DynamicContext;
+import sparksoniq.spark.DataFrameUtils;
+
+import java.util.List;
+
+public class AnnotateFunctionIterator extends DataFrameRuntimeIterator {
+
+    private static final long serialVersionUID = 1L;
+
+    public AnnotateFunctionIterator(
+            List<RuntimeIterator> arguments,
+            ExecutionMode executionMode,
+            ExceptionMetadata iteratorMetadata
+    ) {
+        super(arguments, executionMode, iteratorMetadata);
+    }
+
+    @Override
+    public Dataset<Row> getDataFrame(DynamicContext context) {
+        RuntimeIterator inputDataIterator = this.children.get(0);
+        RuntimeIterator schemaIterator = this.children.get(1);
+        ObjectItem schemaItem;
+
+        // materialize singleton pattern
+        schemaIterator.open(context);
+        if (!schemaIterator.hasNext()) {
+            throw new UnexpectedTypeException(
+                    "Schema provided to annotate function can not be an empty sequence.",
+                    getMetadata()
+            );
+        }
+        schemaItem = (ObjectItem) schemaIterator.next();
+        if (schemaIterator.hasNext()) {
+            throw new UnexpectedTypeException(
+                    "Schema provided to annotate function must be a singleton object.",
+                    getMetadata()
+            );
+        }
+        schemaIterator.close();
+
+        if (inputDataIterator.isDataFrame()) {
+            // TODO: perform type checking and throw exception if it does not match the given dataframe
+            return inputDataIterator.getDataFrame(context);
+        }
+
+        if (inputDataIterator.isRDD()) {
+            JavaRDD<Item> rdd = inputDataIterator.getRDD(context);
+            return DataFrameUtils.convertItemRDDToDataFrame(rdd, schemaItem, getMetadata());
+        }
+
+        List<Item> items = inputDataIterator.materialize(context);;
+        return DataFrameUtils.convertLocalItemsToDataFrame(items, schemaItem, getMetadata());
+    }
+
+}
--- a/docs/Error codes.md
+++ b/docs/Error codes.md
@@ -25,6 +25,12 @@ An unrecognized parameter is used in query while operating with a Rumble ML clas
 - [RBML0003] - Invalid Rumble ML Param
 Provided parameter does not match the expected type or value for the referenced Rumble ML class.
 
+- [RBML0004] - Input is not a DataFrame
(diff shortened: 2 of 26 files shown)


src/main/java/sparksoniq/spark/ml/GetEstimatorFunctionIterator.java
```diff
@@ -112,7 +112,7 @@ public Item next() {
                 List<SequenceType> paramTypes = Collections.unmodifiableList(
                     Arrays.asList(
                         new SequenceType(
-                                new ItemType(ItemTypes.Item), // TODO: revert back to ObjectItem
```
CanBerker: This causes iterator to be wrapped again in TypePromotionIterator. Since this TPI is only RDD enabled(no DF implementation yet), the results are writtent as an RDD to the DynamicContext. This creates the need of reading out RDD or local variables from DC, converting them to a DataFrame of primitives, so that they can be passed into SparkML to be processed.

src/main/resources/test_files/runtime-spark/RumbleML/MLTransformerMultiple1.jq.disabled
```diff
@@ -1,8 +1,8 @@
 (:JIQS: ShouldRun; Output="" :)
 let $tokenizer := get-transformer("Tokenizer")
 let $hashingTF := get-transformer("HashingTF")
-let $data := (structured-json-file("./src/main/resources/queries/rumbleML/sample-ml-numeric-data.json"))
-let $intermediate := $tokenizer($data, {"inputCol": "values", "outputCol": "output"})
+let $data := structured-json-file("./src/main/resources/queries/rumbleML/sample-ml-string-data.json")
+let $intermediate := $tokenizer($data, {"inputCol": "sentence", "outputCol": "output"})
 return $hashingTF($intermediate, {"inputCol": "output", "numFeatures": 2})
```
CanBerker: This test is still disabled. 
Result of 1st transformer contains a column of Vector type. This vector type is converted to an ObjectItem and stored. However, for second operation this ObjectItem needs to be converted back into a native Java type so that it can be processed by SparkML. This is not implemented yet at the time of this comment.

src/main/resources/test_files/runtime-spark/RumbleML/MLTransformerVectorAssembler2.jq
```diff
@@ -0,0 +1,8 @@
+(:JIQS: ShouldRun; Output="({ "id" : 1, "age" : 20, "weight" : 68.8, "features" : [ 20, 68.8 ] }, { "id" : 2, "age" : 35, "weight" : 72.4, "features" : [ 35, 72.4 ] }, { "id" : 3, "age" : 50, "weight" : 76.3, "features" : [ 50, 76.3 ] })" :)
+let $transformer := get-transformer("VectorAssembler")
+let $rdd-data := json-file("./src/main/resources/queries/rumbleML/sample-ml-age-weight-data.json")
+for $i in $transformer(
+  $rdd-data,
+  {"inputCols": ["age", "weight"], "outputCol": "features"}
+)
+return $i
```
CanBerker: This test checks that atomic items (currently implemented) can be converted to native Java types to be processed in SparkML.

CanBerker:
@ghislainfourny Most of the machinery for the annotate function is in place. Date time types have caused some issues for now. I will take another look, but if it's not fruitful, we may need to re-evaluate the priority of these.

I would particularly appreciate the feedback on any missing atomic types or use cases with annotate() that I may have missed.

src/main/java/sparksoniq/spark/ml/AnnotateFunctionIterator.java
```diff
@@ -0,0 +1,67 @@
...
+                    "Schema provided to annotate function must be a singleton object.",
+                    getMetadata()
+            );
+        }
+        schemaIterator.close();
+
+        if (inputDataIterator.isDataFrame()) {
+            // TODO: alternatively, throw an 'IsAlreadyDataFrame" exception
+            return inputDataIterator.getDataFrame(context); // return unchanged
+        }
```
CanBerker: @ghislainfourny how should the desired behavior for annotate() be when called on an existing dataframe?

ghislainfourny: I would say we simply check that the types from the second parameter match the dataframe schema types (which should be easy), and if the check succeeds, return the dataframe as is.

src/main/java/sparksoniq/spark/ml/AnnotateFunctionIterator.java
```diff
@@ -0,0 +1,67 @@
...
+        super(arguments, executionMode, iteratorMetadata);
+    }
+
+    @Override
+    public Dataset<Row> getDataFrame(DynamicContext context) {
+        RuntimeIterator inputDataIterator = this.children.get(0);
+        RuntimeIterator schemaIterator = this.children.get(1);
+        ObjectItem schemaItem;
+
+        // materialize singleton pattern
```
ghislainfourny: Do we not have a materialize function for this that works out of the box? Any reason why we cannot use it?

CanBerker: As far as I know we have two materialize methods: one that materializes everything and another one that materializes the first result if it exists (otherwise returns null). The 'materialize singleton' as I call it has not been implemented yet. That is, materializes a singleton result and throws an error result set is not a singleton(either empty or multiple). Once implemented this can be used in multiple places across the project.

Regarding the signature and type-checks, I can take your word for them as I have not extensively worked with them. I have a feeling that since these mechanisms are in place, we can be safely remove manual checks that exist in the codebase in the near future. For the time being however, to be consistent with all the other iterators that implement these manual checks, I feel we should add these to the annotate function too. What do you think ?

src/main/java/sparksoniq/spark/ml/AnnotateFunctionIterator.java
CanBerker: That makes sense. If you don't mind it, I would like to merge this PR as is and address this type-checking in a separate PR. I believe some of the utilities I have been implementing for the vectorization of features PR can be re-used for this scheduled improvement.

src/main/java/sparksoniq/spark/ml/AnnotateFunctionIterator.java
ghislainfourny: Yes, it's fine to keep them for now, we can still remove later.

For the materialize method, if it is not implemented, it suffices to open an issue.

CanBerker:
@ghislainfourny I fixed two unqualified accesses from the sequenceType PR in the last commit. I'm sharing the image below from the maven compilation logs to document how to spot these easily.

image

PR #472, opened 2020-02-12
Refactor postfix expressions.

ghislainfourny:
Clean up of postfix expressions.

Extensions are a detail of the context-free grammar in the Abstract Syntax Tree that should not be visible beyond parsing. In the expression tree, each expression lives its own life. This also reduces the size of the code.


src/main/java/org/rumbledb/expressions/postfix/ArrayUnboxingExpression.java
```diff
@@ -18,41 +18,39 @@
...
+    public ArrayUnboxingExpression(Expression main, ExceptionMetadata metadata) {
+        super(main, metadata);
     }
 
-    public Expression getExpression() {
-        return this.expression;
+    @Override
+    public String serializationString(boolean prefix) {
+        String result = "(arrayUnboxing [ ])";
+        return result;
```
CanBerker: I think the serialization is missing a call to the serialization of the "main" Expression

ghislainfourny: Done, thanks a lot!

src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
```diff
@@ -757,10 +755,10 @@ public Void visitCastExpr(JsoniqParser.CastExprContext ctx) {
     @Override
     public Void visitUnaryExpr(JsoniqParser.UnaryExprContext ctx) {
         // TODO [EXPRVISITOR] jump from unary to postfix
-        PostFixExpression mainExpression;
+        Expression mainExpression;
         UnaryExpression node;
         this.visitSimpleMapExpr(ctx.main_expr);
-        mainExpression = (PostFixExpression) this.currentExpression;
```
CanBerker: This is an extremely beneficial cleanup. Looking at the previous code I was confounded what kind of a special relationship between a PostfixExpression and Unary Expression existed and I had to check our grammar.

src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
```diff
@@ -111,7 +110,6 @@
     private Expression otherCurrentExpression;
 
     private PrimaryExpression currentPrimaryExpression;
-    private PostfixExtension currentPostFixExtension;
```
CanBerker: I am very glad that number of these special variables in the state of the class has decreased.

src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
ghislainfourny: Thanks!

Indeed, this was a detail from the parser, which is a "breach in encapsulation".

In the parser and the AST, the precedence of the syntactic blocks is important. In the expression tree, there is no need for this any more, because the precedence is implicit from the tree itself: every child is just an Expression, period.

ghislainfourny:
Thanks, @CanBerker !

PR #473, opened 2020-02-13
Remove parenthesized expression

ghislainfourny:
ParenthesizedExpressions are also not expressions, they are a syntactic artifact. This cleans it up.


src/main/java/org/rumbledb/expressions/module/MainModule.java
```diff
@@ -33,20 +33,20 @@
 public class MainModule extends Expression {
 
     private final Prolog prolog;
```
CanBerker: For completeness, should we also refactor Prolog type with Expression type here?

src/main/java/sparksoniq/jsoniq/compiler/JsoniqExpressionTreeVisitor.java
```diff
@@ -109,7 +107,6 @@
     private Expression currentExpression;
     private Expression otherCurrentExpression;
 
-    private PrimaryExpression currentPrimaryExpression;
```
CanBerker: Nice !

ghislainfourny: I anticipated that you would like this one!

src/main/java/org/rumbledb/expressions/module/MainModule.java
ghislainfourny: Actually a prolog is not an expression (but this part is not refactored yet).

PR #474, opened 2020-02-14
Simplify expression tree.

ghislainfourny:
This removes inactive expressions -- this was also a syntactic artifact. If an expression was inactive, then it is now replaced with its only child expression, which makes expression trees much smaller.

I also moved over some classes to the new package scheme.


ghislainfourny:
> This is very nice! The smaller expression tree will make reasoning and debugging so much easier in the future.

Thank you!

PR #475, opened 2020-02-14
Rename packages.

ghislainfourny:
These are package renames (nothing else).


src/main/java/org/rumbledb/items/ItemComparatorForSequences.java
```diff
@@ -36,7 +36,7 @@
      * It compares 2 atomic items (non-null)
      * Non-atomics and nulls throw an exception
      *
-     * @return -1 if v1 < v2; 0 if v1 == v2; 1 if v1 > v2;
+     * @return -1 if v1 &lt; v2; 0 if v1 == v2; 1 if v1 &gt; v2;
```
CanBerker: where the semicolons after $lt and gt$ intentional here? If so, shall we use a different separator between the cases?

PR #477, opened 2020-02-21
Fix switch expression to allow several cases and a few cosmetic fixes.

ghislainfourny:


src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -105,11 +105,8 @@
 public class TranslationVisitor extends org.rumbledb.parser.JsoniqBaseVisitor<Void> {
```
CanBerker: This comment is orthogonal to the content of this PR, but I thought it could be addressed here.

Since we have seen quite a bit of refactoring in this area, I was wondering if we should add a few lines of comments here explaining what "Translation" in this context is.

I have seen the comment in the line above "// used to build AST, will override methods" but did not find it very useful and even potentially incorrect. If I am not mistaken, isn't this visitor used to build the Node(previously ExpressionOrClause) tree. AST should already be generated by the ANTLR parser. What do you think?

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -248,39 +245,39 @@ else if (content instanceof JsoniqParser.TypeSwitchExprContext)
     // TODO [EXPRVISITOR] count
     @Override
     public Void visitFlowrExpr(JsoniqParser.FlowrExprContext ctx) {
```
CanBerker: The naming here has been catching my attention for quite some time. Since we are touching FLWOR expression and clauses here, I was wondering if it might be the correct time to address it. Since we have a policy of using full names for variables functions etc., could we refactor visitFlowrExpr with visitFlowrExpression.

Additionally, the name Flowr is also a typo that exists also in the grammar file. How would you feel about opening an issue and/or addressing this in a new PR (as autogenerated files will probably show up as quite a lot of line changes).

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -248,39 +245,39 @@ else if (content instanceof JsoniqParser.TypeSwitchExprContext)
...
-        FlworClause startClause, childClause;
-        List<FlworClause> contentClauses = new ArrayList<>();
+        Clause startClause, childClause;
```
CanBerker: Google's java style guide is strictly in favor of using [one variable per declaration](https://google.github.io/styleguide/javaguide.html#s4.8.2-variable-declarations). Could we put these variables on different lines?

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -865,20 +867,8 @@ public Void visitObjectConstructor(JsoniqParser.ObjectConstructorContext ctx) {
         return null;
     }
 
-    // TODO[EXPRVISITOR]? not supported in Pair constructor
     @Override
     public Void visitPairConstructor(JsoniqParser.PairConstructorContext ctx) {
```
CanBerker: This method looks empty now. Can it be completely removed?

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -1124,13 +1125,6 @@ public Void visitSwitchExpr(JsoniqParser.SwitchExprContext ctx) {
 
     @Override
     public Void visitSwitchCaseClause(JsoniqParser.SwitchCaseClauseContext ctx) {
```
CanBerker: This method too is now empty and can be removed.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -847,9 +844,14 @@ public Void visitObjectConstructor(JsoniqParser.ObjectConstructorContext ctx) {
...
+                if (currentPair.lhs != null) {
+                    this.visitExprSingle(currentPair.lhs);
+                    keys.add(this.currentExpression);
+                } else {
+                    keys.add(new StringLiteralExpression(currentPair.name.getText(), createMetadataFromContext(ctx)));
+                }
+                this.visitExprSingle(currentPair.rhs);
+                values.add(this.currentExpression);
             }
             this.currentExpression = new ObjectConstructorExpression(keys, values, createMetadataFromContext(ctx));
```
CanBerker: I have a question here. The objectConstructorExpression seems to take lists of expressions for its keys and values it is always passed a singleton(this is the only place this class seems to be used). I was wondering if this is in your agenda for issues to address.

src/main/java/org/rumbledb/expressions/flowr/FlworVarDecl.java
```diff
@@ -31,7 +31,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
-public abstract class FlworVarDecl extends FlworClause {
+public abstract class FlworVarDecl extends Clause {
```
CanBerker: Same idea with using full variable names: shall we refactor this to FlworVariableDeclaration?

src/main/java/org/rumbledb/expressions/flowr/Clause.java
```diff
@@ -27,13 +27,13 @@
  * GRAMMAR:flowrExpression
  */
 
-public abstract class FlworClause extends Node {
+public abstract class Clause extends Node {
```
CanBerker: I was initially reserved about using 'Clause' name instead of FlworClause with the worry of this keyword having a different meaning in some niche part of the codebase. However, after seeing that visitSwitchCaseClause is refactored out ('SwitchCaseClause' lives effectively only in the grammar now), I am more confident in this change.

src/main/java/org/rumbledb/expressions/flowr/Clause.java
```diff
@@ -27,13 +27,13 @@
  * GRAMMAR:flowrExpression
  */
```
CanBerker: I believe this comment may be outdated and misleading. I'd suggest removing it.

src/main/java/org/rumbledb/expressions/control/SwitchCase.java
```diff
@@ -21,33 +21,32 @@
...
-import org.rumbledb.expressions.AbstractNodeVisitor;
 import org.rumbledb.expressions.Expression;
-import org.rumbledb.expressions.Node;
 
 import java.util.ArrayList;
 import java.util.List;
 
-public class SwitchCaseExpression extends Expression {
+/**
+ * This is a helper class that organizes the children expressions of a child expression.
```
CanBerker: Did you mean to say "Children expressions of a **Switch** expression?"

src/main/java/org/rumbledb/expressions/control/SwitchExpression.java
```diff
@@ -74,8 +76,22 @@ public Expression getDefaultExpression() {
...
+        result.append("(SwitchExpression switch ");
+        result.append(this.testCondition.serializationString(false));
+        for (SwitchCase c : this.cases) {
+            for (Expression e : c.getConditionExpressions()) {
+                result.append("case ");
+                result.append(e.serializationString(false));
+                result.append(" ");
+            }
+            result.append("return ");
+            result.append(c.getReturnExpression().serializationString(false));
```
CanBerker: I have a feeling if we were to extract the body of this SwitchCase c: ... loop to a serializationgString() method for SwitchCase class (even though it doesn't live in the Node hierarchy as a class), it could make for nicer code. How would you feel about this?

src/main/java/org/rumbledb/expressions/control/SwitchCase.java
```diff
@@ -21,33 +21,32 @@
...
+ * From a tree perspective, all expressions in there are considered
```
CanBerker: For added clarity, could you spell out what 'there' means in the context. ie. 'in this class' or 'in conditionExpressions' etc.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Great idea. DONE.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: You are correct. However, I think that this means changing the g4 grammar. This should be done in a separate PR in which all we do is change the grammar.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Agreed. DONE.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Apparently, it can because it is not abstract. DONE.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: DONE.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: I am not sure I understand your question. Could you elaborate? Thanks!

src/main/java/org/rumbledb/expressions/flowr/FlworVarDecl.java
ghislainfourny: I have not refactored that part yet, but this is on the agenda as well.

src/main/java/org/rumbledb/expressions/flowr/Clause.java
ghislainfourny: Indeed the use of "clause" for switch case is a syntactic detail that disappears in the expression tree.

src/main/java/org/rumbledb/expressions/flowr/Clause.java
ghislainfourny: Good catch. DONE.

src/main/java/org/rumbledb/expressions/control/SwitchCase.java
ghislainfourny: Good catch, thanks.

src/main/java/org/rumbledb/expressions/control/SwitchExpression.java
ghislainfourny: I consciously avoided that in order to not give a higher status to the class SwitchCase than it should have. This is not a node but only a helper class. This is why I am not overloading it with more methods.

Let me know if you still feel that we could add a serialization method there that the SwichExpression class can use as a utility function.

src/main/java/org/rumbledb/expressions/control/SwitchCase.java
ghislainfourny: DONE.

ghislainfourny:
> It's very nice to see that ancient TODO cleaned up :) PR looks good. I have listed many small issues with the hopes of getting them right once so that we do not have to revisit these core components for some time. I hope you don't find the review overly picky.

These are useful comments that contributed to improving the code. Thank you!

ghislainfourny:
@CanBerker I'll look at the other PR after this one is merged. Then, hopefully the diff will be fixed (it seems to be a GitHub issue).

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
CanBerker: Exactly. Will we be addressing this soon or shall we create an issue to tackle it later?

src/main/java/org/rumbledb/expressions/control/SwitchExpression.java
CanBerker: Your reasoning sounds reasonable and I am glad to hear that this was a conscious decision. We can keep this as is.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
CanBerker: I misinterpreted the code here last time I looked at it. On a second look, it looks good as it is.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: I'll open an issue.

PR #478, opened 2020-02-21
Clean expression tree (typeswitch).

ghislainfourny:
Continuing the cleanup.

A new project is about to start, which will use the expression tree -- I am making sure it is as clean as possible.


src/main/java/org/rumbledb/expressions/control/SwitchExpression.java
```diff
@@ -81,16 +80,14 @@ public String serializationString(boolean prefix) {
...
-    		result.append("return ");
-    		result.append(c.getReturnExpression().serializationString(false));
+        for (SwitchCase c : this.cases) {
+            for (Expression e : c.getConditionExpressions()) {
+                result.append("case ");
+                result.append(e.serializationString(false));
+                result.append(" ");
+            }
+            result.append("return ");
+            result.append(c.getReturnExpression().serializationString(false));
```
CanBerker: This will potentially not cause issues with merging but just a pointer: The formatting was already fixed in the latest commit of SwitchFix. I believe this branch is not up to date with SwitchFix (target branch) and shows a larger diff.

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -941,28 +940,19 @@ public RuntimeIterator visitSwitchExpression(SwitchExpression expression, Runtim
...
     @Override
     public RuntimeIterator visitTypeSwitchExpression(TypeSwitchExpression expression, RuntimeIterator argument) {
-        List<TypeSwitchCase> cases = new ArrayList<>();
-        for (TypeSwitchCaseExpression caseExpression : expression.getCases()) {
-            String caseVariableName = null;
-            if (caseExpression.getVariableReference() != null) {
-                caseVariableName = caseExpression.getVariableReference().getVariableName();
-            }
+        List<TypeSwitchRuntimeIteratorCase> cases = new ArrayList<>();
+        for (TypeswitchCase caseExpression : expression.getCases()) {
```
CanBerker: I really like that you have removed the unnecessary complexity with the variable name usage.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -1131,52 +1135,46 @@ public Void visitTypeSwitchExpr(JsoniqParser.TypeSwitchExprContext ctx) {
...
                 defaultCase,
-                varRefDefault,
+                defaultVariableName,
                 createMetadataFromContext(ctx)
         );
         return null;
     }
 
     @Override
     public Void visitCaseClause(JsoniqParser.CaseClauseContext ctx) {
```
CanBerker: This method can be removed  I think.

src/main/java/org/rumbledb/expressions/control/TypeSwitchExpression.java
```diff
@@ -5,57 +5,56 @@
...
 import java.util.List;
 
 public class TypeSwitchExpression extends Expression {
 
     private final Expression testCondition;
-    private final List<TypeSwitchCaseExpression> cases;
+    private final List<TypeswitchCase> cases;
     private final Expression defaultExpression;
-    private final VariableReferenceExpression defaultVariableReferenceExpression;
+    private final String defaultVariableName;
```
CanBerker: I have an alternative suggestion on how to handle the default case. How about instead of storing the defaultVariableName and defaultExpression as separate fields, we refactored these into a TypeSwitchCase as well? I believe this would make code much cleaner in some places (ie. in visitors)

src/main/java/org/rumbledb/expressions/control/TypeswitchCase.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This is a helper class that organizes the children expressions of a typeswitch expression.
+ * From a tree perspective, all expressions in there are considered
+ * to be direct children of the TypeswitchExpression.
+ */
+public class TypeswitchCase {
```
CanBerker: Is this name intentional or should the s in TypeSwitchCase be capitalized?

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
ghislainfourny: Happy to hear!

src/main/java/org/rumbledb/expressions/control/SwitchExpression.java
ghislainfourny: Yes, I was aware of this. This should be good now.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Yes, DONE.

src/main/java/org/rumbledb/expressions/control/TypeSwitchExpression.java
ghislainfourny: Very good idea. this is done.

src/main/java/org/rumbledb/expressions/control/TypeswitchCase.java
ghislainfourny: I am trying to unify with the original spec (https://www.w3.org/TR/xquery-30/#id-typeswitch) where this is capitalized as Typeswitch

ghislainfourny:
> This PR seems to be not fully up to date with the target branch so I did my best to not repeat the same comments from #477 . It would be very nice to fix issues with #477 and merge it into this once so that I can take a final and quick look at hopefully what is to be a much smaller diff.

Now the diff should be clean again. Thanks!

src/main/java/org/rumbledb/expressions/control/TypeswitchCase.java
CanBerker: Alright that sounds great! Thanks for the clarification

src/main/java/org/rumbledb/expressions/control/TypeswitchCase.java
```diff
@@ -0,0 +1,59 @@
...
+            String variableName,
+            List<FlworVarSequenceType> union,
+            Expression returnExpression
+    ) {
+        this.variableName = variableName;
+        this.union.addAll(union);
+        this.returnExpression = returnExpression;
+    }
+
+    // for default
```
CanBerker: Very tiny remark. As code evolves this comment may become outdated and even misleading. I believe with modern IDEs developers could easily lookup where this constructor is used.

src/main/java/org/rumbledb/expressions/control/TypeswitchCase.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * This is a helper class that organizes the children expressions of a typeswitch expression.
+ * From a tree perspective, all expressions in there are considered
+ * to be direct children of the TypeswitchExpression.
+ */
+public class TypeswitchCase {
+
+    private String variableName;
+    private List<FlworVarSequenceType> union = new ArrayList<>();
```
CanBerker: How about we moved this initialization to the 1st constructor(with 3 parameters) and explicitly set the union field to null in the 2nd constructor(with 2 parameters, used for the default case). I believe this would both avoid the redundant initialization for the default case as well as potentially making the intention of the 2nd constructor clearer.

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -940,32 +940,23 @@ public RuntimeIterator visitSwitchExpression(SwitchExpression expression, Runtim
...
-        String defaultCaseVariableName = null;
-        if (expression.getDefaultVariableReferenceExpression() != null) {
-            defaultCaseVariableName = expression.getDefaultVariableReferenceExpression().getVariableName();
-        }
-        defaultCase = new TypeSwitchCase(
-                defaultCaseVariableName,
-                this.visit(expression.getDefaultExpression(), argument)
+        TypeswitchRuntimeIteratorCase defaultCase = new TypeswitchRuntimeIteratorCase(
+                expression.getDefaultCase().getVariableName(),
+                this.visit(expression.getDefaultCase().getReturnExpression(), argument)
```
CanBerker: Looks much cleaner now!

src/main/java/org/rumbledb/expressions/control/TypeSwitchExpression.java
CanBerker: I'm glad to hear that. Please refer to my other comment for hopefully addressing this overhead as well.

src/main/java/org/rumbledb/expressions/control/TypeswitchCase.java
ghislainfourny: This is now removed.

src/main/java/org/rumbledb/expressions/control/TypeswitchCase.java
ghislainfourny: Very good remark. This is done.

CanBerker:
I fixed the null pointer exception rooted in typeSwitchRuntimeIterator. The PR is approved. Feel free to merge if you are on board with this last commit.

ghislainfourny:
Thanks!

PR #479, opened 2020-02-21
Clean up expression tree (eliminate duplicate sequence type classes).

ghislainfourny:


src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -214,14 +213,14 @@ private RuntimeTupleIterator visitFlowrClause(
         if (clause instanceof ForClause) {
             for (ForClauseVar var : ((ForClause) clause).getForVariables()) {
                 RuntimeIterator assignmentIterator = this.visit(var.getExpression(), argument);
-                if (var.getAsSequence() != null && var.getAsSequence().getSequence() != null) {
+                if (var.getSequenceType() != null && var.getSequenceType() != null) {
```
CanBerker: Both sides of && are identical now

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -239,14 +238,14 @@ private RuntimeTupleIterator visitFlowrClause(
         } else if (clause instanceof LetClause) {
             for (LetClauseVar var : ((LetClause) clause).getLetVariables()) {
                 RuntimeIterator assignmentIterator = this.visit(var.getExpression(), argument);
-                if (var.getAsSequence() != null && var.getAsSequence().getSequence() != null) {
+                if (var.getSequenceType() != null && var.getSequenceType() != null) {
```
CanBerker: Both sides of && are identical now

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -268,14 +267,14 @@ private RuntimeTupleIterator visitFlowrClause(
                 RuntimeIterator groupByExpressionIterator = null;
                 if (groupByExpression != null) {
                     groupByExpressionIterator = this.visit(groupByExpression, argument);
-                    if (var.getAsSequence() != null && var.getAsSequence().getSequence() != null) {
+                    if (var.getSequenceType() != null && var.getSequenceType() != null) {
```
CanBerker: Both sides of && are identical now

src/main/java/sparksoniq/semantics/types/SequenceType.java
```diff
@@ -56,14 +60,23 @@ public boolean isEmptySequence() {
...
+        if (this.isEmptySequence) {
+            throw new OurBadException("Empty sequence type has no arity");
+        }
         return this.arity;
     }
 
     public boolean isSubtypeOf(SequenceType superType) {
+        if (this.isEmptySequence) {
+            return this.arity == Arity.OneOrZero || this.arity == Arity.ZeroOrMore;
+        }
```
CanBerker: Empty sequence gets its arity set to null during construction. Don't these boolean expressions always evaluate to false?

src/main/java/sparksoniq/semantics/types/SequenceType.java
```diff
@@ -74,6 +87,12 @@ public boolean equals(Object o) {
         if (!(o instanceof SequenceType))
             return false;
         SequenceType sequenceType = (SequenceType) o;
```
CanBerker: naming this variable 'other' could be nice for readability

src/main/java/sparksoniq/semantics/types/SequenceType.java
```diff
@@ -22,6 +22,8 @@
```
CanBerker: SequenceType.ToString() is used in multiple places for serialization of the SequenceType. I think this method should be implemented to properly resemble (or match) the serialization of the removed class FlworVarSequenceType.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -130,7 +130,7 @@ public StaticContext visitFunctionDeclaration(InlineFunctionExpression expressio
             .forEach(
                 (paramName, flworVarSequenceType) -> functionDeclarationContext.addVariable(
                     paramName,
-                    flworVarSequenceType.getSequence(),
+                    flworVarSequenceType,
```
CanBerker: It would be great if we could rename this variable. It looks out of place as is.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -226,9 +226,9 @@ public StaticContext visitCountClause(CountClause expression, StaticContext argu
     private StaticContext visitFlowrVarDeclaration(FlworVarDecl expression, StaticContext argument) {
         StaticContext result = new StaticContext(argument);
         // TODO for now we only suppot as/default, no inference, flags
-        SequenceType type = expression.getAsSequence() == null
-            ? new SequenceType()
-            : expression.getAsSequence().getSequence();
+        SequenceType type = expression.getSequenceType() == null
+            ? SequenceType.emptySequence
```
CanBerker: I love the use of static class field instead of a magic value here.

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
ghislainfourny: Good catch, done.

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
ghislainfourny: Good catch, done.

src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
ghislainfourny: Good catch, done.

src/main/java/sparksoniq/semantics/types/SequenceType.java
ghislainfourny: That was a mistake, it should be the superType's arity.

src/main/java/sparksoniq/semantics/types/SequenceType.java
ghislainfourny: Done

src/main/java/sparksoniq/semantics/types/SequenceType.java
ghislainfourny: I added toString(), with a serialization closer to JSONiq.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: Now I made sure sequence types are never null (this was a mistake in the translator that was incorrectly setting them to null rather than item*).

src/main/java/org/rumbledb/expressions/flowr/ForClauseVar.java
```diff
@@ -53,10 +53,9 @@ public ForClauseVar(
         // because TreatIterator is wrapping the whole assignment expression,
         // meaning there is not one TreatIterator for each variable we loop over.
         if (sequenceType != null)
```
CanBerker: I believe this sequenceType != null check is redundant now. Super()'s constructor ensures that sequence type should not be null

src/main/java/org/rumbledb/runtime/operational/InstanceOfIterator.java
```diff
@@ -68,10 +68,12 @@ public Item next() {
...
-                ItemType itemType = this.sequenceType.getItemType();
-                for (Item item : items) {
-                    if (!item.isTypeOf(itemType)) {
-                        return ItemFactory.getInstance().createBooleanItem(false);
+                if (!this.sequenceType.isEmptySequence()) {
+                    ItemType itemType = this.sequenceType.getItemType();
+                    for (Item item : items) {
+                        if (!item.isTypeOf(itemType)) {
+                            return ItemFactory.getInstance().createBooleanItem(false);
+                        }
```
CanBerker: From a correctness perspective this is absolutely acceptable. However, I get a little bit of spaghetti code feeling when I look at how this code block and isInvalidArity() function works together to cover various combinations. What do you think of this ?

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -1036,20 +1031,14 @@ public Void visitNamedFunctionRef(JsoniqParser.NamedFunctionRefContext ctx) {
 
     @Override
     public Void visitInlineFunctionExpr(JsoniqParser.InlineFunctionExprContext ctx) {
-        Map<String, FlworVarSequenceType> fnParams = new LinkedHashMap<>();
-        FlworVarSequenceType fnReturnType = new FlworVarSequenceType(
-                mostGeneralSequenceType,
-                createMetadataFromContext(ctx)
-        );
+        Map<String, SequenceType> fnParams = new LinkedHashMap<>();
+        SequenceType fnReturnType = SequenceType.mostGeneralSequenceType;
```
CanBerker: Beautiful

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
CanBerker: I am very glad that we have clarified that and the code already looks much nicer. I believe there may be a few spots where sequenceType can still be null in an expression. I will mark these with comments. Once we ensure that TranslationVisitor assigns every sequenceType instance to a non-null value. I believe the need for the 'type' variable will go away in this 'StaticContextVisitor.java' file at lines 229 and 265.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -1183,12 +1172,12 @@ public Void visitQuantifiedExpr(JsoniqParser.QuantifiedExprContext ctx) {
...
             Expression varExpression;
-            FlworVarSequenceType sequenceType = null;
+            SequenceType sequenceType = null;
             this.visitVarRef(currentVariable.varRef());
             varRef = (VariableReferenceExpression) this.currentExpression;
             if (currentVariable.sequenceType() != null) {
                 this.visitSequenceType(currentVariable.sequenceType());
-                sequenceType = (FlworVarSequenceType) this.currentExpression;
+                sequenceType = this.currentSequenceType;
             }
```
CanBerker: sequenceType may be null here.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -917,29 +915,26 @@ public Void visitContextItemExpr(JsoniqParser.ContextItemExprContext ctx) {
     @Override
     public Void visitSequenceType(JsoniqParser.SequenceTypeContext ctx) {
         if (ctx.item == null)
-            this.currentExpression = new FlworVarSequenceType(createMetadataFromContext(ctx));
+            this.currentSequenceType = SequenceType.emptySequence;
```
CanBerker: Very clean

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -655,13 +653,13 @@ public Void visitInstanceOfExpr(JsoniqParser.InstanceOfExprContext ctx) {
     @Override
     public Void visitTreatExpr(JsoniqParser.TreatExprContext ctx) {
         Expression mainExpression;
-        FlworVarSequenceType sequenceType;
+        SequenceType sequenceType;
```
CanBerker: Can we move these 2 variable declarations inside the if statement as they have no functionality outside that scope.

CanBerker: Can we move these 2 variable declarations inside the if statement as they have no functionality outside that scope.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -445,15 +443,15 @@ public Void visitOrderByExpr(JsoniqParser.OrderByExprContext ctx) {
...
         Expression expr = null;
         String uri = null;
         this.visitVarRef(ctx.var_ref);
         var = (VariableReferenceExpression) this.currentExpression;
 
         if (ctx.seq != null) {
             this.visitSequenceType(ctx.seq);
-            seq = (FlworVarSequenceType) this.currentExpression;
+            seq = this.currentSequenceType;
         }
```
CanBerker: else condition missing, sequenceType will be left as null

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
CanBerker: Can we move these 2 variable declarations inside the if statement as they have no functionality outside that scope.

src/main/java/org/rumbledb/runtime/operational/TreatIterator.java
```diff
@@ -45,8 +45,10 @@ public TreatIterator(
         this.iterator = iterator;
         this.sequenceType = sequenceType;
         this.shouldThrowTreatException = shouldThrowTreatException;
-        this.itemType = this.sequenceType.getItemType();
-        this.sequenceTypeName = ItemTypes.getItemTypeName(this.itemType.getType().toString());
+        if (!this.sequenceType.isEmptySequence()) {
+            this.itemType = this.sequenceType.getItemType();
```
CanBerker: I see a null pointer exception risk here. this.itemType is not initialized (is null) when sequenceType is emptySequence.

At line 116, this.nextResult.isTypeOf(this.itemType) call is made.

For example, if this.nextResult is a BooleanItem, isTypeOf method on the boolean item does not handle null parameters and it will cause a NPE. 

It may be that execution does not reach there if sequence is specified to be an emptySequence. But it would take me a bit more time to fully grasp all the implications, so I wanted to get your opinion as well.

ghislainfourny: Yes, thank you for being thorough. I actually had this in mind and did a few more fixes yesterday, but github had internal server errors. My last commit fixes issues related to this.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: Yes, my last commit fixes this and makes sure sequenceType is always initiated in the translator.

src/main/java/org/rumbledb/expressions/flowr/FlworVarDecl.java
```diff
@@ -37,7 +37,7 @@
     protected Expression expression;
 
     // asSequence is null by default if the type of the variable in the for/let/groupBy clause is not specified.
```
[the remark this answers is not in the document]
ghislainfourny: Correct, it is removed.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: This is fixed.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/runtime/operational/InstanceOfIterator.java
ghislainfourny: This is fixed.

src/main/java/org/rumbledb/expressions/flowr/ForClauseVar.java
ghislainfourny: Done.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -226,9 +226,7 @@ public StaticContext visitCountClause(CountClause expression, StaticContext argu
     private StaticContext visitFlowrVarDeclaration(FlworVarDecl expression, StaticContext argument) {
         StaticContext result = new StaticContext(argument);
         // TODO for now we only suppot as/default, no inference, flags
-        SequenceType type = expression.getSequenceType() == null
-            ? SequenceType.emptySequence
-            : expression.getSequenceType();
+        SequenceType type = expression.getSequenceType();
```
CanBerker: we can even remove this line and just call expression.getSequenceType() at line 232

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -262,9 +260,7 @@ public StaticContext visitQuantifiedExpressionVar(QuantifiedExpressionVar expres
 
         // create a child context, add the variable and return it
         StaticContext result = new StaticContext(argument);
-        SequenceType type = expression.getSequenceType() == null
-            ? SequenceType.emptySequence
-            : expression.getSequenceType();
+        SequenceType type = expression.getSequenceType();
```
CanBerker: Same comment as above

src/main/java/org/rumbledb/expressions/flowr/FlworVarDecl.java
```diff
@@ -36,8 +36,8 @@
     protected VariableReferenceExpression variableReferenceExpression;
     protected Expression expression;
 
-    // asSequence is null by default if the type of the variable in the for/let/groupBy clause is not specified.
-    protected FlworVarSequenceType asSequenceType;
+    // asSequence is never null.
```
CanBerker: The field name seems out of date here. But in any case, I think we can remove this comment fully as the code throws an error if sequenceType is null in the constructor.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -333,14 +331,16 @@ public Void visitForClause(JsoniqParser.ForClauseContext ctx) {
     @Override
     public Void visitForVar(JsoniqParser.ForVarContext ctx) {
         VariableReferenceExpression var, atVarRef = null;
```
CanBerker: Could we fix this to have one variable per declaration.

src/main/java/org/rumbledb/runtime/operational/TreatIterator.java
CanBerker: This is discussed offline. An error check is performed in ther code preventing the execution from reaching this NPE.

src/main/java/org/rumbledb/runtime/operational/TreatIterator.java
```diff
@@ -32,7 +32,7 @@
 
     private Item nextResult;
     private Item currentResult;
-    private int childIndex;
+    private int resultCount;
```
CanBerker: I have renamed this variable to resultCount as I strongly believe that childIndex is a misleading name here. Within the context of iterators child refers to a child iterator. However, this index was merely counting the results that are found so far.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -636,13 +640,13 @@ public Void visitMultiplicativeExpr(JsoniqParser.MultiplicativeExprContext ctx)
     @Override
     public Void visitInstanceOfExpr(JsoniqParser.InstanceOfExprContext ctx) {
         Expression mainExpression;
-        FlworVarSequenceType sequenceType;
+        SequenceType sequenceType;
```
CanBerker: Can we apply the idea we have applied to TreatExpr to all applicable places in this file? These variables do not need to be defined in this outermost context. They can simply be defined within the if statement where they are used.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: Good point, done

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: done

src/main/java/org/rumbledb/expressions/flowr/FlworVarDecl.java
ghislainfourny: done

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: done

src/main/java/org/rumbledb/runtime/operational/TreatIterator.java
ghislainfourny: great, thanks!

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: DONE

ghislainfourny:
> Code has already come a long way. I think as the last item of the list, TreatIterator needs some refactoring to improve code quality as well as correctness. In the current state
> `parallelize((1,2,3)) treat as ()` this query fails with `Error [err: RBST0004 ] Empty sequence type has no arity` triggered by the `checkMoreThanOneItemSequence` at line 95 of TreatIterator.java
> 
> Firstly, TreatExpression.java determines that if sequenceType is emptySequence, this iterator always evaluates locally. So for the RDD execution, we should definitely add an OurBadException if execution reaches there when sequenceType is emptySequence. At the moment, a similar check is performed in TreatAsClosure. It should be removed from there and moved to getRDD() method as an OurBadException.
> 
> Then, we should re-iterate on how `check...` functions are implemented and used. Hopefully, this will be very beneficial for avoiding bugs similar to the one given above in the future. I think better naming for the functions as well as better division of tasks to the functions could go a long way. What do you think of these suggested changes?

Yes, this will be addressed separately.

CanBerker:
Thanks for taking the time to address all the issues. This PR is good to go

CanBerker:
Thanks for taking the time to address all the issues. This PR is good to go

CanBerker:
Thanks for taking the time to address all the issues. This PR is good to go

PR #480, opened 2020-02-21
Start of research project 1: implementation of functions related to current date and time.

ioanas96:
Added implementation for the following functions:

- current-dateTime

- current-date

- current-time


src/main/java/org/rumbledb/items/ItemFactory.java
```diff
@@ -62,6 +62,10 @@ public Item createDayTimeDurationItem(Period p) {
         return new DayTimeDurationItem(p);
     }
 
+    public Item createCurrentDateTimeItem() {
```
ghislainfourny: I would avoid overloading the item factory with more specific functions, and rather use 

createDateTimeItem(new DateTime(), true)

on the caller side instead.

Otherwise, it creates a precedent and this class may become too big.

The same for the other two methods.

ioanas96: Thank you for your comments.
I have made the changes, but this required to change the visibility modifiers for the constructors of DateTimeItem, DateItem and TimeItem. I checked with other item classes and the public modifiers seem to be used in the majority of places. Please let me know if this was not a desired change in this case.

src/main/java/org/rumbledb/runtime/functions/datetime/CurrentDateFunctionIterator.java
```diff
@@ -0,0 +1,39 @@
...
+            ExceptionMetadata iteratorMetadata
+    ) {
+        super(arguments, executionMode, iteratorMetadata);
+    }
+
+    @Override
+    public Item next() {
+        if (this.hasNext) {
+            this.hasNext = false;
+            return new DateItem(new DateTime(), true);
```
ghislainfourny: This needs to go through the factory -- this is why constructors are not public.

Like so:

ItemFactory.getInstance().createDateItem(new DateTime(), true);

(the same goes for the other three functions).

Thanks!

ioanas96: Thank you for the comment, I made the changes just now.

PR #481, opened 2020-02-24
Brp2 dev

ghislainfourny:


ghislainfourny:
@CodingKaiser as you can see, the tests are failing. The reason is our cosmetic syntax checker, spotless.

If you run mvn spotless:apply on your laptop and commit, this should then work again.

src/main/java/org/rumbledb/runtime/functions/base/Functions.java
```diff
@@ -123,19 +123,7 @@
...
-import org.rumbledb.runtime.functions.strings.NormalizeSpaceFunctionIterator;
-import org.rumbledb.runtime.functions.strings.StartsWithFunctionIterator;
-import org.rumbledb.runtime.functions.strings.StringFunctionIterator;
-import org.rumbledb.runtime.functions.strings.StringJoinFunctionIterator;
-import org.rumbledb.runtime.functions.strings.StringLengthFunctionIterator;
-import org.rumbledb.runtime.functions.strings.SubstringAfterFunctionIterator;
-import org.rumbledb.runtime.functions.strings.SubstringBeforeFunctionIterator;
-import org.rumbledb.runtime.functions.strings.SubstringFunctionIterator;
-import org.rumbledb.runtime.functions.strings.TokenizeFunctionIterator;
+import org.rumbledb.runtime.functions.strings.*;
```
CanBerker: If I may, we agreed against using wildcard imports in the project some time ago. I have verified that this isn't picked up by spotless for the time being but I will investigate if we can add this to our configurations as well. For the time being, it would be great if we could keep explicit imports as they are.

If you are using IntelliJ IDEA, [this link](https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports) can be helpful. Otherwise, I'm sure other IDE's support something similar as well.

ghislainfourny: Thanks, @CanBerker 

Could you also comment on the parser files? Were they also post-processed with IntelliJ to remove the wildcards (in this case, I do not think they were changed, so their change can simply be reverted)?

CanBerker: Gladly. I was curious about these as well. Once formatted by spotless, as you mentioned, all changes except the wildcards disappear. It's extremely likely the wildcards used by ANTLR were removed by IDEA at some point.

ghislainfourny: Thank you -- it would be ideal if we could find a way to do the same with eclipse, in order to stick to the no-wildcards convention. An alternative is to completely remove these automatically generated files from the repository, as previously suggested by Stefan. Then there is no need to worry about it any more (but it may require setting up the test queue appropriately).

CanBerker: I have looked into this but it appears that Eclipse does not expose a configuration that we can use in a project-wide manner. Neither eclipse JDT through the spotless plugin, nor the eclipse IDE itself exposes such an option. 

We could consider removing the auto-generated and discuss its tradeoffs. Yet, for this case, I believe it would not help anything. The wildcard imports popped up in the Functions.java and they can appear basically in any file. Removing auto-generated files simply reduce the number of files wildcard imports could pop up in by 4. 

For the time being, I am planning to add guidelines for eclipse and IDEA configurations to the Rumble's wiki as I do not see a programmatic solution presently available. For the time being, I believe we have to stick to peer-reviews in enforcing this formatting decision.

CodingKaiser: I noticed this change as well. It appears using code completion for classes not yet imported automatically changes the explicit imports for classes associated with that package to a wildcard import, at least given the default settings in IntelliJ. I'll pay attention to this for future pull requests.

CanBerker: Thanks for your attention on the matter. Now that you have applied spotless, the only changes that remain in the auto-generated files seem to be the wildcard imports. If you have applied the configuration as shown in this [wiki page](https://github.com/RumbleDB/rumble/wiki/Wildcard-imports), running `IntelliJ - optimize imports` on auto-generated files should remove the wild-card imports. This should hopefully bring the files back to the original state. If all goes well, changes in the auto-generated files in this PR should disappear.

CodingKaiser:
@ghislainfourny Not sure if you have seen my email yet, but I had some problems using `boolean?`as a return type, specifically for the `codepoint-equal` function. In the version of the above code, I have simply changed the return type to `item?` as a temporary solution. Otherwise, I have made all the necessary changes and this branch should be ready to merge, unless I have missed something.

CanBerker:
Hey @CodingKaiser, I ran into to this issue with another type yesterday and I can point you to the cause. If you check Functions.java line 329, you will see a static map for sequence types. The type you are trying to use "boolean?" is not defined here and is probably causing null pointer exceptions. If you add the type to the map, this problem should go away. Does it help?

@ghislainfourny I'm not sure if there was a reason why full cartesian product of types and arities was not added to this map previously. I would suggest either adding all of these, and/or implementing a better 'reading out from the map' logic so that a specific error is thrown when the type does not exist in the map. Should I create an issue for this ?

CodingKaiser:
> Hey @CodingKaiser, I ran into to this issue with another type yesterday and I can point you to the cause. If you check Functions.java line 329, you will see a static map for sequence types. The type you are trying to use "boolean?" is not defined here and is probably causing null pointer exceptions. If you add the type to the map, this problem should go away. Does it help?
> 
> @ghislainfourny I'm not sure if there was a reason why full cartesian product of types and arities was not added to this map previously. I would suggest either adding all of these, and/or implementing a better 'reading out from the map' logic so that a specific error is thrown when the type does not exist in the map. Should I create an issue for this ?

Yes, thanks! I've made the change.

ghislainfourny:
@CodingKaiser thanks! You can just press the Merge pull request button -- it will merge everything into your main branch branch-research-project-2. Then you can create more branches and more pull requests against branch-research-project-2.

@CanBerker indeed this will be worth an improvement -- maybe using enums rather than strings would make the code more robust to avoid any null pointer exceptions. You could open an issue if you want.

PR #486, opened 2020-02-26
Improve readability of forClause schema generation from tuples

CanBerker:
ForClauseLocalToRowClosure is renamed to ForClausedLocalTupleToRowClosure (clarity for additions to coming with RumbleML)

Removed redundancy of schema generation for each tuple.

Removed wildcard imports from Functions.java . @ghislainfourny as far as I know, eclipse permits at most 99 imports from a single file until it switches to wildcards. This number can be configured up to 999 in IntelliJ. So every time you optimize imports with your IDE for the Functions.java file, wildcards might come back.


ghislainfourny:
Thanks! I can set it to 999.

PR #494, opened 2020-03-01
Implementation of format functions for date and time

ioanas96:


src/main/java/org/rumbledb/items/DateItem.java
```diff
@@ -26,13 +26,13 @@ public DateItem() {
         super();
     }
 
-    DateItem(DateTime value, boolean hasTimeZone) {
+    public DateItem(DateTime value, boolean hasTimeZone) {
```
ghislainfourny: The "public" keywords can now be removed again.

PR #501, opened 2020-03-03
Item types

ghislainfourny:
There were 5-6 classes doing the same thing. All we need is ItemType and SequenceType, and their package was renamed to the new rumble scheme.

The ItemTypes enum is still exposed in the code, but in the future this will also be encapsulated.


CanBerker:
Can you update the PR with the recent master, I see changes from previous PRs.

src/main/java/org/rumbledb/runtime/operational/CastableIterator.java
```diff
@@ -21,12 +22,12 @@
 
 public class CastableIterator extends LocalRuntimeIterator {
     private static final long serialVersionUID = 1L;
-    private final SingleType singleType;
+    private final SequenceType singleType;
```
CanBerker: For consistency with CastIterator, I would recommend renaming this to sequenceType as well. Alternatively, CastIterator can keep the name singleType if it is meaningful.

src/main/java/org/rumbledb/runtime/functions/base/Functions.java
```diff
@@ -299,30 +299,30 @@
...
+        itemTypes.put("date", ItemType.dateItem);
+        itemTypes.put("time", ItemType.timeItem);
 
-        itemTypes.put("hexBinary", new ItemType(ItemTypes.HexBinaryItem));
-        itemTypes.put("base64Binary", new ItemType(ItemTypes.Base64BinaryItem));
+        itemTypes.put("hexBinary", ItemType.hexBinaryItem);
+        itemTypes.put("base64Binary", ItemType.base64BinaryItem);
 
-        itemTypes.put("null", new ItemType(ItemTypes.NullItem));
+        itemTypes.put("null", ItemType.nullItem);
```
CanBerker: Beautiful

src/main/java/org/rumbledb/types/ItemTypes.java
```diff
@@ -18,7 +18,7 @@
  *
  */
 
-package sparksoniq.semantics.types;
+package org.rumbledb.types;
 
 public enum ItemTypes {
```
CanBerker: I was initialy investigating the idea whether ItemTypes  enum and ItemType classes could be combined to generate a single source of truth. But from a code organization perspective the existing organization seems nice. 

I have an alternative suggestion however. The way things are, ItemTypes feels like an underlying type utility enum for ItemType class. Hence, wouldn't it be nice if we made the access modifier package private here? I believe developers should never access ItemTypes directly, and always go through the ItemType class. Do you think that makes a sound design or am I missing something?

src/main/java/org/rumbledb/items/YearMonthDurationItem.java
```diff
@@ -58,26 +57,26 @@ public boolean isTypeOf(ItemType type) {
...
+        return itemType.getType() == ItemTypes.YearMonthDurationItem
             ||
-            itemType.equals(AtomicTypes.DayTimeDurationItem)
+            itemType.getType() == ItemTypes.DayTimeDurationItem
             ||
-            itemType.equals(AtomicTypes.DurationItem)
+            itemType.getType() == ItemTypes.DurationItem
             ||
-            itemType.equals(AtomicTypes.StringItem);
+            itemType.getType() == ItemTypes.StringItem;
```
CanBerker: Is this intentional to push performance to its limits through bypassing the equals method implementation on ItemType?

Otherwise, I would suggest simply calling `itemType.equals(ItemType.stringItem)`. This facilitates the package private access I mentioned in my other comment by hiding the ItemTypes class from direct access outside ItemType class.

src/main/java/org/rumbledb/expressions/operational/CastExpression.java
```diff
@@ -1,18 +1,25 @@
...
 
-    public CastExpression(Expression mainExpression, FlworVarSingleType singleType, ExceptionMetadata metadata) {
+    public CastExpression(Expression mainExpression, SequenceType type, ExceptionMetadata metadata) {
         super(mainExpression, Operator.CAST, metadata);
-        this.singleType = singleType;
+        this.type = type;
+        if (type.getArity() != Arity.OneOrZero && type.getArity() != Arity.One) {
+            throw new OurBadException(
+                    "Cast expressions cannot have an arity of more than one, something went wrong with the parser."
+            );
```
CanBerker: Very good! Always nice to see more checks in place

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -839,17 +838,19 @@ public SequenceType processSequenceType(JsoniqParser.SequenceTypeContext ctx) {
...
+        ItemType itemType = new ItemType(ctx.item.getText());
         if (ctx.question.size() > 0) {
-            return new FlworVarSingleType(item, true, createMetadataFromContext(ctx));
+            return new SequenceType(
+                    itemType,
+                    SequenceType.Arity.OneOrZero
+            );
         }
-        return new FlworVarSingleType(item, createMetadataFromContext(ctx));
+        return new SequenceType(itemType);
```
CanBerker: Much cleaner

src/main/java/org/rumbledb/types/ItemTypes.java
ghislainfourny: Exactly -- the masterplan is to completely hide ItemTypes from outside ItemType.

However, this is a lot of work to do because it is used at a lot of places, and this will be done later in a separate PR.

src/main/java/org/rumbledb/items/YearMonthDurationItem.java
ghislainfourny: Absolutely -- this will all go away with a comparison on the ItemType objects, in a future PR. Then, it will be an equals() call.

PR #502, opened 2020-03-03
RumbleML: Implement Feature Vector Generation with FeaturesCol param

CanBerker:
In SparkML featuresCol is a string parameter that contains the name of the column that is expected to contain Vector type data.

In RumbleML this implementation is changed:
- featuresCol: the same name is used
- It expects a String[] rather than a single string
- The columns specified in the String[] are put through VectorAssembler to form a vector in a new column having a UUID  as name for collision safety.
- Internally featuresCol param is set to match this UUID name to reference the column with features.
- After transformer's execution, this intermediate column of generated features is dropped. This achieves a clean output from user's perspective.

Additionally:
- A lot more SparkML tests are added
- SparkML Test files are structured into folders with better file naming and comments
- Small bugfix with annotate() function where conversion from double to int caused exceptions.


src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -27,8 +27,12 @@
...
         JavaRDD<Row> rowRDD = itemRDD.map((Function<Item, Row>) item -> ItemParser.getRowFromItem(item));
         try {
-            return SparkSessionManager.getInstance().getOrCreateSession().createDataFrame(rowRDD, schema);
-        } catch (IllegalArgumentException ex) {
+            Dataset<Row> result = SparkSessionManager.getInstance()
+                .getOrCreateSession()
+                .createDataFrame(rowRDD, schema);
+            result.take(1);
+            return result;
+        } catch (ClassCastException | IllegalArgumentException ex) {
```
CanBerker: Bugfix with Annotate:
HybridRunTimeIterator::getRDD() method receives the dataframe result of annotate() calculated here. Then df.javaRDD() Spark action is called on the result which triggers a ClassCastException when it fails to cast doubles to integers(as specified by the schema). 
image

I have added the take(1) action here to trigger the action and hence the exception earlier, so that we can give the nice error message as shown in the image above. It may not be ideal, but let me know if you agree with this solution or if you have a better alternative @ghislainfourny .

src/main/resources/test_files/runtime-spark/RumbleML/AnnotateFunction/MLAnnotateForChaining10.jq.disabled
```diff
@@ -0,0 +1,11 @@
+(:JIQS: ShouldRun; Output="" :)
+let $data := structured-json-file("./src/main/resources/queries/rumbleML/sample-ml-data-sentence.json")
+
+let $tokenizer := get-transformer("Tokenizer")
+let $intermediate-data := $tokenizer($data, {"inputCol": "sentence", "outputCol": "output"})
+let $df-intermediate-data := annotate($intermediate-data, {"id": "integer", "sentence": "string", "output": "[\"string\"]"})
```
ghislainfourny: when we do support it, the syntax will be "output" : [ "string" ] :)

src/main/java/sparksoniq/spark/ml/ApplyTransformerRuntimeIterator.java
```diff
@@ -40,6 +45,45 @@ public ApplyTransformerRuntimeIterator(
...
+
+                featuresColValue = RumbleMLUtils.convertParamItemToJava(
+                    featuresColParamName,
+                    featureColumnsParam,
+                    rumbleMLFeatureColumnsJavaTypeName,
+                    getMetadata()
+                );
+            }
+
+            // TODO Optimization: Don't generate features column if it is already in feature format (ie. loaded by as
```
ghislainfourny: Perfect!

ghislainfourny:
@CanBerker it looks great.

About the outside look, I am wondering if we should not give a new name to this column to avoid any confusion, like "featureColumns", and later also support "featureCols" with its standard semantics if a user prefers to submit arrays of doubles (that can be converted to vectors internally) directly.

src/main/java/sparksoniq/spark/DataFrameUtils.java
ghislainfourny: This is fine for now (correctness first) and we can optimize later.

ghislainfourny:
After today's discussion, we decided to keep the same name featuresCol, taking an array of strings. These should be column names with a double type, even though internally any column names that have a vector-of-double type (which the end user cannot create directly) will be automatically handled as well.

src/main/resources/test_files/runtime-spark/RumbleML/AnnotateFunction/MLAnnotateForChaining10.jq.disabled
CanBerker: Fixed

PR #505, opened 2020-03-04
Annotate() function: type checking for dataframes

CanBerker:


src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -274,6 +275,58 @@ private static void addValue(
...
+            case "null":
+                return DataTypes.NullType;
+            case "date":
+                return DataTypes.DateType;
+            case "datetime":
+                return DataTypes.TimestampType;
+            case "hexbinary":
+                return DataTypes.BinaryType;
+            case "[\"item\"]":
+                return vectorType;
```
CanBerker: This is a temporary representation for the vector types. Once we implement the support for array type representation, extra set of quotes will go away. If you have any preferences on how to represent vectors, please let me know

src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -23,9 +24,18 @@
...
+        JavaRDD<Row> rowRDD = itemRDD.map(
+            new Function<Item, Row>() {
+                private static final long serialVersionUID = 1L;
+
+                @Override
+                public Row call(Item item) {
+                    return ItemParser.getRowFromItem(item);
+                }
+            }
+        );
```
CanBerker: Fixed the inline Spark function definition.

src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -84,39 +101,101 @@ private static StructType generateSchemaFromSchemaItem(ObjectItem schemaItem, Ex
...
-        } else if (itemTypeName.equals("string")) {
-            return DataTypes.StringType;
-        } else if (itemTypeName.equals("null")) {
-            return DataTypes.NullType;
-        } else if (itemTypeName.equals("date")) {
-            return DataTypes.DateType;
-        } else if (itemTypeName.equals("datetime")) {
-            return DataTypes.TimestampType;
-        } else {
-            throw new RuntimeException("Unexpected item type found.");
```
CanBerker: This is moved to itemParser

src/main/java/sparksoniq/spark/ml/RumbleMLUtils.java
```diff
@@ -174,7 +174,7 @@ private static Object convertRumbleAtomicToJava(AtomicItem atomicItem, String ja
             case "double":
                 return atomicItem.castAs(AtomicTypes.DoubleItem).getDoubleValue();
             case "long":
-                return (long) atomicItem.castAs(AtomicTypes.DoubleItem).getDoubleValue();
+                return atomicItem.castAs(AtomicTypes.DecimalItem).getDecimalValue().longValue();
```
CanBerker: Fixed a misconception on DecimalItem to DataType.Long conversion.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
ghislainfourny: From a data independence perspective, vectors are a hidden physical detail -- this representation is thus only a debug feature known from us only and users should not be able to create vectors directly (only VectorAssembler would do this from double fields, as well as other transformers and estimators that preprocess double fields to vectors to pass it on to Spark ML).

As such, this name is fine for now, or we could also call it "vector" -- but we should not document it anywhere in the end user documentation.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -274,6 +275,58 @@ private static void addValue(
...
+        } else if (DataTypes.NullType.equals(dataType)) {
+            return "null";
+        } else if (DataTypes.DateType.equals(dataType)) {
+            return "date";
+        } else if (DataTypes.TimestampType.equals(dataType)) {
+            return "datetime";
+        } else if (DataTypes.BinaryType.equals(dataType)) {
+            return "hexbinary";
+        } else if (vectorType.equals(dataType)) {
+            return "[\"item\"]";
```
ghislainfourny: If my understanding is correct that this is also only used to check the dataframe schema types against the types passed to annotate, this is fine too (also not documented).

src/main/resources/test_files/runtime-spark/RumbleML/AnnotateFunction/MLAnnotateDF2.jq
```diff
@@ -0,0 +1,7 @@
...
+annotate(
+    libsvm-file("./src/main/resources/queries/rumbleML/sample-libsvm-data-short.txt"),
+    {"label": "double", "features": "[\"item\"]"}
```
ghislainfourny: Yes -- and for our eyes only.

src/main/java/sparksoniq/spark/ml/RumbleMLUtils.java
ghislainfourny: Nice fix.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
CanBerker: I would much rather use the vector name here as well. But the problem is, then the user would be directly exposed the vector keyword. The reason being, annotate() function applies verifies the schema on the input dataframe with this new implementation instead of directly passing it along. Hence if users writes the following query:  
```
annotate(libsvm-file(....), {"label": "integer", "features": "?????" )
```
 They have to specify the correct type for features which maps to vectors. This is why I initially opted for passing dataframe as is, instead of applying a typecheck.

CanBerker: After giving it some thought I have an idea for a different representation here. Data read by libsvm is always read into a SparseVector which is converted to an object in Rumble Data Model. Hence, how about we expose the keyword "object" or "{ }" in the short run?

ghislainfourny: Your idea goes in the right direction, but let me think about this more. In the meantime, feel free to go ahead and merge this branch with the understanding that users do not know about this special type, and are only permitted to use annotate on flat objects with known atomic types, which is our main use case (which the output of libsvm is not).

In other words, let us just consider "[\"item\"]" a user-defined type known only to us -- in the schema language, type names can be any string anyway.


About this query:

annotate(libsvm-file(....), {"label": "integer", "features": "?????" )

Is there any scenario when the end user strictly has to use this, and otherwise would be incapable of executing a specific ML query? The output of libsvm-file can be directly processed by estimators and transformers and annotate leaves it unchanged anyway, correct?

CanBerker: Precisely. This addition is more aimed at completeness rather than addressing a need. Libsvm data can be directly used in ML. However if the user decides to call annotate on it directly, which only serves to validate the schema, only then we need to cover the vector case.

ghislainfourny: Perfect, thank you. Then it's all good.

PR #506, opened 2020-03-05
Bringing up to spec so that boolean() has the expected behaviour

CodingKaiser:
For boolean(), the rules [here](https://www.w3.org/TR/xpath-functions-30/#func-boolean) indicate that types derived from string and anyURI should also be handled. I have implemented this here in preparation for my anyURI implementation by calling `canBePromotedTo`, for which anyURI should return string, if I have interpreted this correctly.


ghislainfourny:
Yes, this is perfect. Thank you!

PR #507, opened 2020-03-05
Item types encapsulation

ghislainfourny:
This is the follow up that encapsulates the ItemTypes enum.


src/main/java/org/rumbledb/types/ItemType.java
```diff
@@ -217,4 +217,46 @@ public String toString() {
...
+        String itemPostfix = "Item";
+        if (!itemPostfix.equals(fullTypeName) && fullTypeName.endsWith("Item")) {
+            return Character.toLowerCase(fullTypeName.charAt(0))
+                +
+                fullTypeName.substring(1, fullTypeName.length() - itemPostfix.length());
+        }
+        return fullTypeName;
+    }
+
+    enum ItemTypes {
```
CanBerker: I think this can be even private now.

src/main/java/org/rumbledb/types/ItemType.java
```diff
@@ -217,4 +217,46 @@ public String toString() {
         }
         throw new OurBadException("Unrecognized type.");
     }
+
+    public static String convertClassNameToItemTypeName(String fullTypeName) {
```
CanBerker: I think the parameter should be renamed to className here.

Furthermore, I believe the variable in line 222 was defined to use within the fullTypeName.endsWith(....) call. 

Lastly, how about we use the [uncapitalize() function](http://commons.apache.org/proper/commons-lang//apidocs/org/apache/commons/lang3/StringUtils.html#uncapitalize-java.lang.String-) from apache commons? I believe this would make the body of the if statement more readable.

src/main/java/org/rumbledb/types/ItemType.java
```diff
@@ -125,12 +125,12 @@ public ItemType(String text) {
     }
```
CanBerker: I actually have a bigger concern regarding this class in conjunction with the convertClassNameToItemTypeName() method and ItemType(String text) constructor. The manipulation of these string values which appear to be 'magic values' feels a bit scattered across the place. By all means, at the moment, they function correctly. However the fragility of convertClassNameToItemTypeName() as it directly handles classnames, and the duplication of string magic values in this file and in the ItemParser methods make me uneasy.

I am currently investigating the idea of adding the string representations of item types directly to the itemType class as a field. This way, each item type will statically have their String serialization value and these static and final variables can act as a single source of truth across the project for these serialized strings. Furthermore, if we were to place these within a map that serves as a catalogue(potentially with string representations as the keys), the itemType constructor with the string parameter and the convertClassNameToItemTypeName() method can be rendered redundant. What do you think ?

ghislainfourny: You are right, I was keeping it for later but this is now addressed too. This method is gone.

Instead, items can "self-declare" their dynamic type, but this should only be used for error messages as JSONiq does not expose dynamic types other than through instance of tests.

ghislainfourny: Regarding string representation of item types : this already exists (toString() does that). The problem was to display error messages on items.

src/main/java/org/rumbledb/api/Item.java
```diff
@@ -546,4 +546,13 @@ public Item modulo(Item other) {
...
     }
+
+    /**
+     * Returns the dynamic type of the item (for error message purposes).
+     * 
+     * @return the dynamic type as an item type.
+     */
+    public ItemType getDynamicType() {
+        throw new UnsupportedOperationException("Operation not defined");
+    }
```
CanBerker: I really like the precision of your name choice here with the 'dynamicType'. It was a fun exercise for me to evaluate the dynamic vs static nature of types in this context.

src/main/java/org/rumbledb/api/Item.java
```diff
@@ -306,7 +306,7 @@ public boolean canBePromotedTo(ItemType type) {
     public Item promoteTo(ItemType type) {
         if (!this.canBePromotedTo(type)) {
             throw new RuntimeException(
-                    ItemType.convertClassNameToItemTypeName(this.getClass().getSimpleName())
+                    this.getDynamicType().toString()
```
CanBerker: I love how short and easy to understand this line has become

PR #510, opened 2020-03-05
Implemented anyURI atomic type with corresonding constructor

CodingKaiser:


src/main/java/org/rumbledb/items/AnyURIItem.java
```diff
@@ -0,0 +1,140 @@
...
+    public AnyURIItem() {
+        super();
+    }
+
+    public AnyURIItem(String value) {
+        super();
+        this.value = parseAnyURIString(value);
+    }
+
+    static URI parseAnyURIString(String anyURIString) throws IllegalArgumentException {
```
ghislainfourny: private

PR #516, opened 2020-03-08
Extend annotate() function and Implement DecimalType support

### LINKED ISSUE #484: Convert Spark SQL DecimalType to JSONiq decimal
it is missing in the mapping

### LINKED ISSUE #513: Improve annotate() exceptions to hide underlying implementation details
```
annotate(
      (
          {"id": 1, "age":  20, "weight": 68.8},
          {"id": 2, "age":  35, "weight": 72.4},
          {"id": 3, "age":  50, "weight": 76.3}
      ),
       {"id": "integer", "age": "integer", "weight": "integer"}
)
```
>⚠️  ️Error [err: RBML0005]LINE:1:COLUMN:0:Schema error in annotate(): Error while applying the >schema: java.lang.Double cannot be cast to java.lang.Integer

Implement better exception messages that do no expose the underlying implementation details

### LINKED ISSUE #517: Improve Item class exceptions
Item class exceptions throw RuntimeExceptions when an invalid access for value is performed, eg. calling getStringValue on an Integer. 
Add the item causing the exception to the exception message. Preferably use a more specific error Type rather than RuntimeException.

CanBerker:
LongType is commonly used in SparkSQL DataFrames for many common integers. This type directly maps to DecimalItem in Rumble. I opted for implementing the DecimalType to Item and vice versa conversions in this PR as it made a much cleaner solution compared to potential work-arounds.
- DecimalType maps to DecimalItem as well (just as LongType)
- DecimalItem maps back to DecimalType. 
- A static DecimalType of precision 30 and scale 15 is defined at the moment to prevent rounding of values (this can be tweaked or improved later on)
- In the context of SparkML, transformers and estimators are not fit to work with BigDecimal types. This makes DecimalType unwieldy to use. Using a flag ('forSparkML') schema and row generation for DataFrames has been configured to map decimalItems to DoubleType in the context of SparkML.

Annotate()'s schema now accepts objects and arrays in a nested manner.
ArrayItem and ObjectItem to row conversion is implemented.

Nested ArrayType in RowToItemParser was not working. This is also fixed.

Fixes #513
Fixes #484 
Fixes #517


src/main/java/org/rumbledb/api/Item.java
```diff
@@ -138,7 +139,7 @@ public Item compareItem(Item other, OperationalExpressionBase.Operator operator,
      * @return the list of the array members.
      */
     public List<Item> getItems() {
-        throw new RuntimeException("Item is not an array.");
+        throw new OurBadException("Item '" + this.serialize() + "' is not an array.");
```
ghislainfourny: After my other PR is merged, you could use this.getDynamicType().toString()

ghislainfourny:
@CanBerker thanks for doing this. It looks good overall, just a common on double vs. decimal.

The forSparkML flag feels "itchy" and is probably going to come back to bite us later.

How about:
1. always cleanly mapping decimal to DecimalType and back
2. and automatically promoting numbers to the expected number type (or throw an error if not possible) (i.e., implement your TODO)
?

That way, if a decimal item is annotated as double, then it is converted to a double. Also, this is the spirit of JSONiq to do so as type promotion is also done when calling functions.

Then we can document that decimal types should not be used when annotating features with ML estimators, and we can ask to use double types instead.

src/main/java/org/rumbledb/api/Item.java
CanBerker: For exactly which part of the error message? ie.  'is not a .... (getDynamicType)' ?

CanBerker:
I like your suggestion regarding the flag. It was a way of getting the code running & tests passing while sending the message in code that this is alternative mapping is performed solely for the SparkML part. 

I will implement your suggestions soon. It should make a very nice user experience :)

src/main/java/org/rumbledb/api/Item.java
ghislainfourny: Instead of serializing the whole item to shorten the message. But I leave this choice up to you.

CanBerker: I understand now. My goal was to provide as much clarity as possible to the user. Currently the error messaged look like this.
```
annotate(
     (
         {"id": 1, "age":  20, "weight": 68.8},
         {"id": 2, "age":  35, "weight": 72.4},
         {"id": 3, "age":  50, "weight": 76.3}
     ),
     {"id": "integer", "age": "string", "weight": "decimal"}
)
```

> ⚠️  ️Error [err: RBML0005]LINE:1:COLUMN:0:Schema error in annotate(); Error while applying the schema; Schema does not match the data;  Item '20' is not a string.

If the object causing the error is error message may be populated indeed. However, I believe this serialization could be indispensable in debugging type error oversights in a large datasets.

I was also thinking of mentioning the name of the column that the problematic data is located in. While this could be used as an alternative to the serialization of the problematic item, I still believe the serialization would be helpful in the scenerio described above.

CanBerker:
```
rumble$ annotate(
    (
        {"id": 1, "age":  20, "weight": 68.8},
        {"id": 2, "age":  35, "weight": 72.4},
        {"id": 3, "age":  50, "weight": 76.3}
    ),
    {"id": "integer", "age": "string", "weight": "decimal"}
)
```

> ⚠️  ️Error [err: RBML0005]LINE:1:COLUMN:0:Schema error in annotate(); Data does not fit to the given schema in field 'age';  Item '20' is not a string.\

The final look of the error messages. 
- Redundant re-throws removes
- field information for errors added
- column keyword replaced with field in error messages for consistency and hiding column concept from the user

ghislainfourny: Yes, it makes sense. Thanks!

src/main/java/sparksoniq/spark/ml/ApplyTransformerRuntimeIterator.java
```diff
@@ -97,6 +97,16 @@ public ApplyTransformerRuntimeIterator(
 
             return result;
         } catch (IllegalArgumentException e) {
+            if (e.getMessage().matches(".*DecimalType.*is not supported.*")) {
```
ghislainfourny: Note for the future: promotion is more general than mapping decimals to doubles (e.g., integer to decimal, anything to double). Maybe opening an issue to remember for later could be useful if you believe not all cases of the spec are covered.

src/main/resources/test_files/runtime-spark/RumbleML/AnnotateFunction/MLAnnotateForEstimator1.jq
```diff
@@ -7,7 +7,7 @@ let $local-data := (
     {"id": 4, "col1": 9.1, "col2": 9.1, "col3": 9.1},
     {"id": 5, "col1": 9.2, "col2": 9.2, "col3": 9.2}
 )
-let $df-data := annotate($local-data, {"id": "integer", "col1": "double", "col2": "double", "col3": "double"})
+let $df-data := annotate($local-data, {"id": "integer", "col1": "decimal", "col2": "decimal", "col3": "decimal"})
 
 let $est := get-estimator("KMeans")
```
CanBerker: @ghislainfourny, following our discussion today, the reason decimal types here work with KMeans here is the underlying VectorAssembler usage. At line15, these decimal columns are vectorized to form a featuresVector. Having a sparse vector of doubles, KMeans can operate on them. The debugging screenshot is given below:

image

To summarize,  I'm assuming most of the ML components do not support operations on decimalType. VectorAssembler however, is an exceptional transformer that is capable of operating on Decimals. We will note if we come across any more Decimal supporting transformers or estimators along the way.

In the current state, if a decimal supported exception is thrown upon calling a fit or a transform, this will be caught with regex and will throw an appropriate exception.

CanBerker:
@ghislainfourny I need your approval to merge this PR.

src/main/resources/test_files/runtime-spark/RumbleML/AnnotateFunction/MLAnnotateForEstimator1.jq
ghislainfourny: Perfect. Thank you for the clarification!

PR #519, opened 2020-03-10
Type expressions

ghislainfourny:
Continuing the flattening and simplification of the expression tree for other dependent projects.


src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -760,14 +760,9 @@ public RuntimeIterator visitNotExpr(NotExpression expression, RuntimeIterator ar
     public RuntimeIterator visitUnaryExpr(UnaryExpression expression, RuntimeIterator argument) {
         // compute +- final result
         int result = 1;
```
CanBerker: Is this variable still used?

ghislainfourny: Nice catch!

PR #522, opened 2020-03-11
Refactor3

ghislainfourny:


src/main/java/org/rumbledb/expressions/postfix/DynamicFunctionCallExpression.java
```diff
@@ -30,16 +31,21 @@
...
     ) {
-        super(mainExpression, metadata);
+        super(metadata);
+        if (mainExpression == null) {
+            throw new OurBadException("Main expression cannot be null in a postfix expression.");
+        }
+        this.mainExpression = mainExpression;
         this.arguments = arguments;
         if (this.arguments == null) {
             this.arguments = new ArrayList<>();
```
CanBerker: Can you take another look at the initHighestExecutionMode() method of this class. The override is probably redundant now as it doesn't inherit from PostFixExpression. But the explicit definition and comment may be helpful for clarity.

ghislainfourny: Good catch! It is removed.

PR #523, opened 2020-03-11
Paralellized head function

CodingKaiser:
Based on our discussion on Monday, and the existing code of this function, it would appear to me that this function is already "parallelized". As far as I understood it, since we are materializing a single item here, we can just use the LocalIterator, as it would make little sense to return an RDD or DF. Or is there a reason to use HybridIterator here?

I did nest the second part of the function in an else statement, as this was being executed even if the incoming collection was an RDD.


ghislainfourny:
Very good catch! Indeed it seems it was already implemented, but you found and fixed a nice bug. It would be great if you could double-check that there is a test to check the behavior, or add one otherwise:

head(parallelize(1 to 10000))

Thanks!

CodingKaiser:
Thanks you!
The test case would appear to already by implemented here:

./src/main/resources/test_files/runtime-spark/Pushdown/ObjectLookup3.jq

ghislainfourny:
> Thank you!
> The test case would appear to already be implemented here:
> 
> ./src/main/resources/test_files/runtime-spark/Pushdown/Head.jq
> 
> ```
> (:JIQS: ShouldRun; Output="1" :)
> head(parallelize(1 to 1000000))
> ```

Thank you for checking!

PR #526, opened 2020-03-11
Simplify itemType class and get rid of magic strings for types

### LINKED ISSUE #521: Annotate() fn: Refactor magic type names with ItemType.toString()

CanBerker:
Fixes #521 
- ItemType enumhas been made the single source of truth for itemtypes and their string representations (for both construction and serialization)
- The class has been converted to an enum to achieve referential equality that stays consistent even across deep copying of SequenceTypes. 
- ItemTypes enum is now redundant
- ItemTypes defined in Functions.java are cleaned up as they are redundant.
- Magic values for itemtype-strings in RumbleML are cleaned up
- RumbleML schema itemtype keywords are updated to match the grammar which is 'lowerCamelCase'
- OrderByClauseRuntimeIterator and UDFs are also cleaned up from the magic type strings.


src/main/java/org/rumbledb/items/FunctionItem.java
```diff
@@ -237,4 +238,19 @@ public void read(Kryo kryo, Input input) {
...
+            oos.writeObject(this);
+            oos.flush();
+            byte[] data = bos.toByteArray();
+            ByteArrayInputStream bis = new ByteArrayInputStream(data);
+            ObjectInputStream ois = new ObjectInputStream(bis);
+            return (FunctionItem) ois.readObject();
+        } catch (IOException | ClassNotFoundException e) {
+            throw new OurBadException("Error while deep copying the function body runtimeIterator");
+        }
+    }
```
CanBerker: Exracted to a method, cosmetic

src/main/java/org/rumbledb/runtime/functions/base/Functions.java
```diff
@@ -295,98 +295,66 @@
...
-        itemTypes.put("date", ItemType.dateItem);
-        itemTypes.put("time", ItemType.timeItem);
-
-        itemTypes.put("hexBinary", ItemType.hexBinaryItem);
-        itemTypes.put("base64Binary", ItemType.base64BinaryItem);
-
-        itemTypes.put("null", ItemType.nullItem);
-
-    }
-
```
CanBerker: These were redundant as we have static final instances of ItemTypes.

ghislainfourny:
@CanBerker thanks for the proposal -- let us discuss it at the next meeting.

While it is true that, in the current state, an enum does the trick, this will not be true in the future (user-defined types, function types). I am thus afraid that this refactoring may not be compatible with the future evolution of Rumble.

CanBerker:
Sounds good. Looking forward to it

ghislainfourny:
Thanks! By the way, some of the suggested changes in this PR that are orthogonal to making the class an enum and can probably be taken over, so this work was useful.

ghislainfourny:
@CanBerker would it be possible to just revert, in this PR, the conversion of the class to an enum, i.e., convert back the enum values to static constants for forward-compatibility? I think that this should still preserve many of your other (useful) changes. Then I will take a look again. Thanks!

CanBerker:
I was previously pushing to enforce types as singletons that do not get recreated at any part of the program. Enums were very helpful in ensuring this across serialization/deserialization which deep copying also uses. However, with the new implementation, even though i managed to get Java serialization to preserve the singleton nature, I believe kryo remained problematic.

Following these efforts, I have simply implemented an equals method that performs a name check  and dropped the effort into preserving the singleton pattern. As you have mentioned further changes with user defined types, this current implementation should hopefully be sufficient.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -819,7 +819,7 @@ public SequenceType processSequenceType(JsoniqParser.SequenceTypeContext ctx) {
         if (ctx.item == null) {
             return SequenceType.emptySequence;
         }
-        ItemType itemType = new ItemType(ctx.item.getText());
+        ItemType itemType = ItemType.getItemTypeByName(ctx.item.getText());
```
ghislainfourny: This looks good.

src/main/java/org/rumbledb/items/FunctionItem.java
```diff
@@ -237,4 +238,19 @@ public void read(Kryo kryo, Input input) {
     public ItemType getDynamicType() {
         return ItemType.functionItem;
     }
+
+    public FunctionItem deepCopy() {
```
ghislainfourny: This looks good too.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -292,51 +293,58 @@ private static void addValue(
...
+        if (itemTypeName.equals(ItemType.dateTimeItem.getName())) {
+            return DataTypes.TimestampType;
+        }
+        if (itemTypeName.equals(ItemType.hexBinaryItem.getName())) {
+            return DataTypes.BinaryType;
+        }
+        if (itemTypeName.equals("object")) {
+            return vectorType;
+        }
+        throw new IllegalArgumentException("Unexpected item type found: '" + itemTypeName + "'.");
```
ghislainfourny: This looks good

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -292,51 +293,58 @@ private static void addValue(
...
     }
 
     public static String getItemTypeNameFromDataFrameDataType(DataType dataType) {
         if (DataTypes.BooleanType.equals(dataType)) {
-            return "boolean";
+            return ItemType.booleanItem.getName();
```
ghislainfourny: Also looks good, and elses can all be removed.

src/main/java/org/rumbledb/runtime/flwor/clauses/OrderByClauseSparkIterator.java
```diff
@@ -40,6 +40,7 @@
 import org.rumbledb.runtime.flwor.expression.OrderByClauseAnnotatedChildIterator;
 import org.rumbledb.runtime.flwor.udfs.OrderClauseCreateColumnsUDF;
 import org.rumbledb.runtime.flwor.udfs.OrderClauseDetermineTypeUDF;
+import org.rumbledb.types.ItemType;
```
ghislainfourny: This file also looks good.

src/main/java/org/rumbledb/runtime/flwor/udfs/OrderClauseDetermineTypeUDF.java
```diff
@@ -165,7 +141,7 @@ public OrderClauseDetermineTypeUDF(
                         expressionWithIterator.getIterator().getMetadata()
                 );
             } else {
-                throw new OurBadException("Unexpected type found.");
+                this.result.add(this.nextItem.getDynamicType().getName());
```
ghislainfourny: Very nice!

src/main/java/org/rumbledb/runtime/functions/FunctionItemCallIterator.java
```diff
@@ -120,7 +120,7 @@ private void wrapArgumentIteratorsWithTypeCheckingIterators() {
                     TypePromotionIterator typePromotionIterator = new TypePromotionIterator(
                             this.functionArguments.get(i),
                             this.functionItem.getSignature().getParameterTypes().get(i),
-                            "Invalid argument for " + this.functionItem.getIdentifier().getName() + "function. ",
+                            "Invalid argument for " + this.functionItem.getIdentifier().getName() + " function. ",
```
ghislainfourny: good catch

src/main/java/org/rumbledb/runtime/functions/base/Functions.java
```diff
@@ -295,98 +289,66 @@
 
     private static List<FunctionIdentifier> currentUnsetUserDefinedFunctionIdentifiers;
 
-    private static final Map<String, ItemType> itemTypes;
```
ghislainfourny: Very nice

CanBerker:
Thanks!

PR #527, opened 2020-03-12
Flatten multiplication.

ghislainfourny:


src/main/java/org/rumbledb/expressions/arithmetic/MultiplicativeExpression.java
```diff
@@ -0,0 +1,106 @@
...
+            case "*":
+                return MultiplicativeOperator.MUL;
+            case "div":
+                return MultiplicativeOperator.DIV;
+            case "mod":
+                return MultiplicativeOperator.MOD;
+            case "idiv":
+                return MultiplicativeOperator.IDIV;
+        }
+        throw new OurBadException("Unrecognized multiplicative symbo: " + symbol);
```
CanBerker: Typo here in the error message "symbo(l): ...."

src/main/java/org/rumbledb/expressions/arithmetic/MultiplicativeExpression.java
```diff
@@ -0,0 +1,106 @@
...
+    public static MultiplicativeOperator getMultiplicativeOperatorFromSymbol(String symbol) {
+        switch (symbol) {
+            case "*":
+                return MultiplicativeOperator.MUL;
+            case "div":
+                return MultiplicativeOperator.DIV;
+            case "mod":
+                return MultiplicativeOperator.MOD;
+            case "idiv":
+                return MultiplicativeOperator.IDIV;
```
CanBerker: I believe these strings are almost never expected to change so this idea is not critical here. But I believe, especially for other places where similar enums or constant strings might turn up,  that it might make a more robust implementation if we referenced the name from Multiplicative operator enums rather than using magic strings as the case values. It would make it much easier to refactor as well as reduce the possibility of introducing bugs due to string mismatches, typos and so on

ghislainfourny: Also very good point.

src/main/java/org/rumbledb/expressions/arithmetic/MultiplicativeExpression.java
ghislainfourny: Good catch.

src/main/java/org/rumbledb/expressions/arithmetic/MultiplicativeExpression.java
```diff
@@ -0,0 +1,108 @@
...
+            if (symbol.contentEquals(MUL.toString())) {
+                return MUL;
+            }
+            if (symbol.contentEquals(DIV.toString())) {
+                return DIV;
+            }
+            if (symbol.contentEquals(MOD.toString())) {
+                return MOD;
+            }
+            if (symbol.contentEquals(IDIV.toString())) {
```
CanBerker: Very interesting choice for the string comparison method. I was not aware such an alternative to the equals() method existed. I have looked into the comparison of the two, I would like to hear your motivation for using it in this context.

src/main/java/org/rumbledb/runtime/operational/MultiplicativeOperationIterator.java
```diff
@@ -98,7 +107,7 @@ public Item next() {
             } catch (RuntimeException e) {
                 throw new UnexpectedTypeException(
                         " \""
-                            + this.operator.name().toLowerCase()
+                            + this.multiplicativeOperator.toString().toLowerCase()
```
CanBerker: I don't think toLowerCase() call here is needed anymore.

src/main/java/org/rumbledb/expressions/arithmetic/MultiplicativeExpression.java
ghislainfourny: This was unintentional (autocomplete). I have reverted.

ghislainfourny:
Thanks!

PR #528, opened 2020-03-12
Fix comparison.

ghislainfourny:
Now the expression hierarchy is fully flat.

This should also bring a performance improvement with less virtual method calls.


src/main/java/org/rumbledb/expressions/comparison/ComparisonExpression.java
```diff
@@ -0,0 +1,152 @@
...
+                this == VC_NE
+                ||
+                this == VC_LT
+                ||
+                this == VC_LE
+                ||
+                this == VC_GT
+                ||
+                this == VC_GE;
+        }
```
CanBerker: How about introducing isValueComparison as a property on the enums themselves ? Just like a name. I think It would take up more memory, but would reduce the needed computation with equality checks.

src/main/java/org/rumbledb/runtime/primary/StringRuntimeIterator.java
```diff
@@ -20,7 +20,7 @@
 
 package org.rumbledb.runtime.primary;
 
-import org.apache.commons.lang3.StringEscapeUtils;
```
CanBerker: I wanted to point out that these imports have been updated in ObjectItem ArrayItem and here. I wasn't sure if this was performed by you or the IDE. I doubt this change affects anything

ghislainfourny: That's on purpose. It fixes a warning that has been there for a long type: the old package was deprecated.

PR #529, opened 2020-03-12
Parallelized tail function

CodingKaiser:


src/main/resources/test_files/runtime/FunctionSequence/General/FunctionTail2.jq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="99" :)
+count(tail(parallelize(1 to 100))),
+tail(parallelize(1 to 1)),
+tail(parallelize(()))
```
ghislainfourny: to be extra sure I would test that tail(parallelize(1 to 5)) gives 2 3 4 5

CodingKaiser:
Ah yes, that makes sense. I attempted to come up with a solution using `mapPartitionsWithIndex`, but unfortunately I ran into an error in the case where the sequence is of length 1. Thank you for the pointer on `zipWithIndex`. Will try this now and add the additional tests.

CodingKaiser:
I made the corresponding changes. The tests are all passing now.

PR #532, opened 2020-03-12
Fix rdd to DF conversion for arrays, add tests and data

CanBerker:


src/main/java/sparksoniq/spark/DataFrameUtils.java
```diff
@@ -102,7 +102,7 @@ private static StructField generateStructFieldFromNameAndItem(String columnName,
     private static DataType generateDataTypeFromItem(Item item) {
         if (item.isArray()) {
             validateArrayItemInSchema(item);
-            Item arrayContentsTypeItem = item.getItems().get(0);
+            Item arrayContentsTypeItem = item.getItemAt(0);
```
ghislainfourny: Nice catch.

src/main/resources/queries/rumbleML/sample-ml-data-nested2.json
```diff
@@ -0,0 +1,6 @@
+{"label": 0, "object": {"intarray": [1,2,3], "stringarray": ["abc"]}, "array": [{"string": "a", "decimal": "3.5"}]}
+{"label": 1, "object": {"intarray": [4,5,6], "stringarray": ["abc"]}, "array": [{"string": "a", "decimal": "3.5"}]}
+{"label": 2, "object": {"intarray": [7,8,9], "stringarray": ["abc"]}, "array": [{"string": "a", "decimal": "3.5"}]}
+{"label": 3, "object": {"intarray": [1,4,7], "stringarray": ["abc"]}, "array": [{"string": "a", "decimal": "3.5"}]}
+{"label": 4, "object": {"intarray": [2,5,8], "stringarray": ["abc"]}, "array": [{"string": "a", "decimal": "3.5"}]}
+{"label": 5, "object": {"intarray": [3,6,9], "stringarray": ["abc"]}, "array": [{"string": "a", "decimal": "3.5"}]}
```
ghislainfourny: How about also varying the size of the arrays rather than keep them constant? (In my understanding, this should work as well in dataframes, as array sizes are open).

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -386,12 +386,14 @@ private static Object getRowColumnFromItemUsingSchemaField(Item item, StructFiel
...
+                Object[] arrayItemsForRow = new Object[arrayItems.size()];
                 DataType elementType = ((ArrayType) dataType).elementType();
-                for (Item arrayItem : item.getItems()) {
-                    arrayContents.add(getRowColumnFromItemUsingDataType(arrayItem, elementType));
+                for (int i = 0; i < arrayItems.size(); i++) {
+                    Item arrayItem = item.getItemAt(i);
+                    arrayItemsForRow[i] = getRowColumnFromItemUsingDataType(arrayItem, elementType);
                 }
-                return arrayContents;
+                return arrayItemsForRow;
```
ghislainfourny: Is my understand correct that the issue was that a list was returned, but it should be a java array? Thanks!

CanBerker: Precisely

src/main/resources/queries/rumbleML/sample-ml-data-nested2.json
CanBerker: Very good suggestion! Added a new test for that

PR #538, opened 2020-03-14
Improvements with spotless and basedir usage in pom.xml

CanBerker:
According maven documentation [here](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Available_Variables) "project.basedir" should be used as alternatives such as "basedir" or "pom.basedir" are now deprecated.

Spotless configuration for file inclusion and exclusion is added as described [here](https://github.com/diffplug/spotless/tree/master/plugin-maven#file-includes-and-excludes). These relative paths are unfortunately not prefixed with a form of "basedir". This is because no form of maven basedir specification worked with these. They do however function properly in the current state with relative paths.

Added spotless error for unused imports.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
                     <source>1.8</source>
                     <target>1.8</target>
                     <compilerArguments>
-                        <properties>${basedir}/org.eclipse.jdt.core.prefs</properties>
+                        <properties>${project.basedir}/org.eclipse.jdt.core.prefs</properties>
                     </compilerArguments>
                 </configuration>
                 <dependencies>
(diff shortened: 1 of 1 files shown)


ghislainfourny:
> Added spotless error for unused imports.

I cannot see this -- could it be not added?

pom.xml
```diff
@@ -179,9 +179,18 @@
...
+                        </includes>
+
+                        <excludes>
+                            <exclude>src/main/java/org/rumbledb/parser/*</exclude>
+                        </excludes>
                         <eclipse>
-                            <file>${basedir}/spotless-formatter-eclipse-jdt-configurations.xml</file>
+                            <file>${project.basedir}/spotless-formatter-eclipse-jdt-configurations.xml</file>
                         </eclipse>
+                        <removeUnusedImports/>
```
CanBerker: This line handles the unused imports. If any unused import is found in the code and spotless:check is applied, it throws errors (I believe spotless:apply should remove unused imports but I haven't double checked)

pom.xml
```diff
@@ -179,9 +179,18 @@
                 </executions>
                 <configuration>
                     <java>
+                        <includes>
+                            <!-- include all java files in "java" folders under "src" -->
+                            <include>src/**/java/**/*.java</include>
+                        </includes>
+
+                        <excludes>
+                            <exclude>src/main/java/org/rumbledb/parser/*</exclude>
```
CanBerker: To answer your question on the 2nd point of the description. Yes includes and excludes statements work currently with spotless. However if we try to add any sort of ${basedir}/${project.basedir} etc. to the beginning of these relative paths, they stop working. 

These basedir variables work in other places of the pom.xml, just not in these include/exclude configs of spotless. I don't know the exact reason for this.

PR #547, opened 2020-03-15
Fix SparkExceptions disappearing in Shell error handling

CanBerker:


CanBerker:
Great suggestion! I have applied the fix there as well.

PR #555, opened 2020-03-16
Fix static context multipass for undefined function calls stored in variables

### LINKED ISSUE #553: Execution Mode access error on function call typo
parallellize function does not exist(this is a typo from parallelize). There is probably some intricate bug with function calls and variable's storing rdds.

```
let $raw-data := parallellize((
    {"id": 0, "category": "a", "categoryIndex": 1},
    {"id": 1, "category": "b", "categoryIndex": 2},
    {"id": 2, "category": "c", "categoryIndex": 3},
    {"id": 3, "category": "a", "categoryIndex": 1},
    {"id": 4, "category": "a", "categoryIndex": 1},
    {"id": 5, "category": "c", "categoryIndex": 3}
))

return $raw-data
```
>⚠️  ️Error [err: RBST0004 ] An execution mode is accessed without being set.

CanBerker:
Fixes #553


ghislainfourny:
@CanBerker many thanks for fixing this.

It looks good, except I have concerns on transforming the boolean method parameter (to deactivate errors) into a static variable. In general, static variables for states are not a recommended software engineering practice -- consider for example that somebody uses the compiler on several trees in parallel: this would break. Does it make sense?

If you just revert the static variable change and restore these flags to a boolean parameter (or make it several boolean parameters if this was something you needed for fixing this bug), I will approve.

CanBerker:
I understand your concerns regarding the static variables. Thanks a lot for informing me of the best practices alongside an enlightening example with the several trees. It was very helpful in understanding where this might be problematic.

I initially opted this for approach to make the refactoring as small as possible. I'm afraid the alternative approach requires that we propagate these "suppressErrors" flags across all executionMode getters for each expression.

Reason being, in the early passes of StaticContextVisitor:
parallellize -> includes typo -> unset functionCall
let $a := parallellize(...) -> unset FLWOR variable (currently the source of error)
non-local expressions using $a -> unset ...

So we need to propagate through all expressions, allowing unset variables by suppressing errors.

ghislainfourny:
Thanks, @CanBerker !

If you want to do it even cleaner, you can pass the flags as an int parameter to all these methods, and have a small util class that queries the flags from the int. That way, it will make adding and removing flag easier in the future, too.

src/main/java/org/rumbledb/expressions/comparison/ComparisonExpression.java
```diff
@@ -56,10 +56,10 @@
                 case "lt":
                 case "ge":
                 case "gt":
-                    isValueComparison = true;
+                    this.isValueComparison = true;
```
ghislainfourny: good catch

PR #556, opened 2020-03-16
Cosmetic improvements to Functions.java

CanBerker:
Only cosmetic changes to functions.java

When I look at the BuiltinFunctions Catalog, we have bunch of static final class objects. It looks like defining an Enum type for these would be fitting and it would simplify the code. I was thinking of creating an issue for this to address it later. Do you have any reservations against this idea @ghislainfourny?


CanBerker:
Issue for enums created at #572

Issue #544, opened 2020-03-15
RumbleML: Missing support for vectors in parameters other than featuresCol


CanBerker:
```
let $data := annotate(
    json-file("./src/main/resources/queries/rumbleML/sample-ml-data-flat.json"),
    { "label": "integer", "binaryLabel": "integer", "name": "string", "age": "double", "weight": "double", "booleanCol": "boolean", "nullCol": "null", "stringCol": "string", "stringArrayCol": ["string"], "intArrayCol": ["integer"],  "doubleArrayCol": ["double"],  "doubleArrayArrayCol": [["double"]] }
)

let $est := get-estimator("BucketedRandomProjectionLSH")
let $tra := $est(
    $data,
    { "inputCol": "weight" }
)
for $result in $tra(
    $data,
    { }
)
return {
    "label": $result.label,
    "name": $result.name,
    "age": $result.age,
    "weight": $result.weight,
    "result": $result
}
```

>⚠️  ️Error [err: RBML0003]LINE:6:COLUMN:12:Invalid Param; Parameter provided to BucketedRandomProjectionLSH causes the following error: requirement failed: Column weight must be of type struct<type:tinyint,size:int,indices:array<int>,values:array<double>> but was actually double.

[BucketedRandomProjectionLSH](https://spark.apache.org/docs/2.4.4/api/java/org/apache/spark/ml/feature/BucketedRandomProjectionLSH.html) requires a vector in its inputCol. Currently we only support vectors with an internal trick on the featuresCol

PR #561, opened 2020-03-17
Implement vectorization for inputCol in specific Estimator and transformers

CanBerker:
Fixes most of #544.

We were lacking support on a number of estimators and transformers due to our limited support of vectors. I have extended this support to necessary estimator and transformers with a somewhat quick & dirty solution. I have opted for this as I will be focusing on the report again soon. I would like to note that this brought our coverage of sparkML from %60-70s to over %90. If you see some low hanging fruits we can try to address these quickly.

Issue #544 points out 2 more expected vectors for params.
- Interaction transformer expects inputCols vector(s)
- ElementwiseProduct expects scalingVec as vector.

If you find the current approach acceptable @ghislainfourny, then I will make the necessary adjustments for these two params as well in a separate PR.


src/main/java/sparksoniq/spark/ml/ApplyTransformerRuntimeIterator.java
```diff
@@ -46,43 +44,74 @@ public ApplyTransformerRuntimeIterator(
...
-        boolean isFeaturesColumnGenerated = false;
-
-        if (transformerExpectsFeaturesColParam) {
-            Object featuresColValue = new String[] { featuresColParamDefaultValue };
-            if (paramMapItem.getItemByKey(featuresColParamName) != null) {
-                RumbleMLCatalog.validateTransformerParameterByName(
+        // update input dataset and paramMapItem based on the needs of special params
+        for (String specialParamName : RumbleMLCatalog.specialParams) {
+            if (
+                !RumbleMLCatalog.getTransformerParams(this.transformerShortName, getMetadata())
```
CanBerker: The code on estimator and transformer sides are %99 identical with only changes being the called getter methods from the RumbleMLCatalog. When the system has matured, I believe that estimators and transformers can be merged together to have a unified API with shared maps and getter methods. However, I deliberately did not take this approach in the initial implementation as before said maturity is reached, I wanted to keep the code flexible enough to accommodate unforeseen peculiarities of Estimators and transformers.

PR #563, opened 2020-03-17
Implement root reader.

ghislainfourny:
This introduces:
- not overriding the chosen serializer if the user sets it on the CLI
- root-file(), which only works if --conf spark.serializer=org.apache.spark.serializer.JavaSerializer is set on the CLI (does not work with Kryo -> thus no tests)
- pushdown of object lookup to DataFrames
- tests for object lookup (also work with root-file(), massively more efficient)


src/main/java/sparksoniq/spark/SparkSessionManager.java
```diff
@@ -100,24 +101,30 @@ private void initialize() {
...
-            FlworTuple.class,
-            FlworKey.class,
-            RuntimeIterator.class,
-            RuntimeTupleIterator.class };
-
-        this.configuration.registerKryoClasses(serializedClasses);
+        String serializer = null;
+        try {
+            serializer = this.configuration.get("spark.serializer");
+        } catch (NoSuchElementException e) {
```
CanBerker: Couldn't we write this without resorting to exception catching?

CanBerker:
I have a suggestion regarding tests for rootFile. In RuntimeTests.java we are configuring spark in setupSparkSession(). One potential approach could be to update the spark.serializer config for certain tests (we would have to double check whether spark allows updating configs once session is instantiated) but I think this would get messy.

A potentially better approach could be to define a separate TestRunner with the new serializer that you desire (ie. a 3rd test group alogisde RuntimeTests and SparkRuntimeTests, such as JavaSerializerTests or something). I think this should work as you can define the tests for RootFile with the desired Java configuration.

If you decide that you want to proceed without tests for this function, I can give my approval on the PR.

src/main/java/sparksoniq/spark/SparkSessionManager.java
ghislainfourny: Yes, very good point, fixed now.

ghislainfourny:
> I have a suggestion regarding tests for rootFile. In RuntimeTests.java we are configuring spark in setupSparkSession(). One potential approach could be to update the spark.serializer config for certain tests (we would have to double check whether spark allows updating configs once session is instantiated) but I think this would get messy.
> 
> A potentially better approach could be to define a separate TestRunner with the new serializer that you desire (ie. a 3rd test group alogisde RuntimeTests and SparkRuntimeTests, such as JavaSerializerTests or something). I think this should work as you can define the tests for RootFile with the desired Java configuration.
> 
> If you decide that you want to proceed without tests for this function, I can give my approval on the PR.

For the moment we are trying to keep it light, as we are still experimenting with root. Also, we expect that the underlying parser should at some point support Kryo (the issue is raised). So no tests for now. Thanks!

PR #569, opened 2020-03-19
Simplify for clause.

ghislainfourny:
According to the spec

https://www.w3.org/TR/xquery-30/#id-xquery-for-clause

on the logical level, every for clause only has one variable. This helps us flatten the expression tree further.

Also, a few bugs were fixed on the way.


src/main/java/org/rumbledb/expressions/flowr/ForClause.java
```diff
@@ -22,58 +22,106 @@
...
+
+        this.variableHighestStorageMode = ExecutionMode.LOCAL;
+    }
+
+    public ExecutionMode getVariableHighestStorageMode(VisitorConfig visitorConfig) {
+        if (
+            !visitorConfig.suppressErrorsForAccessingUnsetExecutionModes()
+                && this.variableHighestStorageMode == ExecutionMode.UNSET
+        ) {
+            throw new OurBadException("An variable storage mode is accessed without being set.");
```
CanBerker: A* variable storage mode

src/main/java/org/rumbledb/expressions/flowr/ForClause.java
```diff
@@ -22,58 +22,106 @@
...
+        );
+        this.expression = expression;
+
+    }
+
+    public String getVariableName() {
+        return this.variableName;
+    }
+
+    public boolean isAlowsEmpty() {
```
CanBerker: isAl(l)owsEmpty

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -266,9 +277,22 @@ public Node visitFlowrExpr(JsoniqParser.FlowrExprContext ctx) {
...
+
+            // there might have been multiple variables in the syntactic starting clause
+            // and each one is mapped to a separate start clause on the logical level, so we need to rewind
+            // and populate content clauses appropriately
+            clausesTemp.clear();
+            Clause lastChildClause = childClause;
+            while (childClause.getPreviousClause() != null) {
+                clausesTemp.add(childClause);
+                childClause = childClause.getPreviousClause();
+            }
```
CanBerker: The code here seems duplicated from above. Would it be possible to extract it to a method which I am hoping would increase readability as well. Also the comments seem to be identical, is this intentional or should some editing take place there? Lastly, I believe multiple variables can happen in any for and let variables and not only in the starting clauses. I was worrying if the current statement can be misleading.

ghislainfourny: This is because it is only temporary. This code is already removed in the next PR (let clause)

src/main/java/org/rumbledb/expressions/flowr/ForClause.java
ghislainfourny: Thanks, this is fixed.

src/main/java/org/rumbledb/expressions/flowr/ForClause.java
ghislainfourny: Thanks, this is fixed.

PR #570, opened 2020-03-19
Implemented parallelized index-of function

CodingKaiser:


src/main/resources/test_files/runtime/FunctionSequence/Value/FunctionIndexOf8.jq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="(2, 5, 7)" :)
+index-of(parallelize((10, 2e+1, 3e+1, 30, 20, 10, 20.0)), 20)
+
```
ghislainfourny: Very nice!

PR #573, opened 2020-03-20
Let clause

ghislainfourny:


src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -212,61 +210,62 @@ public RuntimeIterator visitFlowrExpression(FlworExpression expression, RuntimeI
...
     private RuntimeTupleIterator visitFlowrClause(
             Clause clause,
-            RuntimeIterator argument,
-            RuntimeTupleIterator previousIterator
+            RuntimeIterator argument
     ) {
+        RuntimeTupleIterator previousIterator = null;
+        if (clause.getPreviousClause() != null) {
+            previousIterator = this.visitFlowrClause(clause.getPreviousClause(), argument);
+        }
```
CanBerker: Very nice

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -127,39 +127,58 @@ public StaticContext visitFunctionCall(FunctionCallExpression expression, Static
...
-        this.visit(expression.getExpression(), argument);
-        expression.initHighestExecutionAndVariableHighestStorageModes(this.visitorConfig);
-        return visitFlowrVarDeclaration(expression, argument);
+    public StaticContext visitLetClause(LetClause clause, StaticContext argument) {
+        // TODO visit at...
+        this.visit(clause.getExpression(), argument);
+        clause.initHighestExecutionMode(this.visitorConfig);
+
+        StaticContext result = new StaticContext(argument);
+        // TODO for now we only support as/default, no inference, flags
```
CanBerker: Can we extract these TODOs to issues?

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -234,111 +232,110 @@ public Node visitExprSingle(JsoniqParser.ExprSingleContext ctx) {
     // TODO [EXPRVISITOR] count
```
CanBerker: Same with the TODO here

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -234,111 +232,110 @@ public Node visitExprSingle(JsoniqParser.ExprSingleContext ctx) {
...
             }
-            childClause.setPreviousClause(previousFLWORClause);
-            previousFLWORClause = childClause;
-            contentClauses.add(childClause);
+
+            previousFLWORClause.chainWith(clause.getFirstClause());
+            previousFLWORClause = clause.getLastClause();
         }
 
         // visit return
```
CanBerker: This comment can be probably removed

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -234,111 +232,110 @@ public Node visitExprSingle(JsoniqParser.ExprSingleContext ctx) {
...
-        ForClauseVar child;
+        ForClause clause = null;
         for (JsoniqParser.ForVarContext var : ctx.vars) {
-            child = (ForClauseVar) this.visitForVar(var);
-            vars.add(child);
+            ForClause newClause = (ForClause) this.visitForVar(var);
+            if (clause != null) {
+                clause.chainWith(newClause.getFirstClause());
+            }
+            clause = (ForClause) newClause.getLastClause();
```
CanBerker: Are the getFirstClause and getLastClause calls needed here? Isn't the result of visitForVar a single clause that is non-chained? I see that getFirstClause and getLastClause calls are idempotent when the clause chain consists of only 1 function so this is a non-issue from a correctness perspective. Yet, I was curious if there is some underlying chaining with variable definitions that I am missing.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -234,111 +232,110 @@ public Node visitExprSingle(JsoniqParser.ExprSingleContext ctx) {
...
-        LetClauseVar child;
+        LetClause clause = null;
         for (JsoniqParser.LetVarContext var : ctx.vars) {
-            child = (LetClauseVar) this.visitLetVar(var);
-            vars.add(child);
+            LetClause newClause = (LetClause) this.visitLetVar(var);
+            if (clause != null) {
+                clause.chainWith(newClause.getFirstClause());
+            }
+            clause = (LetClause) newClause.getLastClause();
```
CanBerker: Same comment as the forClause regarding method calls

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -127,39 +127,58 @@ public StaticContext visitFunctionCall(FunctionCallExpression expression, Static
...
-        for (Clause clause : expression.getContentClauses()) {
+        Clause clause = expression.getReturnClause().getFirstClause();
+        StaticContext result = this.visit(clause, argument);
+        while (clause != null) {
             result = this.visit(clause, result);
+            clause = clause.getNextClause();
         }
-
-        result = this.visit(expression.getReturnClause(), result);
         return result;
```
CanBerker: I have a feeling that returning 'argument' here instead of result would be more correct. I think return of this function means going out of scope for the FLWOR clause which means all the variables defined in there should go out of scope. Hence the 'argument' should be returned instead of 'result' to facilitate this.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: I think this was an overlook, there is nothing else than as for let clauses.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: I think that was also a leftover. The cloud clause is now supported.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Done

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Very good catch! I simplified.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Very good catch! I simplified.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: Very good catch too. Now it is easier to see. Let's see if tests still pass.

ghislainfourny: I also did the same change for other visit calls doing the same mistake. It doesn't change anything anyway: results from visitABC() on expressions are always discarded, it is only used for clauses. But it is more correct.

CanBerker:
Dear @ghislainfourny could you fix these unqualified accesses as well?

[WARNING] /Users/can/Projects/rumble-project/rumble/src/main/java/org/rumbledb/expressions/flowr/LetClause.java:[100,3658] Unqualified access to the field LetClause.expression
[WARNING] /Users/can/Projects/rumble-project/rumble/src/main/java/org/rumbledb/expressions/flowr/FlworExpression.java:[72,2552] Unqualified access to the field FlworExpression.returnClause
[WARNING] /Users/can/Projects/rumble-project/rumble/src/main/java/org/rumbledb/expressions/flowr/ForClause.java:[119,4246] Unqualified access to the field ForClause.expression

ghislainfourny:
@CanBerker thanks a lot for the thorough comments. Back to you.

PR #575, opened 2020-03-20
Support array unboxing for dataframes.

ghislainfourny:


src/main/java/org/rumbledb/runtime/postfix/ArrayUnboxingIterator.java
```diff
@@ -114,4 +121,35 @@ private void setNextResult() {
...
+                    .sql(
+                        String.format(
+                            "SELECT explode(`%s`) AS `%s` FROM array",
+                            SparkSessionManager.atomicJSONiqItemColumnName,
+                            SparkSessionManager.atomicJSONiqItemColumnName
+                        )
+                    );
+            }
+        }
+        return childDataFrame.sparkSession().sql("SELECT * FROM array WHERE false");
```
CanBerker: This line to return an empty result will be used for many similar function calls I think. Would it make sense to have a static empty DataFrame, and return this from all similar places instead of making this sql() call? I feel that this could go a long way for readability.

ghislainfourny: Very good catch! This is fixed.

CanBerker: Looks very clean now. Thanks

PR #576, opened 2020-03-20
Add support for array lookup in DF.

ghislainfourny:


src/main/resources/test_files/runtime-spark/Pushdown/ArrayLookup7.jq
```diff
@@ -0,0 +1,2 @@
+(:JIQS: ShouldRun; Output="(1, 2)" :)
+annotate(({ "foo" : [] } , { "foo" : [ 1 ] }, { "foo" : [ 2, 3 ] } ), { "foo" : [ "integer" ] }).foo[[1]]
```
CanBerker: Nice thinking with the use of annotate here

src/main/java/org/rumbledb/runtime/postfix/ArrayLookupIterator.java
```diff
@@ -150,4 +157,36 @@ public void setNextResult() {
...
+                            SparkSessionManager.atomicJSONiqItemColumnName,
+                            Integer.toString(this.lookup - 1),
+                            SparkSessionManager.atomicJSONiqItemColumnName,
+                            SparkSessionManager.atomicJSONiqItemColumnName,
+                            Integer.toString(this.lookup)
+                        )
+                    );
+            }
+        }
+        return childDataFrame.sparkSession().sql("SELECT * FROM array WHERE false");
```
CanBerker: Same comment as other PR, I'm waiting on this discussion before approving both

PR #577, opened 2020-03-22
Parallelization of remove-keys function

ioanas96:


src/main/java/org/rumbledb/runtime/functions/object/ObjectRemoveKeysFunctionIterator.java
```diff
@@ -123,4 +122,62 @@ private Item removeKeys(Item objItem, List<String> removalKeys) {
...
+    @Override
+    protected boolean hasNextLocal() {
+        return this.hasNext;
+    }
+
+    @Override
+    protected void resetLocal(DynamicContext context) {
+        this.iterator.open(this.currentDynamicContextForLocalExecution);
+        List<Item> removalKeys = this.children.get(1).materialize(this.currentDynamicContextForLocalExecution);
+        if (removalKeys.isEmpty()) {
```
ghislainfourny: This code is duplicate, could we transform this to a method call called at both places?

ioanas96: Done.

PR #579, opened 2020-03-23
Branch research project 2 staging

ghislainfourny:


ghislainfourny:
@CanBerker the code has all been reviewed already, so the only step needed is to clean up the automatically generated ANTLR files.

PR #581, opened 2020-03-23
Increase Laurelin version to fix cache issue in root-file(),

ghislainfourny:
When reading a root-file() multiple times in the shell, the first access was cached, and some unused fields are missing in subsequent calls. This was fixed in version 1.0.1 of Laurelin.


ghislainfourny:
@CanBerker would you like to take the review over?

PR #582, opened 2020-03-23
Support DF for subsequence.

ghislainfourny:


src/main/java/org/rumbledb/runtime/functions/sequences/general/SubsequenceFunctionIterator.java
```diff
@@ -79,6 +86,53 @@ public SubsequenceFunctionIterator(
...
+        df.createOrReplaceTempView("input");
+        df = df.sparkSession()
+            .sql(
+                String.format(
+                    "SELECT * FROM input LIMIT %s",
+                    Integer.toString(this.startPosition + this.length - 1)
+                )
+            );
+
+        df = FlworDataFrameUtils.zipWithIndex(df, 1L, "foo");
```
CanBerker: "foo" is very likely to collide with a user column. Can we use a UUID for the column name here?

ghislainfourny: Ah, yes! That was the plan and I forgot. Thanks! Very good eyes.

CanBerker:
This is optional but shall we also add tests for the behavior at the edge values ? Lower than first index, bigger than last index etc.

PR #586, opened 2020-03-25
Simplification and cleanup of quantified variables

ghislainfourny:


ghislainfourny:
@CanBerker the test queue works again :)

src/main/java/org/rumbledb/expressions/quantifiers/QuantifiedExpression.java
```diff
@@ -80,10 +80,18 @@ public QuantifiedOperators getOperator() {
...
+        for (QuantifiedExpressionVar v : this.variables) {
+            sb.append(separator);
+            sb.append(v.toString());
+            separator = ", ";
+        }
+        return sb.toString();
     }
 
-    public enum QuantifiedOperators {
+    public enum Quantification {
```
CanBerker: This is absolutely a non-issue. I just wanted to bring something I noticed to your attention. The name 'quantifier' is chosen for variables (for example in the constructor) while the enum is called Quantification. How do you feel about the name 'Quantifier' for the enum?

src/main/java/org/rumbledb/expressions/quantifiers/QuantifiedExpressionVar.java
```diff
@@ -20,29 +20,20 @@
...
 import org.rumbledb.expressions.Expression;
-import org.rumbledb.expressions.Node;
-import org.rumbledb.expressions.primary.VariableReferenceExpression;
 import org.rumbledb.types.SequenceType;
 
-import java.util.ArrayList;
-import java.util.List;
-
-public class QuantifiedExpressionVar extends Node {
-    private final VariableReferenceExpression variableReference;
```
CanBerker: Very nice. I really like seeing this redundancy cleaned up

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -1036,18 +1036,18 @@ public Node visitTypeSwitchExpr(JsoniqParser.TypeSwitchExprContext ctx) {
     @Override
     public Node visitQuantifiedExpr(JsoniqParser.QuantifiedExprContext ctx) {
         List<QuantifiedExpressionVar> vars = new ArrayList<>();
-        QuantifiedExpression.QuantifiedOperators operator;
+        QuantifiedExpression.Quantification operator;
         Expression expression = (Expression) this.visitExprSingle(ctx.exprSingle());
         if (ctx.ev == null) {
```
CanBerker: This is outside the scope of this PR but I just wanted to suggest that we also work on the names for variables used in the grammar at some point in the future.

ghislainfourny: Yes, absolutely. With time we will keep cleaning up. Thank you for mentioning it.

PR #588, opened 2020-03-25
Fixed bug in anyURI related to kryo read/write

CodingKaiser:


CodingKaiser:
Is there a particular reason none of the checks are being run as of late?

ghislainfourny:
@CanBerker it would be great if you could take a look at Travis and see why it is stuck

PR #590, opened 2020-03-26
Prevent silent failing of unhandled log paths

CanBerker:
The query given in the example below would never create an execution time log. The reason is the given log paths fits none of the expected patterns which are: 
- "file://"
- "hdfs://"
- "./"
```
noglob spark-submit --master local[*] --conf "spark.driver.host"="127.0.0.1" 
spark-rumble-1.4-jar-with-dependencies.jar --query-path my-query.jq 
--output-path my-output.json --log-path my-log.txt
```

This PR adds an OurBadException so that this does not fail silently for the future. The path given in the example above still needs to be handled. Should I create an issue for this or would you like to take this PR over @ghislainfourny ?


CanBerker:
Sounds good, do you require any action from me on the Travis issue or do we just wait?

ghislainfourny:
@CanBerker I think you need a round of spotless

PR #595, opened 2020-03-26
External variable cli

ghislainfourny:


ghislainfourny:
@CanBerker this PR introduces basic support for passing external variables from the CLI. It would be great if you could apply IntelliJ to the changed parser files. Thanks a lot!

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -57,10 +57,10 @@
 
 public class JsoniqQueryExecutor {
     public static final String TEMP_QUERY_FILE_NAME = "Temp_Query";
-    private SparksoniqRuntimeConfiguration configuration;
```
CanBerker: I'm glad to see this rename

CanBerker:
#597 is created to format generated files. Feel free to merge it if it looks good.

ghislainfourny:
@CanBerker has the generated files with wildcard removal and spotless applied. The diff is big, but if we agree to stick to this convention, diffs may be small again in the future.

src/main/java/org/rumbledb/compiler/DynamicContextVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+import org.rumbledb.expressions.AbstractNodeVisitor;
+import org.rumbledb.expressions.Node;
+import org.rumbledb.expressions.module.VariableDeclaration;
+import org.rumbledb.items.ItemFactory;
+import org.rumbledb.types.SequenceType.Arity;
+
+import sparksoniq.semantics.DynamicContext;
+
+/**
+ * Static context visitor implements a multi-pass algorithm that enables function hoisting
```
CanBerker: This comment should be removed. Feel free to add a documentation for the DynamicContextVisitor's purpose as well.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -288,4 +289,18 @@ public StaticContext visitTypeSwitchExpression(TypeSwitchExpression expression,
     }
     // endregion
 
+    // region FLWOR vars
```
CanBerker: I think this region specifier may have been accidentally copied.

src/main/java/org/rumbledb/compiler/DynamicContextVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+        List<Item> values = new ArrayList<>();
+        if (value != null) {
+            Item item = ItemFactory.getInstance().createStringItem(value);
+            values.add(item);
+            if (
+                variableDeclaration.getSequenceType().isEmptySequence()
+                    || !item.isTypeOf(variableDeclaration.getSequenceType().getItemType())
+            ) {
+                throw new UnexpectedTypeException(
+                        "External variable value is empty and does not match the expected type.",
```
CanBerker: It seems to me that the phrasing of the error message (empty 'and' type mistmatch) does not match the if condition (empty 'or' type mismatch) . Is this intentional?

src/main/java/org/rumbledb/compiler/DynamicContextVisitor.java
ghislainfourny: Done

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: Done

src/main/java/org/rumbledb/compiler/DynamicContextVisitor.java
ghislainfourny: Done

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
ghislainfourny: I also renamed the context package. :)

PR #597, opened 2020-03-26
Format generated parser files

CanBerker:
@ghislainfourny feel free to merge this if it looks good to you.

I don't mind performing these formatting tasks but I am wondering if we really need to IntelliJ touch for this or not.

Since spotless dictates our formatting anyways, I believe all IntelliJ does it converting wildcard imports into explicit imports. I remember that you had also configured your Eclipse settings to not use wildcard imports. I would be curious to know if same result is obtained when Eclipse IDE is used.

### DIFF
--- a/src/main/java/org/rumbledb/parser/JsoniqBaseVisitor.java
+++ b/src/main/java/org/rumbledb/parser/JsoniqBaseVisitor.java
@@ -11,658 +11,1029 @@ import org.antlr.v4.runtime.tree.AbstractParseTreeVisitor;
  * of the available methods.
  *
  * @param <T> The return type of the visit operation. Use {@link Void} for
- * operations with no return type.
+ *            operations with no return type.
  */
 public class JsoniqBaseVisitor<T> extends AbstractParseTreeVisitor<T> implements JsoniqVisitor<T> {
-	/**
-	 * {@inheritDoc}
-	 *
-	 * <p>The default implementation returns the result of calling
-	 * {@link #visitChildren} on {@code ctx}.</p>
-	 */
-	@Override public T visitModule(JsoniqParser.ModuleContext ctx) { return visitChildren(ctx); }
-	/**
-	 * {@inheritDoc}
-	 *
-	 * <p>The default implementation returns the result of calling
-	 * {@link #visitChildren} on {@code ctx}.</p>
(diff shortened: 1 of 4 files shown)


ghislainfourny:
@CanBerker so I managed to get eclipse to remove the wildcard imports. However, generating with ant, fixing wildcards, and then applying spotless nevertheless creates a big diff. We can also decide that this is not such a big deal (in the long term, Stefan suggested we leave out automatically generated files from the repo).

ghislainfourny:
@CanBerker I was able to get Eclipse to remove wildcard imports, however using this and spotless nevertheless generates a big diff.

We could also agree to stop using the IntelliJ trick, and instead just apply spotless and wildcard removal to the antlr output, to all get the same results (and small diffs). Do you agree?

Another possibility in the long term as Stefan suggests is to leave out automatically generated files from the repo.

CanBerker:
>  I was able to get Eclipse to remove wildcard imports, however using this and spotless nevertheless generates a big diff.

I am glad to hear that you've got that configuration in place as well. I think an overwhelming majority of the diff is generated by the spotless plugin. Hence our IDE choice plays only an extremely minor role in the diff size.

> We could also agree to stop using the IntelliJ trick, and instead just apply spotless and wildcard removal to the antlr output, to all get the same results (and small diffs). Do you agree?

I would like to emphasize that what we call the IntelliJ trick here is merely removing the wildcard imports (unless I am missing a tiny unnoticed detail in what it performs). All file formatting performed by spotless overrides the IntelliJ's formatting according to our style guide. 

> Another possibility in the long term as Stefan suggests is to leave out automatically generated files from the repo.

This a good suggestion to address the diff sizes. However, we need to update our travis pipelines as  well as the developer guidelines (so that new people coming on board know how to generate the antlr files).

CanBerker:
I have an experiment suggestion by the way, could you create a PR with your Eclipse + spotless formatted parser files. I will apply intelliJ formatting + spotless on top and we can see if any diff is produced

ghislainfourny:
This is done (#600) Interestingly, eclipse fixes more things. If however I apply spotless after Eclipse, the diff is empty.

ghislainfourny:
I took over the four files manually.

PR #598, opened 2020-03-27
Group by clean

ghislainfourny:


pom.xml
```diff
@@ -150,7 +150,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-release-plugin</artifactId>
-                <version>2.5.3</version>
```
CanBerker: Does this change mean that it will always be the most recent version?

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -179,20 +179,27 @@ public StaticContext visitLetClause(LetClause clause, StaticContext argument) {
...
-        } else {
-            // if a variable is only referenced, use the context as is
-            groupByClauseContext = argument;
+    public StaticContext visitGroupByClause(GroupByClause clause, StaticContext argument) {
+        StaticContext groupByClauseContext = new StaticContext(argument);
+        for (GroupByVariableDeclaration variable : clause.getGroupVariables()) {
+            if (variable.getExpression() != null) {
+                // if a variable declaration takes place
+                this.visit(variable.getExpression(), argument);
+                // initialize execution and storage modes and then add the variable to the context
```
CanBerker: I think this comment is outdated/redundant now.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -210,18 +217,6 @@ public StaticContext visitCountClause(CountClause expression, StaticContext argu
...
-        // TODO for now we only suppot as/default, no inference, flags
-        result.addVariable(
-            expression.getVariableReference().getVariableName(),
-            expression.getSequenceType(),
-            expression.getMetadata(),
-            expression.getVariableHighestStorageMode(this.visitorConfig)
-        );
-        return result;
-    }
-    // endregion
```
CanBerker: Very nice !

src/main/java/org/rumbledb/runtime/flwor/clauses/GroupByClauseSparkIterator.java
```diff
@@ -165,25 +164,26 @@ private void setAllLocalResults() {
                     groupVariableExpression.close();
 
                     // if a new variable is declared inside the group by clause, insert value in tuple
-                    inputTuple.putValue(expression.getVariableReference().getVariableName(), newVariableResults);
+                    inputTuple.putValue(expression.getVariableName(), newVariableResults);
                     results.addAll(newVariableResults);
 
                 } else { // if grouping on a variable reference
-                    VariableReferenceIterator groupVariableReference = expression.getVariableReference();
```
CanBerker: It's great to see that you've managed to tame this beast :) I had tried about a month ago but there was no way I could unentangle these as quickly as I had hoped for.

pom.xml
ghislainfourny: I was trying to suppress a warning. I think the version is imposed by another plugin version and this was colliding.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: I removed it in the other branch

PR #600, opened 2020-03-27
Experiment: Apply Eclipse format.

ghislainfourny:
I imported the XML file with the Rumble eclipse preferences, and asked Eclipse to apply format.


CanBerker:
Interestingly, when I applied IntelliJ formatting and then spotless, there was still quite some diff that I believe spotless does not really worry about. Yet IDE configurations have their own preferences.  

If you'd like, we can close this PR and merge the other one and stick to IntelliJ configurations for the time being. We can hopefully remove the autogenerated files in the future.

ghislainfourny:
An alternative is to not worry about the diffs, and only requires that
(i) spotless tests pass
and
(ii) imports wildcards were removed

That way I wouldn't have to disturb you every time these files get changed. In any case we can still merge the PR you prepared for this time.

Does it work?

CanBerker:
Sounds great

PR #601, opened 2020-03-27
Typeswitch fix2

### LINKED ISSUE #594: TypeSwitch Inconsistent Behavior - Potential Bug
TypeSwitch has an inconsistent behavior with input string length and how default case operates on it. I can't quite guess the reason but it seems buggy to me nonetheless. @ghislainfourny could you take a look at this as well?

The query - no string case, only default:
```
for $i in json-file("./src/main/resources/queries/test.txt")
where typeswitch($i.str)
    case integer | decimal return false
    default return true
return $i
```

The data - test.txt:
```
{"str": 1}
{"str": 2}
{"str":  3.5}
{"str":  "asdffwe"}
{"str":  "cd"}
```
The result:
```
{ "str" : "asdffwe" }
{ "str" : "cd" }
```

The data - test.txt:
** Notice the longer string ** 
```
{"str": 1}
{"str": 2}
{"str":  3.5}
{"str":  "asdffwesdafafasdf"}
{"str":  "cd"}
```

The result:
```
{ "str" : "cd" }
```
The long string disappears for the second dataset. 

In contrast, if the query is updated to have a case for string, this anomaly does not occur.
The query - added string case:
```
for $i in json-file("./src/main/resources/queries/test.txt")
where typeswitch($i.str)
    case integer | decimal return false
    case string return true
    default return true
return $i
```

The data - test.txt:
```
{"str": 1}
{"str": 2}
{"str":  3.5}
{"str":  "asdffwe"}
{"str":  "cd"}
```
The result:
```
{ "str" : "asdffwe" }
{ "str" : "cd" }
```

The data - test.txt:
** Notice the longer string ** 
```
{"str": 1}
{"str": 2}
{"str":  3.5}
{"str":  "asdffwesdafafasdf"}
{"str":  "cd"}
```

The result:
```
{ "str" : "asdffwesdafafasdf" }
{ "str" : "cd" }
```

ghislainfourny:
Fixes #594


src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
```diff
@@ -114,6 +126,7 @@ private boolean testTypeMatch(TypeswitchRuntimeIteratorCase typeSwitchCase) {
                 }
             }
         }
+        this.matchingIterator = null;
```
ghislainfourny: I think this was causing the issue

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
```diff
@@ -45,16 +45,25 @@ public TypeswitchRuntimeIterator(
...
     }
 
     @Override
     public Item next() {
         if (this.hasNext) {
-            this.matchingIterator.open(this.currentDynamicContextForLocalExecution);
             Item nextItem = this.matchingIterator.next();
             this.matchingIterator.close();
-            this.hasNext = false;
+            this.hasNext = this.matchingIterator.hasNext();
```
ghislainfourny: This is a fix to make sure that it forwards *all* items, not just one.

CanBerker:
It seems spotless formatting is needed

src/main/resources/test_files/runtime/ControlIterators/TypeswitchExpression7.jq
```diff
@@ -0,0 +1,12 @@
...
+{"str": 1},
+{"str": 2},
+{"str":  3.5},
+{"str":  "asdffwesdafafasdf"},
+{"str":  "cd"}
+)
+where typeswitch($i.str)
+    case integer | decimal return false
+    default return true
+return $i
```
CanBerker: Thanks for taking care of this. The tests look good

ghislainfourny:
> The tests look good and I believe this fixes the correctness issue. As a general remark, the use of matchingIterator in this file looked like spaghetti code to me. The state is modified everywhere and code is hard to trace and understand. I think we should try to do make it clearer. What do you think?

Actually this fixes the state too: matchingIterator is now null except when it is being actually used, and is also correctly closed and set back to null when it is no longer used.

CanBerker:
From a correctness perspective I see what you mean, I will highlight my concerns at respective places in the code now

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
```diff
@@ -45,16 +45,23 @@ public TypeswitchRuntimeIterator(
     @Override
     public void open(DynamicContext context) {
         super.open(context);
+        if (this.matchingIterator != null) {
+            this.matchingIterator.close();
+        }
+        this.matchingIterator = null;
```
CanBerker: I may be missing something here so please feel free to correct me if you sense a misunderstanding.

The code here looks quite suspicious to me. The iterator initialization takes place at line 52, and the class initializes the matching iterator as null at line 22. So matching iterator should be conclusively null when open() executes. To my understanding, this is not the case, this could indicate a mishandling of the state. So this should not be fixed in open() but where the mishandling occurred.

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
CanBerker: This line also looks quite interesting. Line 59 closes matching iterator and a closed iterator always has its hasNext set to false. So line 60 is equivalent to this.hasNext = false.

Furthermore, this makes the if check at line 61 redundant as we know !this.hasNext will always evaluate to true.

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
CanBerker: This is a very nice catch.

I will discuss an orthogonal issue to what you have fixed here. I believe this function does 'too many' things in contrast to what its name implies. Lines type checks performed in 112-114 alongside returning true or false indeed perform a type check. But this function modifies state by adding variables and updating the matching iterator as well. This makes it a very likely spot to create future bugs. For example, a developer might instinctively move the code at line 89:
`this.testValue = test.materializeFirstItemOrNull(this.currentDynamicContextForLocalExecution);` after the loop that's performing the testTypeMatch() to increase locality of the testValue variable by bringing it closer to where it's used. This could create some nasty bugs due to the obfuscated modification of the state that alters the dynamic context.

I would recommend only keeping the lines I mentioned earlier in this function to make sure it 'only' performs a type matching. And the modification of the state in lines 115-199: 
```
if (typeSwitchCase.getVariableName() != null) {
    ....
}
 this.matchingIterator = typeSwitchCase.getReturnIterator(); 
``` 
as well as the line you added at 126 `this.matchingIterator = null;` is taken out of this function. They could either be directly placed within the loop in initializeIterator or extracted to a separate function.

ghislainfourny: Done, I have made the iterator the return value, so that initializeIterator is now responsible for modifying the state, and it is now handling cases and the default case with similar code.

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
ghislainfourny: This was a bug, the first close call should have been removed as well.

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
ghislainfourny: Also a good point. Replaced with a check and exception throwing.

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
```diff
@@ -40,21 +40,35 @@ public TypeswitchRuntimeIterator(
         this.testField = test;
         this.cases = cases;
         this.defaultCase = defaultCase;
+        this.matchingIterator = null;
     }
 
     @Override
     public void open(DynamicContext context) {
         super.open(context);
+        // this.matchingIterator is null at that point;
```
CanBerker: I think we can leave out this comment. It restates what is in the code

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
CanBerker: Looks beautiful now. Thank you

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
CanBerker: Glad to help. This looks very clean now

ghislainfourny:
@CanBerker I added tests and also removed the fields passed as parameters to initializeIterator()

PR #602, opened 2020-03-28
Implementation of parallel intersect function.

ioanas96:


src/main/java/org/rumbledb/runtime/functions/object/ObjectIntersectFunctionIterator.java
```diff
@@ -96,4 +104,54 @@ public Item next() {
...
+    protected JavaRDD<Item> getRDDAux(DynamicContext context) {
+        JavaRDD<Item> childRDD = this.iterator.getRDD(context);
+        Function2<Item, Item, Item> transformation = new ObjectIntersectClosure();
+        Item reduceResult = childRDD.reduce(transformation);
+
+        // flatten values for each key
+        LinkedHashMap<String, List<Item>> keyValuePairs = new LinkedHashMap<>();
+        for (String key : reduceResult.getKeys()) {
+            Item value = reduceResult.getItemByKey(key);
+            List<Item> flatValues = flatten(Collections.singletonList(value));
```
ioanas96: @ghislainfourny, by default the reduce operation doesn't flatten the values of the resulted keys. Flattening the results of the intersection keys is not parallelized here, do you think it would be desired? I am not sure if it is worth parallelizing this because it will also need additional data movements between data structures. What do you think?

ghislainfourny: I think we do not need to parallelize for now, however, what is the result of this?

intersect(for $i in parallelize(1 to 100) return { "foo" : [ [ "bar" ] ] })

ioanas96: For nested arrays, the current implementation returns a completely flattened array indeed: { "foo" : [ "bar", "bar", ... "bar", "bar" ] } for this case, when the actual result when not parallelized is { "foo" : [ [ [ "bar" ] ], [ [ "bar" ] ], ... [ [ "bar" ] ], [ [ "bar" ] ] ] }. But I don't see an imediate change to be done to the current implementation, because it will need to know the actual number of objects with this key (maybe this could be counted globally?)

ioanas96: I have added a fix to this problem with the changes we talked about: each individual value is enclosed in an ArrayItem at first, then 2 values are kept into an ArrayValue during the reduce operation. I also added a test similar to the one above.

PR #608, opened 2020-03-31
Fix access to other storage layers than HDFS.

ghislainfourny:
@CanBerker this is the first release since the introduction of the URL validator. Unfortunately, it does not work when reading from other file systems than HDFS on Amazon EMR (like S3). I had to fix the code as we need to use FileContext, not FileSystem.

### DIFF
--- a/src/main/java/org/rumbledb/runtime/functions/input/UrlValidator.java
+++ b/src/main/java/org/rumbledb/runtime/functions/input/UrlValidator.java
@@ -23,13 +23,12 @@ public class UrlValidator {
         } catch (URISyntaxException e) {
             throw new OurBadException("Malformed URL: " + url);
         }
-        if (locator.isAbsolute() && !Arrays.asList(allowedSchemes).contains(locator.getScheme()))
-        {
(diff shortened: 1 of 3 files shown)


src/main/java/org/rumbledb/runtime/functions/input/UrlValidator.java
```diff
@@ -16,28 +16,34 @@
...
+                return url.contains("*") || fileContext.util().exists(path);
+
+            } catch (UnsupportedFileSystemException e) {
+                throw new CannotRetrieveResourceException(
+                        "No file system is configured for scheme " + url + "!",
+                        metadata
+                );
             } catch (IOException e) {
-                throw new SparksoniqRuntimeException("Error while accessing hadoop filesystem.");
+                e.printStackTrace();
```
CanBerker: Is this printStackTrace intentionally kept here?

ghislainfourny: Yes, to debug more easily if something happens.

PR #609, opened 2020-03-31
Clean up cli

ghislainfourny:
This is a first step to clean up the Rumble execution. Now the code to execute locally or on a cluster is the same, which should reduce the risk of issues when releasing. 

All paths (absolute, relative, any system) now should work with --query-path and --log-path.

Also, the shell no longer writes the query to a temporary file.


src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -100,51 +102,22 @@ public void runLocal(String queryFile, String outputPath) throws IOException {
...
         if (result.isRDD() && outputPath != null) {
             JavaRDD<Item> rdd = result.getRDD(dynamicContext);
             JavaRDD<String> output = rdd.map(o -> o.serialize());
             output.saveAsTextFile(outputPath);
         } else {
             String output = runIterators(result, dynamicContext);
             if (outputPath != null) {
+                File outputFile = new File(outputPath);
                 List<String> lines = Arrays.asList(output);
                 org.apache.commons.io.FileUtils.writeLines(outputFile, "UTF-8", lines);
```
ghislainfourny, line 115: The plan is to also use the hadoop filesystem for writing back small output, similar to everything else, but I have not gotten to it yet.

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -194,30 +167,6 @@ public String runInteractive(java.nio.file.Path queryFile) throws IOException {
         return rddOutput;
     }
 
-    private JsoniqLexer getInputSource(String arg) throws IOException {
-        arg = arg.trim();
-        // return embedded file
```
ghislainfourny, line 199: This was causing a lot of issues, it is now all handled in a unified way with the hadoop filesystem methods.

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -21,73 +21,65 @@
...
+    private void checkOutputFile(String outputPath) throws IOException {
         if (outputPath != null) {
-            outputFile = new File(outputPath);
-            if (outputFile.exists()) {
+            if (UrlValidator.exists(outputPath, new ExceptionMetadata(0, 0))) {
                 if (!this.configuration.getOverwrite()) {
                     System.err.println(
                         "Output path " + outputPath + " already exists. Please use --overwrite yes to overwrite."
                     );
                     System.exit(1);
```
ghislainfourny, line 66: The Hadoop FileContext class will overwrite, no need to delete any more.

src/main/java/org/rumbledb/runtime/functions/input/FileSystemUtil.java
```diff
@@ -0,0 +1,152 @@
...
+        } catch (Exception e) {
+            e.printStackTrace();
+            throw new CannotRetrieveResourceException(
+                    "Error while accessing the " + locator.getScheme() + " filesystem.",
+                    metadata
+            );
+        }
+    }
+
+    public static boolean delete(String url, ExceptionMetadata metadata) {
```
ghislainfourny, line 56: Not actually needed any more, but might be in the future.

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -108,79 +101,24 @@ public void runLocal(String queryFile, String outputPath) throws IOException {
...
-            JavaRDD<String> output = rdd.map(o -> o.serialize());
-            output.saveAsTextFile(outputPath);
-        }
         long endTime = System.currentTimeMillis();
         long totalTime = endTime - startTime;
-        if (this.configuration.getLogPath() != null) {
-            writeTimeLog(totalTime);
+        String logPath = this.configuration.getLogPath();
+        if (logPath != null) {
+            String time = "[ExecTime] " + totalTime;
```
CanBerker, line 114: The other execution time message has been refactored to:
`output("The query took " + time + " milliseconds to execute.");`

Was it intentional to leave this one as Exectime?

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
CanBerker: Cleanup on this file looks amazing

PR #611, opened 2020-04-02
Implemented avro function

CodingKaiser:


pom.xml
```diff
@@ -215,6 +215,11 @@
             <version>2.4.3</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-avro_2.12</artifactId>
+            <version>2.4.5</version>
```
ghislainfourny, line 221: does <scope>provided</scope> work on your side (if the Spark environment provides it)? This reduces the size of the jar...

CodingKaiser:
Reason for the bug is the following: am having trouble with spark library necessary to run it, and it seems at first glance to perhaps be something with the configuration.

In the current merge request, I have `org.apache.spark:spark-avro_2.12:2.4.5` listed in pom.xml as a dependency which allows for all tests I've written to run as expected, which makes sense because the rest of the spark dependencies are also running on 2.12.

However, trying to run rumble on the command line, i.e. with (`spark-submit --master local[2] --deploy-mode client target/spark-rumble-1.5-jar-with-dependencies.jar --shell yes --show-error-info yes`) and calling avro-file() results in a fatal error:

```
rumble$ avro-file("src/main/resources/queries/userdata1-limit.avro")
>>> 
>>> 
Exception in thread "main" java.util.ServiceConfigurationError: org.apache.spark.sql.sources.DataSourceRegister: Provider org.apache.spark.sql.avro.AvroFileFormat could not be instantiated
    at java.util.ServiceLoader.fail(ServiceLoader.java:232)
    at java.util.ServiceLoader.access$100(ServiceLoader.java:185)
    at java.util.ServiceLoader$LazyIterator.nextService(ServiceLoader.java:384)
    at java.util.ServiceLoader$LazyIterator.next(ServiceLoader.java:404)
    at java.util.ServiceLoader$1.next(ServiceLoader.java:480)
    at scala.collection.convert.Wrappers$JIteratorWrapper.next(Wrappers.scala:43)
...
```

If I then switch the version of `spark-avro to org.apache.spark:spark-avro_2.11:2.4.3`, everything runs on the command-line error-free, at the cost of causing all of the avro-related .jq and in some cases csv-related .jq tests to fail (I would assume because they are both a part of `org.apache.spark.sql`).

Would you have any intuition as to what the problem here is? StackOverflow has thus far not been very useful.

pom.xml
CodingKaiser: Not exactly sure what you meant, but according to [this page](https://spark.apache.org/docs/latest/sql-data-sources-avro.html), "The `spark-avro` module is external and not included in `spark-submit` or `spark-shell` by default."

ghislainfourny:
@CodingKaiser which version of Java and of Spark are you using?

pom.xml
ghislainfourny: Yes indeed, thank you for checking. Indeed then it is the right configuration.

CodingKaiser:
The Java version is the following: `openjdk version "1.8.0_242"`
The Spark version is `2.4.5`

ghislainfourny:
Thanks! It seems like a known bug with the package. I also have the issue.

https://issues.apache.org/jira/browse/SPARK-27623

If you use the 2.11 package, does it work with the test queue? If so, this would be an acceptable solution, as indicated on the above link as well.

CodingKaiser:
Unfortunately, switching to the 2.11 package breaks some of the tests.

ghislainfourny:
Very good catch. Official Spark releases are packaged with Scala 2.11 (try spark-submit --version). However, we were using a compilation with Scala 2.12. Somehow it worked, and we didn't notice.

Spark switched to Scala 2.12 in version 2.4.2, but reverted to Scala 2.11 in version 2.4.3.

I have reverted to Spark built with Scala 2.11 in pom.xml and it seems to do the trick. We are now using the officially supported Scala version.

CodingKaiser:
Great news! Thanks for helping with this issue.

CodingKaiser:
Also, I was not able to confirm with absolute certainty that these data are not made up, but looking at each of the entries strongly suggests that they are nonsense. Case in point: `{...  "first_name" : "Nicole", "last_name" : "Willis", "email" : "nwillis2g@cmu.edu",  "country" : "Indonesia", "salary" : 258772.36, "title" : "Physical Therapy Assistant" ... }`. I got the data from [kylo](https://github.com/Teradata/kylo).

PR #617, opened 2020-04-06
Throw error if cannot materialize.

ghislainfourny:
TODO: reintroduce materialization cap in the tests, otherwise cannot test for this error.


ghislainfourny:
As suggested by users, we should throw an error if, within a query, the materialization cap is reached (Only exception is the shell if the output of the query is capped, then only a warning).

I am well aware that some of the code is duplicated between tests and JSONiq executor -- the plan is, in the future, to do it with a clean serialization API with parameters, but this would require a lot more time.

docs/Error codes.md
```diff
@@ -42,10 +42,10 @@ big sequence of items for each binding within a big tuple,
 which would lead to a data flow explosion and to a nesting of jobs on the Spark cluster.
 
 - [RBST0004] - Implementation Error.
-We should investigate this 🙈. Please contact us or file an issue on GitHub with your query.
-Link: https://github.com/RumbleDB/rumble/issues
 
-- [SENR0001] - Serialization error. Function items can not be serialized
+- [RBDY0005] - Materialization Error: the sequence is too big to be materialized. Use --result-size to increase the maximum  materialization size.
```
CanBerker, line 46: If you agree as well, I was trying to keep an alphabetic order in Error codes.md and ErrorCodes.java files. Could you move this error before RBML errros?

src/main/java/org/rumbledb/errorcodes/ErrorCodes.java
```diff
@@ -59,6 +59,7 @@
     public static final String UnimplementedErrorCode = "RBST0002";
     public static final String JobWithinAJobErrorCode = "RBST0003";
     public static final String OurBadErrorCode = "RBST0004";
+    public static final String CannotMaterializeErrorCode = "RBDY0005";
```
CanBerker, line 62: Same comment with Error codes.md

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -201,32 +196,29 @@ private String getIteratorOutput(RuntimeIterator iterator, DynamicContext dynami
                 sb.append("\n");
                 itemCount++;
             }
+            if (iterator.hasNext() && itemCount == this.configuration.getResultSizeCap()) {
+                System.err.println(
+                    "Warning! The output sequence contains a large number of items but its materialization was capped at "
+                        + SparkSessionManager.COLLECT_ITEM_LIMIT
+                        + " items. This value can be configured with the --result-size parameter at startup"
+                );
+            }
```
CanBerker, line 205: I believe showing this error after the query results would be much more noticeable to the users, but I'm not sure how much effort that would require

docs/Error codes.md
ghislainfourny: DONE

src/main/java/org/rumbledb/errorcodes/ErrorCodes.java
ghislainfourny: DONE

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
ghislainfourny: I tried it but will require more efforts.

PR #620, opened 2020-04-07
Closure

### LINKED ISSUE #619: Parameter values from outside not included in closure
This query returns an error Error [err: XPDY0130]LINE:4;COLUMN:23;Runtime error retrieving variable coordinates value

declare	function new-coordinates($coordinates as object) as object
{
  {
    "x"	: function() { $coordinates }
  }
};

let $c := new-coordinates({"x" : 2, "y" : 4})
return $c.x()

ghislainfourny:
There was an issue with closures.

Fixes #619

### DIFF
--- a/src/main/java/org/rumbledb/context/DynamicContext.java
+++ b/src/main/java/org/rumbledb/context/DynamicContext.java
@@ -346,6 +346,34 @@ public class DynamicContext implements Serializable, KryoSerializable {
         }
     }
 
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("  Local:\n");
+        for (String name : this.localVariableValues.keySet()) {
+            sb.append("    " + name + " (" + this.localVariableValues.get(name).size() + " items)\n");
+            if (this.localVariableValues.get(name).size() == 1) {
+                sb.append("      " + this.localVariableValues.get(name).get(0).serialize() + "\n");
+            }
+        }
+        sb.append("  Counts:\n");
+        for (String name : this.localVariableCounts.keySet()) {
+            sb.append("    " + name + " (" + this.localVariableCounts.get(name) + " items)\n");
+        }
+        sb.append("  RDD:\n");
+        for (String name : this.rddVariableValues.keySet()) {
+            sb.append("    " + name + " (" + this.rddVariableValues.get(name).count() + " items)\n");
+        }
+        sb.append("  Data Frames:\n");
+        for (String name : this.dataFrameVariableValues.keySet()) {
+            sb.append("    " + name + " (" + this.dataFrameVariableValues.get(name).count() + " items)\n");
+        }
+        if (this.parent != null) {
+            sb.append("Parent context:\n");
+            sb.append(this.parent.toString());
+        }
+        return sb.toString();
+    }
 
 }
 
--- a/src/main/java/org/rumbledb/context/DynamicContext.java
+++ b/src/main/java/org/rumbledb/context/DynamicContext.java
@@ -164,7 +164,7 @@ public class DynamicContext implements Serializable, KryoSerializable {
 
         if (this.rddVariableValues.containsKey(varName)) {
             JavaRDD<Item> rdd = this.getRDDVariableValue(varName, metadata);
-            return SparkSessionManager.collectRDDwithLimit(rdd);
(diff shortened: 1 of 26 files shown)


src/main/java/org/rumbledb/runtime/functions/FunctionItemCallIterator.java
```diff
@@ -212,7 +212,7 @@ private DynamicContext createNewDynamicContextWithArguments(DynamicContext conte
                 localArgumentValues.put(argName, argIterator.materialize(context));
             }
         }
-        return new DynamicContext(context, localArgumentValues, RDDArgumentValues, DFArgumentValues);
+        return new DynamicContext(null, localArgumentValues, RDDArgumentValues, DFArgumentValues);
```
ghislainfourny, line 215: The outer dynamic context is not passed to a function call. It is taken from the closure and parameter instead.

src/main/java/org/rumbledb/runtime/functions/FunctionRuntimeIterator.java
```diff
@@ -46,7 +46,9 @@ public FunctionRuntimeIterator(
     public Item next() {
         if (this.hasNext) {
             this.hasNext = false;
-            return this.item;
+            FunctionItem result = ((FunctionItem) this.item).deepCopy();
+            result.populateClosureFromDynamicContext(this.currentDynamicContextForLocalExecution, getMetadata());
```
ghislainfourny, line 50: First fix: when an inline function expression is evaluated, the closure must be populated from the dynamic context.

src/main/java/org/rumbledb/context/DynamicContext.java
```diff
@@ -346,6 +346,34 @@ public static void mergeVariableDependencies(
         }
     }
 
+    @Override
+    public String toString() {
```
ghislainfourny, line 350: This will be useful for future debugging.

src/main/java/org/rumbledb/items/FunctionItem.java
```diff
@@ -176,6 +178,36 @@ public String serialize() {
         throw new FunctionsNonSerializableException();
     }
 
+    @Override
+    public String toString() {
```
ghislainfourny, line 182: This will be useful for future debugging.

src/main/java/org/rumbledb/runtime/functions/FunctionItemCallIterator.java
```diff
@@ -185,7 +185,7 @@ private FunctionRuntimeIterator generatePartiallyAppliedFunction(DynamicContext
                 RDDArgumentValues,
                 DFArgumentValues
         );
-        return new FunctionRuntimeIterator(partiallyAppliedFunction, ExecutionMode.LOCAL, getMetadata());
+        return new ConstantRuntimeIterator(partiallyAppliedFunction, ExecutionMode.LOCAL, getMetadata());
```
ghislainfourny, line 188: Third fix: after computing a partially applied function, the iterator we return is only a wrapper on the generated function item, which must be returned *as is*. This is not an inline function expression evaluation.

src/main/java/org/rumbledb/runtime/functions/input/ParallelizeFunctionIterator.java
```diff
@@ -59,7 +59,7 @@ public ParallelizeFunctionIterator(
             rdd = SparkSessionManager.getInstance().getJavaSparkContext().parallelize(contents);
         } else {
             RuntimeIterator partitionsIterator = this.children.get(1);
-            partitionsIterator.open(this.currentDynamicContextForLocalExecution);
+            partitionsIterator.open(context);
```
ghislainfourny, line 62: Fourth fix.

src/main/java/org/rumbledb/runtime/ConstantRuntimeIterator.java
```diff
@@ -0,0 +1,52 @@
...
+        super(null, executionMode, iteratorMetadata);
+        this.item = item;
+    }
+
+    @Override
+    public Item next() {
+        if (this.hasNext) {
+            this.hasNext = false;
+            Item result = this.item;
+            return result;
```
CanBerker, line 47: This can be simplified to `return this.item;`

PR #624, opened 2020-04-09
Simple map expression (local implementation)

CodingKaiser:
Implemented the local version of the simple map expression. The parallelized implementation will follow in a separate commit once this one has been merged.


src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -557,6 +558,25 @@ public Node visitMultiplicativeExpr(JsoniqParser.MultiplicativeExprContext ctx)
         return result;
     }
 
+    @Override
+    public Node visitSimpleMapExpr(JsoniqParser.SimpleMapExprContext ctx) {
+        Expression result = (Expression) this.visitPostFixExpr(ctx.main_expr);
+        if (ctx.map_expr == null || ctx.map_expr.isEmpty()) {
```
CodingKaiser, line 564: Noticed this logic for very similar expressions. Will the parser even allow for expression to be empty?

src/main/java/org/rumbledb/runtime/flwor/expression/SimpleMapExpressionIterator.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import java.util.TreeMap;
+
+public class SimpleMapExpressionIterator extends LocalRuntimeIterator {
+
+    private static final long serialVersionUID = 1L;
+    private RuntimeIterator iterator;
```
ghislainfourny, line 43: I recommend more readable names: leftIterator, rightIterator

src/main/java/org/rumbledb/runtime/flwor/expression/SimpleMapExpressionIterator.java
```diff
@@ -0,0 +1,131 @@
...
+        this.iterator.close();
+    }
+
+    @Override
+    public void open(DynamicContext context) {
+        super.open(context);
+        if (this.children.size() < 2) {
+            throw new OurBadException("Invalid simple map! Must initialize map before calling next");
+        }
+        this.mapDynamicContext = new DynamicContext(this.currentDynamicContextForLocalExecution);
```
ghislainfourny, line 88: Is this really needed? It will be created anyway in setNextResult() for each item...

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: Perhaps not... it depends how extra cautious we want to be.

PR #627, opened 2020-04-14
Fix return clause parallel

### LINKED ISSUE #615: Undue materialization
count(let $i := parallelize(1 to 1000) return $i)

No materialization should happen here.

ghislainfourny:
Fixes #615 

This should help with your demo.

Enjoy!

### DIFF
--- a/src/main/java/org/rumbledb/expressions/flowr/ReturnClause.java
+++ b/src/main/java/org/rumbledb/expressions/flowr/ReturnClause.java
@@ -46,18 +46,15 @@ public class ReturnClause extends Clause {
 
     @Override
     public void initHighestExecutionMode(VisitorConfig visitorConfig) {
-        if(this.previousClause.getHighestExecutionMode(visitorConfig).isDataFrame())
-        {
+        if (this.previousClause.getHighestExecutionMode(visitorConfig).isDataFrame()) {
             this.highestExecutionMode = ExecutionMode.RDD;
             return;
         }
-        if(this.returnExpr.getHighestExecutionMode(visitorConfig).isRDD())
-        {
+        if (this.returnExpr.getHighestExecutionMode(visitorConfig).isRDD()) {
             this.highestExecutionMode = ExecutionMode.RDD;
             return;
         }
-        if(this.returnExpr.getHighestExecutionMode(visitorConfig).isDataFrame())
-        {
+        if (this.returnExpr.getHighestExecutionMode(visitorConfig).isDataFrame()) {
             this.highestExecutionMode = ExecutionMode.DATAFRAME;
             return;
         }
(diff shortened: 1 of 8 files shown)


src/main/java/org/rumbledb/runtime/flwor/clauses/ReturnClauseSparkIterator.java
```diff
@@ -63,18 +65,75 @@ public ReturnClauseSparkIterator(
...
-            throw new JobWithinAJobException(
-                    "A return clause expression cannot produce a big sequence of items for a big number of tuples, as this would lead to a data flow explosion.",
-                    getMetadata()
-            );
+            if (this.child.isDataFrame())
+                throw new JobWithinAJobException(
+                        "A return clause expression cannot produce a big sequence of items for a big number of tuples, as this would lead to a data flow explosion.",
+                        getMetadata()
+                );
+            this.tupleContext = new DynamicContext(this.currentDynamicContextForLocalExecution); // assign current
```
CanBerker, line 73: this.currentDynamicContextForLocalExecution should be used only in local methods. This should probably be 'context' variable

src/main/java/org/rumbledb/runtime/flwor/clauses/ReturnClauseSparkIterator.java
```diff
@@ -63,18 +65,75 @@ public ReturnClauseSparkIterator(
...
+    @Override
+    public Dataset<Row> getDataFrame(DynamicContext context) {
+        RuntimeIterator expression = this.children.get(0);
+        if (expression.isRDD()) {
+            if (this.child.isDataFrame())
+                throw new JobWithinAJobException(
+                        "A return clause expression cannot produce a big sequence of items for a big number of tuples, as this would lead to a data flow explosion.",
+                        getMetadata()
+                );
+            this.tupleContext = new DynamicContext(this.currentDynamicContextForLocalExecution); // assign current
```
CanBerker, line 109: Same comment

src/main/java/org/rumbledb/runtime/flwor/clauses/ReturnClauseSparkIterator.java
ghislainfourny: Yes, good catch and that was indeed one of the issues. All is fixed now.

src/main/java/org/rumbledb/expressions/flowr/ReturnClause.java
```diff
@@ -46,10 +46,19 @@ public Expression getReturnExpr() {
...
+        }
+        if (this.returnExpr.getHighestExecutionMode(visitorConfig).isRDD()) {
+            this.highestExecutionMode = ExecutionMode.RDD;
+            return;
+        }
+        if (this.returnExpr.getHighestExecutionMode(visitorConfig).isDataFrame()) {
+            this.highestExecutionMode = ExecutionMode.DATAFRAME;
+            return;
+        }
+        this.highestExecutionMode = ExecutionMode.LOCAL;
```
CanBerker, line 61: Very nicely handled. I felt like I would be prone to making a mistake here :)

ghislainfourny: This was not a mistake -- this was just not implemented yet (because not trivial). :)

PR #628, opened 2020-04-14
Parallelization of descendant-pairs function.

ioanas96:


src/main/java/org/rumbledb/runtime/functions/object/ObjectDescendantPairsFunctionIterator.java
```diff
@@ -120,4 +120,29 @@ private void getDescendantPairs(List<Item> items) {
...
+
+    @Override
+    protected boolean hasNextLocal() {
+        return this.hasNext;
+    }
+
+    @Override
+    protected void resetLocal(DynamicContext context) {
+        this.iterator.open(this.currentDynamicContextForLocalExecution);
+        this.nextResults = new LinkedList<>();
```
ghislainfourny, line 132: why not just clear the list rather than recreating one?

ghislainfourny:
@ioanas96 could you also make sure that the document is updated  to say that these functions are pushed down? Thanks!

ghislainfourny:
@ioanas96 I recommend you try to merge the newest master into your research project branch and solve any conflicts -- indeed, this will make sure the tests are properly executed.

If you run into any issues with this, I can help.

src/main/java/org/rumbledb/runtime/functions/object/ObjectDescendantPairsFunctionIterator.java
ioanas96: Done.

PR #629, opened 2020-04-15
Parallelization for comma expression.

ioanas96:


src/main/java/org/rumbledb/expressions/CommaExpression.java
```diff
@@ -64,9 +64,17 @@ private boolean bypassCurrentExpressionForExecutionModeOperations() {
     @Override
     public void initHighestExecutionMode() {
         if (bypassCurrentExpressionForExecutionModeOperations()) {
+            this.highestExecutionMode = this.expressions.get(0).getHighestExecutionMode();
             return;
         }
-        super.initHighestExecutionMode();
+
+        ExecutionMode currentMode = ExecutionMode.LOCAL;
```
ghislainfourny, line 71: I think that, for a first implementation, it will be easier to do the opposite: only do RDD if all children expressions are RDDs, and otherwise local.

This is because the code can currently only materialize from RDD to local, but not the other way round.

src/main/java/org/rumbledb/expressions/CommaExpression.java
```diff
@@ -64,9 +64,17 @@ private boolean bypassCurrentExpressionForExecutionModeOperations() {
     @Override
     public void initHighestExecutionMode() {
         if (bypassCurrentExpressionForExecutionModeOperations()) {
```
ghislainfourny, line 66: I think the bypass is a leftover from earlier times that should no longer be needed.

PR #630, opened 2020-04-16
Arrow

ghislainfourny:
Support for => syntax (just a syntactic sugar, from XQuery 3.1)

### DIFF
--- a/src/main/java/org/rumbledb/compiler/TranslationVisitor.java
+++ b/src/main/java/org/rumbledb/compiler/TranslationVisitor.java
@@ -607,25 +607,25 @@ public class TranslationVisitor extends org.rumbledb.parser.JsoniqBaseVisitor<No
 
     @Override
     public Node visitArrowExpr(JsoniqParser.ArrowExprContext ctx) {
-        Expression mainExpression = (Expression) this.visitUnaryExpr(ctx.unaryExpr());
+        Expression mainExpression = (Expression) this.visitUnaryExpr(ctx.main_expr);
 
         if (ctx.functionCall() == null) {
             return mainExpression;
         }
 
         String name;
-        if (ctx.functionCall().fn_name != null) {
-            name = ctx.functionCall().fn_name.getText();
+        if (ctx.function_call_expr.fn_name != null) {
+            name = ctx.function_call_expr.fn_name.getText();
         } else {
-            name = ctx.functionCall().kw.getText();
+            name = ctx.function_call_expr.kw.getText();
         }
         if (ctx.functionCall().ns != null) {
-            name = name + ":" + ctx.functionCall().ns.getText();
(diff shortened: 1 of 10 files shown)


ghislainfourny:
The diff is because this time, we stop using the manual IntelliJ clean. I just generated the antlr files, applied spotless, and have Eclipse remove wildcard imports.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -605,6 +605,39 @@ public Node visitCastExpr(JsoniqParser.CastExprContext ctx) {
         return new CastExpression(mainExpression, sequenceType, createMetadataFromContext(ctx));
     }
 
+    @Override
+    public Node visitArrowExpr(JsoniqParser.ArrowExprContext ctx) {
+        Expression mainExpression = (Expression) this.visitUnaryExpr(ctx.unaryExpr());
```
ghislainfourny, line 610: E=>F(A, B, C)

is simply the same as

F(E, A, B, C)

so no need for new expression machinery.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -605,6 +606,39 @@ public Node visitCastExpr(JsoniqParser.CastExprContext ctx) {
...
+            FunctionCallContext functionCallContext = ctx.function_call_expr.get(i);
+            String name;
+            if (functionCallContext.fn_name != null) {
+                name = functionCallContext.fn_name.getText();
+            } else {
+                name = functionCallContext.kw.getText();
+            }
+            if (functionCallContext.ns != null) {
+                name = name + ":" + functionCallContext.ns.getText();
+            }
```
CanBerker, line 623: The code above calculating the name string is identical with visitFunctionCall expression. Can we extract this to a function to avoid duplication?

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -605,6 +606,39 @@ public Node visitCastExpr(JsoniqParser.CastExprContext ctx) {
...
+            List<Expression> arguments = new ArrayList<>();
+            arguments.add(mainExpression);
+            if (functionCallContext.argumentList().args != null) {
+                for (JsoniqParser.ArgumentContext arg : functionCallContext.argumentList().args) {
+                    Expression currentArg = (Expression) this.visitArgument(arg);
+                    arguments.add(currentArg);
+                }
+            }
```
CanBerker, line 631: Similarly, the contents of this if block is identical to the getArgumentsFromArgumentListContext() method. Can we call that method and join the returned list with the arguments list here?

ghislainfourny:
@CanBerker very good comments, this is all done. Thanks!

PR #634, opened 2020-04-17
Support for global variable declaration.

ghislainfourny:
We now support global variables! #4 

(For now, a variable only sees previous variables. In the future, we will generalize this to follow the spec, but this requires a DAG analysis on declarations).


src/main/java/org/rumbledb/runtime/RuntimeIterator.java
```diff
@@ -267,4 +267,18 @@ public void print(StringBuffer buffer, int indent) {
             iterator.print(buffer, indent + 1);
         }
     }
+
+    public void bindToVariableInDynamicContext(
+            DynamicContext targetContext,
+            String variable,
+            DynamicContext executionContext
```
CanBerker, line 274: This one looks beautiful

src/main/resources/test_files/runtime/VariableDeclaration3.jq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldCrash; ErrorCode="XPTY0004"; ErrorMetadata="LINE:2:COLUMN:33:" :)
+declare variable $x as string := 1;
```
CanBerker, line 2: This file should probably be named VariableDeclarationError

src/test/java/iq/RuntimeTests.java
```diff
@@ -139,15 +139,15 @@ protected String getIteratorOutput(RuntimeIterator iterator, DynamicContext dyna
             while (
                 iterator.hasNext()
                     &&
-                    ((itemCount < this.configuration.getResultSizeCap() && this.configuration.getResultSizeCap() > 0)
+                    ((itemCount < configuration.getResultSizeCap() && configuration.getResultSizeCap() > 0)
                         ||
-                        this.configuration.getResultSizeCap() == 0)
+                        configuration.getResultSizeCap() == 0)
```
CanBerker, line 144: What is the motivation behind this change ?

PR #647, opened 2020-04-28
Resolve global variable and function dependencies even when defined further down

ghislainfourny:


ghislainfourny:
@CanBerker don't hesitate to give feedback on this if you are interested. This is not quite done and I will write tests next week and fine tune.

With this, we should have full support for global variables and user-defined functions mutually referencing each other, even if their order is shuffled around.

CanBerker:
I will gladly take a look. Thanks for noting down the implementation goal to give me a rough idea on what to expect. It helps a lot :)

src/main/java/org/rumbledb/api/Rumble.java
```diff
@@ -68,6 +69,8 @@ public SequenceOfItems runQuery(String query) {
             e.initCause(ex);
             throw e;
         }
+        String[] args = new String[0];
+        VisitorHelpers.resolveDependencies(mainModule, new RumbleRuntimeConfiguration(args));
```
CanBerker, line 73: Can we extract new RumbleRuntimeConfiguration(new String[0]) to a variable that highlights the fact that this is just a placeholder configuration with no effects.

src/main/java/org/rumbledb/expressions/module/Prolog.java
```diff
@@ -27,42 +27,43 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class Prolog extends Node {
 
-    private final List<VariableDeclaration> variableDeclarations;
-    private final List<FunctionDeclaration> functionDeclarations;
+    private List<Node> declarations;
```
CanBerker, line 34: I do not exactly see the benefit of this refactoring at the moment. If we are planning to add more declaration types it seems appealing to not have separate lists for each type. However, it seems that the declaration type information is available in the parameters of the constructor. This is then lost, only to be recalculated by the getter methods. This feels redundant from a computational perspective. Is there something that I'm missing?

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+import org.rumbledb.expressions.primary.FunctionCallExpression;
+import org.rumbledb.expressions.primary.InlineFunctionExpression;
+import org.rumbledb.expressions.primary.NamedFunctionReferenceExpression;
+import org.rumbledb.expressions.primary.VariableReferenceExpression;
+import org.rumbledb.expressions.quantifiers.QuantifiedExpression;
+import org.rumbledb.expressions.quantifiers.QuantifiedExpressionVar;
+
+
+/**
+ * Dynamic context visitor. Populates the dynamic context to evaluate the main expression.
```
CanBerker, line 69: This comment needs to be updated

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+        }
+        if (!this.outputVariableDependencies.keySet().contains(node)) {
+            this.outputVariableDependencies.put(node, new TreeSet<String>());
+        }
+        this.outputVariableDependencies.get(node).add(variable);
+    }
+
+
+
+    VariableDependenciesVisitor(RumbleRuntimeConfiguration rumbleRuntimeConfiguration) {
```
CanBerker, line 148: Can we move the constructor to the top of the file for house-keeping?

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+    private RumbleRuntimeConfiguration rumbleRuntimeConfiguration;
+
+    Map<Node, Set<String>> inputVariableDependencies;
+    Map<Node, Set<String>> outputVariableDependencies;
+
+    private void addInputVariableDependencies(Node node, Set<String> variables) {
+        if (variables == null) {
+            addInputVariableDependencies(node, Collections.emptySet());
+            return;
+        }
```
CanBerker, line 82: This if block does not align with what the function name describes. I would expect this function to merely add variables dependencies list. However, if this function is every called with variables set as null, all previously stored set variable dependencies would be wiped and replaced with an emptySet(). This may cause bugs in the future.

Furthermore, I am sure that you are aware as well of the redundancy of creating sets (even as they are empty) when no dependencies exist. I am thinking of an alternative that uses getters returning a static empty set if the dependencies do not exist. This would avoid the redundantly generated sets for each node without variable dependencies.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+        if (!this.inputVariableDependencies.keySet().contains(node)) {
+            this.inputVariableDependencies.put(node, new TreeSet<String>());
+        }
+        this.inputVariableDependencies.get(node).addAll(variables);
+    }
+
+    private void removeInputVariableDependencies(Node node, Set<String> variables) {
+        if (variables == null) {
+            addInputVariableDependencies(node, Collections.emptySet());
```
CanBerker, line 91: This should probably be removeInputVariableDependencies

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+        this.rumbleRuntimeConfiguration = rumbleRuntimeConfiguration;
+    }
+
+    @Override
+    protected Void defaultAction(Node node, Void argument) {
+        for (Node child : node.getChildren()) {
+            visit(child, null);
+            addInputVariableDependencies(node, this.inputVariableDependencies.get(child));
+        }
+        if (!this.inputVariableDependencies.containsKey(node)) {
```
CanBerker, line 160: If I understand correctly, this if block is only triggered when the current node has no children. This logic relies on the implementation of the addInputVariableDependencies() function. As such, I feel it is prone to future bugs.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+        }
+        if (!this.inputVariableDependencies.keySet().contains(node)) {
+            return;
+        }
+        this.inputVariableDependencies.get(node).remove(variable);
+    }
+
+    private void addOutputVariableDependencies(Node node, Set<String> variables) {
+        if (variables == null) {
+            addInputVariableDependencies(node, Collections.emptySet());
```
CanBerker, line 126: Shouldn't this be outputVariableDependencies?

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+        }
+        if (!this.outputVariableDependencies.keySet().contains(node)) {
+            this.outputVariableDependencies.put(node, new TreeSet<String>());
+        }
+        this.outputVariableDependencies.get(node).addAll(variables);
+    }
+
+    private void addOutputVariableDependency(Node node, String variable) {
+        if (variable == null) {
+            addInputVariableDependencies(node, Collections.emptySet());
```
CanBerker, line 137: Shouldn't this be outputVariableDependencies?

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+            }
+            visit(e, null);
+            addInputVariableDependencies(expression, this.inputVariableDependencies.get(e));
+        }
+        return null;
+    }
+
+    @Override
+    public Void visitProlog(Prolog prolog, Void argument) {
+        Map<String, Node> nameToNodeMap = new TreeMap<>();
```
CanBerker, line 379: I would highly recommend extracting the contents of this method to a few functions for readability

src/main/java/org/rumbledb/expressions/module/FunctionDeclaration.java
```diff
@@ -48,6 +50,10 @@ public Expression getExpression() {
         return this.functionExpression;
     }
 
+    public FunctionIdentifier getFunctionIdentifier() {
+        return new FunctionIdentifier(this.functionExpression.getName(), this.functionExpression.getParams().size());
```
CanBerker, line 54: How about creating this identifier in the constructor and storing it in a field?  This would avoid the repeated object creation at each call.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+        }
+        if (!this.inputVariableDependencies.keySet().contains(node)) {
+            this.inputVariableDependencies.put(node, new TreeSet<String>());
+        }
+        this.inputVariableDependencies.get(node).add(variable);
+    }
+
+    private void removeInputVariableDependency(Node node, String variable) {
+        if (variable == null) {
+            addInputVariableDependencies(node, Collections.emptySet());
```
CanBerker, line 115: Is this 'add' here intentional to initialize the node with an empty set of dependencies?

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,466 @@
...
+            return;
+        }
+        if (!this.inputVariableDependencies.keySet().contains(node)) {
+            return;
+        }
+        for (String v : variables) {
+            this.inputVariableDependencies.get(node).remove(v);
+        }
+    }
```
CanBerker, line 100: I had a few comments in this section going back and forth with certain misunderstandings. They are removed now but you have the notifications in your inbox. I have collected my results in this comment.

After checking the usage of variables in visitProlog, I have noticed that you are using the null dependencies of a node to throw a developer safeguard exceptions against non-visited cases. I am fond of this practise to increase developer efficiency and reduce bugs. However, in the current state, I believe the logic of initializing dependencies to an emptySet and updating(add/remove) them are bit too entangled. If I am interpreting this correctly, it reminds me of our discussion regarding the staticContextVisitor and executionMode usage where my initial implementation was similarly entangled in terms of initialization and updates.

How about we implement separate init methods that simply set an emptySet as dependencies for a node. The add/remove methods expect the lists to be initialized and no checks are needed since having a non-initialized node would create NullExceptions during development. 

The add remove methods can now be simplified as: 
- if variables argument is null, we can simply return
- no need to check the dependencies maps for the existence of 'node'
- simply perform the add or remove operation with parameters
- non-existence of node in the map should cause a null pointer exception which we can catch as an OurBadException as dependencies are not initialized. As such, we can remove the null-checks from where you evaluate the dependencies in visitProlog and simplify the code there.

src/main/java/org/rumbledb/expressions/module/Prolog.java
ghislainfourny: This is because the order in this list is important, and it may alternate between variables and functions.

Example:

1. declare variable $x := 1;
2. declare function fn($i) { $i + $x };
3. declare variable $y := fn(2);

The variable dependency visitor will construct a DAG with all declarations, and then re-orders this list in a way compatible with the dependencies, so that the dynamic context are built in the correct order.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: Very useful comments. I will address them. Thank you!

src/main/java/org/rumbledb/api/Rumble.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: Added a long explanation

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: DONE, I added getters.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: Yes, but this is now gone because of the getters.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: Yes, but this is now gone because of the getters.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: It was, but is no longer needed with the getter.

src/main/java/org/rumbledb/expressions/module/FunctionDeclaration.java
ghislainfourny: done, in the InlineFunctionExpression class.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: With getters, the OurBadExceptions were now removed.

src/main/java/org/rumbledb/expressions/module/Prolog.java
CanBerker: > This is because the order in this list is important, and it may alternate between variables and functions.

This was exactly what I needed. Thanks for the detailed explanation!

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
CanBerker: I'm glad to hear that. I will go over the changes now

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,527 @@
...
+
+/**
+ * This visitor resolves dependencies between variable and function declarations.
+ * 
+ * If a variable $x depends on a variable $y, then $y must be evaluated before $x.
+ * 
+ * Example:
+ * 
+ * declare variable $x := 1;
+ * declare variable $y := $x;
```
CanBerker, line 77: I think this example does not match the description given above as y appears to be dependent on x and not the other way around.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,527 @@
...
+ * 
+ * declare function f() { 1 };
+ * declare variable $x := f();
+ * 
+ * If a function f depends on a variable $x, then $x must be evaluated before f's closure is built.
+ *
+ * declare variable $x := 1;
+ * declare function f() { $x };
+ * 
+ * Note that a function cannot depend of a function, as mutually recursive calls are allowed.
```
CanBerker, line 89: cannot depend `on` a function

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,527 @@
...
+ * 
+ * Once all dependencies have been determined, the visited builds a DAG, builds a topological ordering
```
CanBerker, line 91: the `visitor` builds a DAG ?

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,527 @@
...
+    private RumbleRuntimeConfiguration rumbleRuntimeConfiguration;
+
+    /**
+     * Input variable dependencies are lists of variables and functions that an expression depends on.
+     */
+    Map<Node, Set<String>> inputVariableDependencies;
+    /**
+     * Output variable dependencies are lists of variables in the tuples that a clause produces.
+     */
+    Map<Node, Set<String>> outputVariableDependencies;
```
CanBerker, line 106: I really like the clarity you have added here.

If this variable is indeed used 'only' for clauses, how about we specify this relationship in the variable name instead of a comment?

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: Renamed

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,527 @@
...
+            getOutputVariableDependencies(expression.getPreviousClause())
+        );
+        return null;
+    }
+
+    public Void visitCountClause(CountClause expression, Void argument) {
+        visit(expression.getPreviousClause(), null);
+        addOutputVariableDependencies(expression, getOutputVariableDependencies(expression.getPreviousClause()));
+
+        addInputVariableDependencies(expression, new HashSet<String>());
```
CanBerker, line 306: Does this line do anything? Passing an empty set to addInputVariableDependencies simply returns.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,527 @@
...
+
+    @Override
+    public Void visitVariableDeclaration(VariableDeclaration expression, Void argument) {
+        if (expression.getExpression() != null) {
+            visit(expression.getExpression(), null);
+            addInputVariableDependencies(expression, getInputVariableDependencies(expression.getExpression()));
+            System.out.println(String.join(", ", getInputVariableDependencies(expression)));
+            return null;
+        }
+        addInputVariableDependencies(expression, new HashSet<String>());
```
CanBerker, line 516: Similarly, adding an empty set as dependencies.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: Nice catch!

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,527 @@
...
+    public Void visitProlog(Prolog prolog, Void argument) {
+        Map<String, Node> nameToNodeMap = buildNameToNodeMap(prolog);
+        DirectedAcyclicGraph<Node, DefaultEdge> dependencyGraph = buildDependencyGraph(nameToNodeMap, prolog);
+        List<Node> resolvedList = new ArrayList<>();
+        Iterator<Node> iterator = dependencyGraph.iterator();
+        while (iterator.hasNext()) {
+            Node nextDeclaration = iterator.next();
+            resolvedList.add(nextDeclaration);
+        }
+        prolog.setDeclarations(resolvedList);
```
CanBerker, line 502: This is beautiful, great work !

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -0,0 +1,524 @@
...
+     * @param rumbleRuntimeConfiguration the configuration. This is used for trigerring or not debug output.
+     */
+    VariableDependenciesVisitor(RumbleRuntimeConfiguration rumbleRuntimeConfiguration) {
+        this.outputVariableDependenciesForClauses = new HashMap<>();
+        this.inputVariableDependencies = new HashMap<>();
+        this.rumbleRuntimeConfiguration = rumbleRuntimeConfiguration;
+    }
+
+    private void addInputVariableDependencies(Node node, Set<String> variables) {
+        if (variables == null || variables.isEmpty()) {
```
CanBerker, line 119: The code looks great as it is. Thanks a lot for addressing the mentioned issues. 

I have one last suggestion to really push the implementation safety to its limits. We have ensured that getter methods always return emptySet for null cases. Most of the add method calls receive values from the said getter methods. If I'm not missing a case, the remaining add methods are called with variable names extracted from expressions for which we know for sure that the variable name exists and is non-null. 

As such, I feel that we can permit the usage of null parameters to the add methods as in the current ecosystem, such an add call should not happen. In the future, this could hide certain bugs by simply being too lenient for the null parameter case. How would you feel about throwing an 'OurBadException' if null parameter is observed in the add methods.

ghislainfourny:
Thanks! I'll merge and then add guards for nulls as suggested.

PR #653, opened 2020-05-01
Implement colored outputs

CanBerker:
@ghislainfourny I had a visually customized Rumble version running during my thesis demo. I wanted to share the colored output customization by removing the other demo-specific bits.

If you think this would be nice feature we can tweak the shell's look to your liking and add this to the master. 

Default look:
image

Custom look at the moment:
image


ghislainfourny:
This is wonderful, thank you!

PR #655, opened 2020-05-01
Improve getGroupProjectSQL

### LINKED ISSUE #652: Question about group by clause in DataFrames
Currently, `FlworDataFrameUtils.getGroupbyProjectSQL` contains the following [lines](https://github.com/RumbleDB/rumble/blob/fff8f70238b864105f743f0f8963f8fed0536665/src/main/java/org/rumbledb/runtime/flwor/FlworDataFrameUtils.java#L307):

```java
if (groupingKey) {
    queryColumnString.append("array(");
    queryColumnString.append("first(`");
} else {
    queryColumnString.append("collect_list(`");
}
```

If I am not completely wrong, this means that the grouping keys are aggregated with the `first` function and the result is put into an array. Both are weird: First, the grouping keys aren't groups; there is only one of them, so why "aggregate" them? `first` on one value will just return that value. Second, since `first` always returns one value, why put that into an array?

Since everything works, this is problably not a bug. I have tried commenting out these two lines and things still seem to work. Overall, I don't really understand what is happening. If these two function calls are not necessary (as I would guess), I think it would be better to remove them: they are confusing at least and may have a negative performance impact at worst.

CanBerker:
#652 Made it clear that code here is way too obfuscated. I tried to refactor this and make it cleaner.

Fixes #652

### DIFF
--- a/src/main/java/org/rumbledb/runtime/flwor/FlworDataFrameUtils.java
+++ b/src/main/java/org/rumbledb/runtime/flwor/FlworDataFrameUtils.java
@@ -257,11 +257,11 @@ public class FlworDataFrameUtils {
     /**
      * @param inputSchema schema specifies the columns to be used in the query
      * @param duplicateVariableIndex enables skipping a variable
-     * @param trailingComma boolean field to have a trailing comma
-     * @param serializerUdfName
-     * @param groupbyVariableNames
-     * @param dependencies
-     * @param columnNamesByType
+     * @param trailingComma field to have a trailing comma
+     * @param serializerUdfName name of the serializer function
+     * @param groupbyVariableNames names of group by variables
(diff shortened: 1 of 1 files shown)


ingomueller-net:
Can, thanks a lot for the prompt reply and the PR. I don't know whether this is "the Java way", but in other languages I would implement this with a [formatter](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html). This way, you only need to test *once* how to treat each column, set the formatter there, and feed the same formatter in all cases with all possibly needed parameters. Now, you need to do two tests to close the parentheses...

ghislainfourny:
@CanBerker I like it, it is easier to read that way. Thanks!

PR #656, opened 2020-05-03
Parallelized simple map expression

CodingKaiser:
Implementation is working, but seems inefficient.

```
count((1 to 10000000) ! ($$ + 2))
The query took 3913 milliseconds to execute.
```

```
count(parallelize(1 to 10000000) ! ($$ + 2))
The query took 16719 milliseconds to execute.
```


src/main/java/org/rumbledb/runtime/flwor/expression/SimpleMapExpressionIterator.java
```diff
@@ -59,34 +60,54 @@ public SimpleMapExpressionIterator(
...
-        }
-        throw new IteratorFlowException("Invalid next() call in simple map expression", getMetadata());
+    public JavaRDD<Item> getRDDAux(DynamicContext dynamicContext) {
+        JavaRDD<Item> childRDD = this.children.get(0).getRDD(dynamicContext);
+        return childRDD.flatMap((item) -> {
+            List<Item> currentItems = new ArrayList<>();
+            dynamicContext.addVariableValue("$$", currentItems);
+            currentItems.add(item);
+            List<Item> mapValuesRaw = this.rightIterator.materialize(dynamicContext);
+            dynamicContext.removeVariable("$$");
```
ghislainfourny, line 70: This might be working by accident, as this concurrently modifies the same dynamicContext object. It would be best, to have full control, to create a separate closure class as it done everywhere else. There, you can create a DynamicContext field in the closure that is chained with the (same, unchanged) original context, and use it instead.

CodingKaiser: I've implemented the change.

src/main/java/org/rumbledb/runtime/flwor/expression/SimpleMapExpressionClosure.java
```diff
@@ -0,0 +1,54 @@
...
+public class SimpleMapExpressionClosure implements FlatMapFunction<Item, Item> {
+
+
+    private static final long serialVersionUID = 1L;
+    private final RuntimeIterator rightIterator;
+    private final DynamicContext dynamicContext;
+
+    public SimpleMapExpressionClosure(RuntimeIterator rightIterator, DynamicContext dynamicContext) {
+        this.rightIterator = rightIterator;
+        this.dynamicContext = dynamicContext;
```
ghislainfourny, line 41: this should read here

this.dynamicContext = new DynamicContext(dynamicContext)

that way, you can remove line 47, making it considerably faster as you only have one object creation per thread instead of per call.

CodingKaiser: The speed-up given an input sequence of 10.000.000 looks to be about 1 second. I should also mention that I took the original logic from `PredicateIteratorClosure.java`, which looks like the following:

```
    @Override
    public Boolean call(Item v1) throws Exception {
        List<Item> currentItems = new ArrayList<>();
        currentItems.add(v1);
        DynamicContext dynamicContext = new DynamicContext(this.dynamicContext);
        dynamicContext.addVariableValue("$$", currentItems);

        this.expression.open(dynamicContext);
        boolean result = RuntimeIterator.getEffectiveBooleanValue(this.expression);
        this.expression.close();
        return result;

    }
```
Should we change the implementation of this as well to match your suggestions for this closure?

Issue #596, opened 2020-03-26
Rumble output should not be truncated when executed on files


CanBerker:
Example Query:
 ```
spark-submit --master local[*] --conf "spark.driver.host"="127.0.0.1" rumbleml/spark-rumble-1.4-jar-with-dependencies.jar --query-path ./rumbleml/rumbleml-kmeans-libsvm.jq --output-path ./results/rumbleml-kmeans-libsvm-result.json --log-path ./results/rumbleml-kmeans-libsvm-log.txt
```
The following truncation should not happen
>Results have been truncated to:200 items. This value can be configured with the --result-size parameter at startup

**Comment by ghislainfourny** [comment]
@CanBerker could you share your query, and tell me if it still happens with the latest version of the master? Thanks!

**Comment by CanBerker** [comment]
Unfortunately it still occurs. My example queries are given below but it should happen with any query:
```
let $data := libsvm-file("./data/covtype-2500.scale")
let $estimator := get-estimator("KMeans")
let $transformer := $estimator(
    $data,
    {"k": 2, "seed": 1}
)
let $result := $transformer(
    $data,
    { }
)
return $result
```

```
let $data := (
    let $rdd := json-file("./data/IMDB_reviews-2500.json")
    for $entry in $rdd
    let $edited := {
        "movie_id": $entry.movie_id,
        "user_id": $entry.user_id,
        "is_spoiler": $entry.is_spoiler,
        "rating": $entry.rating cast as integer
    }
    return $edited
)

let $df := annotate(
    $data,
    {
        "movie_id": "string",
        "user_id": "string",
        "is_spoiler": "boolean",
        "rating": "decimal"
    }
)

let $estimator := get-estimator("KMeans")
let $transformer := $estimator(
    $df,
    {"featuresCol": ["rating"], "k": 3, "seed": 1}
)

let $result := $transformer(
    $df,
    {"featuresCol": ["rating"]}
)
return $result

```

**Comment by ghislainfourny** [comment]
@CanBerker I see, it's because of the let's. This is a known issue: #615

**Comment by CanBerker** [comment]
I see, feel free to close this issue then

**Comment by ghislainfourny** [comment]
Fixed in PR #627

**Comment by ghislainfourny** [comment]
@CanBerker is it now fixed in the latest master?

**Comment by CanBerker** [comment]
Here is the current behavior:
Query:
```
let $data := libsvm-file("./data/covtype-100k.scale")
let $estimator := get-estimator("KMeans")
[...]

PR #660, opened 2020-05-06
Server

ghislainfourny:


src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,145 @@
...
+                configuration.getQueryPath(),
+                configuration.getOutputPath()
+            );
+
+            Item output = ItemFactory.getInstance().createObjectItem();
+            if (items != null) {
+                Item values = ItemFactory.getInstance().createArrayItem(items);
+                output.putItemByKey("values", values);
+            }
+            output.putItemByKey("status", ItemFactory.getInstance().createIntegerItem(200));
```
CanBerker, line 62: I would like to recommend extracting the HTTP status codes to static variables. While 200 and 500 are well known, if we add more codes in the future, these magic values may become obfuscated.

ghislainfourny: DONE, thanks! Have you tested it?

    $spark-submit target/spark-rumble-1.6.0-jar-with-dependencies.jar --server yes
    $curl --data "parallelize(1 to 10)[2]" -X POST "http://localhost:8000/jsoniq"

CanBerker: Just tested it. Works beautifully 👍  Nicely done

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -202,21 +201,12 @@ private String getIteratorOutput(RuntimeIterator iterator, DynamicContext dynami
                 );
             }
             // remove last comma
```
CanBerker, line 203: This comment is now outdated

src/main/java/org/rumbledb/items/ObjectItem.java
```diff
@@ -116,13 +118,9 @@ public Item getItemByKey(String s) {
...
-        this.values.replaceAll(item -> {
-            if (this.values.indexOf(item) == this.keys.indexOf(s)) {
-                return value;
-            } else {
-                return item;
-            }
-        });
+        this.keys.add(s);
+        this.values.add(value);
+        checkForDuplicateKeys(this.keys, new ExceptionMetadata(0, 0));
```
CanBerker, line 123: Is ExceptionMetadata for line 0:0 intentional here?

src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,203 @@
...
+import org.rumbledb.exceptions.RumbleException;
+import org.rumbledb.items.ItemFactory;
+
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+
+import sparksoniq.spark.SparkSessionManager;
+
+@SuppressWarnings("restriction")
+public class RumbleHandler implements HttpHandler {
```
CanBerker, line 29: I think I would strongly prefer RumbleHttpHandler as a name for this class. RumbleHandler feels very vague

src/main/java/org/rumbledb/server/RumbleHandler.java
CanBerker: `Line 126: ... createIntegerItem(200)`

Can we use the StatusCode Enum here as well?

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
```diff
@@ -94,15 +95,15 @@ public void runQuery(String queryFile, String outputPath) throws IOException {
 
         if (result.isRDD() && outputPath != null) {
             JavaRDD<Item> rdd = result.getRDD(dynamicContext);
-            JavaRDD<String> output = rdd.map(o -> o.serialize());
-            output.saveAsTextFile(outputPath);
+            JavaRDD<String> outputRDD = rdd.map(o -> o.serialize());
+            outputRDD.saveAsTextFile(outputPath);
         } else {
-            String output = getIteratorOutput(result, dynamicContext);
+            outputList = getIteratorOutput(result, dynamicContext);
```
CanBerker, line 101: If I understand this correctly, getIteratorOutput here will truncate the local results prior to writing the results to the file. I think the issue #596 revolves around this point.

src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+        return args;
+    }
+
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
+        try {
+            URI uri = exchange.getRequestURI();
+            String queryString = uri.getQuery();
+            String args[] = getCLIArguments(queryString);
```
CanBerker, line 84: IntelliJ just gave me a heads up that `String args[]` is a C-Style array declaration.

Java Style is as follows: `String[] args`

src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,203 @@
...
+                items = translator.runQuery(
+                    configuration.getQueryPath(),
+                    configuration.getOutputPath()
+                );
+            } else {
+                if (configuration.getOutputPath() != null && !exchange.getRequestMethod().equals("POST")) {
+                    throw new CliException(
+                            "The POST method must be used if specifying an output path, as this has side effects."
+                    );
+                }
```
CanBerker, line 106: This exception seems duplicated in the if and else statements. Can we take move it outside and remove duplication?

src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,203 @@
...
+                if (!exchange.getRequestMethod().equals("GET") && !exchange.getRequestMethod().equals("POST")) {
+                    throw new CliException("Only the GET and POST methods are supported.");
+                }
```
CanBerker, line 109: Is this exception specific to the else statement or can we move this outside as well?

src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,203 @@
...
+            output.putItemByKey("status", ItemFactory.getInstance().createIntegerItem(200));
+            if (configuration.getOutputPath() != null) {
+                output.putItemByKey(
+                    "output-path",
+                    ItemFactory.getInstance().createStringItem(configuration.getOutputPath())
+                );
+            }
+            if (configuration.getLogPath() != null) {
+                output.putItemByKey("log-path", ItemFactory.getInstance().createStringItem(configuration.getLogPath()));
+            }
```
CanBerker, line 135: I would recommend extracting lines 121-135 to a function such as `prepareOutput(...)` for increased readability.

src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,203 @@
...
+                    output.putItemByKey(
+                        "error-message",
+                        ItemFactory.getInstance()
+                            .createStringItem(
+                                "Unexpected error: "
+                                    + ex.getMessage()
+                                    + " We should investigate this. Please contact us or file an issue on GitHub with your query."
+                            )
+                    );
+                    output.putItemByKey("error-code", ItemFactory.getInstance().createStringItem("RBST0004"));
```
CanBerker, line 178: Can we replace the magic value of 'RBST0004' with `ErrorCode.OurBadErrorCode`

src/main/java/org/rumbledb/server/RumbleHandler.java
```diff
@@ -0,0 +1,203 @@
...
+            }
+            Item output = ItemFactory.getInstance().createObjectItem();
+            output.putItemByKey(
+                "error-message",
+                ItemFactory.getInstance()
+                    .createStringItem(
+                        "Unexpected error. We should investigate this. Please contact us or file an issue on GitHub with your query."
+                    )
+            );
+            output.putItemByKey("error-code", ItemFactory.getInstance().createStringItem("RBST0004"));
```
CanBerker, line 195: Can we replace the magic value of 'RBST0004' with `ErrorCode.OurBadErrorCode`

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
ghislainfourny: DONE

src/main/java/org/rumbledb/items/ObjectItem.java
ghislainfourny: yes, because this is outside of any JSONiq query.

src/main/java/org/rumbledb/server/RumbleHandler.java
ghislainfourny: DONE

src/main/java/org/rumbledb/server/RumbleHandler.java
ghislainfourny: Good catch, done

src/main/java/org/rumbledb/cli/JsoniqQueryExecutor.java
ghislainfourny: Yes, very nice that you opened the issue. It would indeed make sense to not limit materialization if this is the final output. This should be handled separately of the server.

src/main/java/org/rumbledb/server/RumbleHandler.java
ghislainfourny: DONE

src/main/java/org/rumbledb/server/RumbleHandler.java
ghislainfourny: DONE

src/main/java/org/rumbledb/server/RumbleHandler.java
ghislainfourny: This was cleaned to be a 405 response.

src/main/java/org/rumbledb/server/RumbleHandler.java
ghislainfourny: DONE

src/main/java/org/rumbledb/server/RumbleHandler.java
CanBerker: I believe this comment remained unaddressed

PR #670, opened 2020-05-14
Implement a custom maven assembly plugin to reduce jar size

CanBerker:
@ghislainfourny You had asked me to investigate the jar size a while ago. I have added a custom assembly descriptor to cleanly specify which files are to be included in the jar but I have not seen a reduction in size. Since we are including a lot of dependencies as well, these may be playing a role in size as well.

I created this PR so that you can take a look and play around with configurations if you'd like to. Maybe I am missing something.


ghislainfourny:
Thanks, Can! I think the resource datasets are less than I imagined in size, which may be good news. In any case, what you did is really useful because now that we have this XML configuration file, it will make it easier for us to fine tune the parameters of the assembly in the future.

What would also need to be done is to move datasets and queries to src/test/resources, and have src/main/resources only contain the banner file. This is the standard Java layout.

CanBerker:
>I think the resource datasets are less than I imagined in size  

To be precise, src/main/resources/queries is exactly at 336 KB at the moment. I believe the increasing size of the JAR is mostly caused by the increasing number of our dependencies since we are packing those in. I remember that at the time of our initial discussion, you had stated that the jar had become 8 MBs instead of 4. Now the size has grown ~15MBs.

> now that we have this XML configuration file, it will make it easier for us to fine tune the parameters of the assembly in the future.

This is exactly the reason why I wanted to put this PR in place. I'm glad we are on the same page.

@ghislainfourny One important remark is that our custom assembly currently results in a slightly bigger size.
Default assembly: 15764 KB
Custom assembly: 16424 KB

> What would also need to be done is to move datasets and queries to src/test/resources, and have src/main/resources only contain the banner file. This is the standard Java layout.

I will take a look at this now.

PR #671, opened 2020-05-14
Add RumbleML Documentation

CanBerker:
@ghislainfourny Here is a draft for the documentation and the updated RumbleML catalog. I would appreciate any feedback to improve it.


ghislainfourny:
Thanks, Can! It looks great. As this is documentation, we should go ahead and publish. We'll continue to update it as we go as this is not constrained by releases.

PR #678, opened 2020-05-20
Register parameter execution modes.

ghislainfourny:


src/main/java/org/rumbledb/compiler/VisitorHelpers.java
```diff
@@ -53,9 +81,13 @@ public static DynamicContext createDynamicContext(Node node, RumbleRuntimeConfig
 
     private static void setLocalExecutionForUnsetUserDefinedFunctions() {
         try {
-            for (
-                FunctionIdentifier functionIdentifier : Functions
+            List<FunctionIdentifier> unsetFunctionIdentifiers = new ArrayList<>();
```
ghislainfourny, line 84: Avoid a concurrent modification exception.

src/main/java/org/rumbledb/compiler/VisitorHelpers.java
```diff
@@ -20,16 +23,33 @@ public static void resolveDependencies(Node node, RumbleRuntimeConfiguration con
...
+
+    public static void populateStaticContext(Node node, RumbleRuntimeConfiguration conf) {
+        printTree(node, conf);
         StaticContextVisitor visitor = new StaticContextVisitor();
         visitor.visit(node, null);
 
+
         visitor.setVisitorConfig(VisitorConfig.staticContextVisitorIntermediatePassConfig);
         int prevUnsetCount = Functions.getUserDefinedFunctionIdentifiersWithUnsetExecutionModes().size();
+        printTree(node, conf);
```
CanBerker, line 45: I think the printTree() function should do one thing that is to actually print the tree. I believe moving `if (conf.isPrintIteratorTree()) {` check into that function does more harm than good in terms of readability. As such, the code has the illusion that it always prints the iterator tree and only when you check the printTree() function you get the idea that it is in fact not printed thanks to the configuration.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
```diff
@@ -102,18 +112,31 @@ public StaticContext visitFunctionDeclaration(FunctionDeclaration declaration, S
...
+                mode = ExecutionMode.LOCAL;
+            }
+            functionDeclarationContext.addVariable(
+                name,
+                expression.getParams().get(name),
+                expression.getMetadata(),
+                mode
             );
+            ++i;
+        }
```
CanBerker, line 135: Is it possible to extract this code block to a function?

src/main/java/org/rumbledb/compiler/VisitorConfig.java
```diff
@@ -8,11 +8,14 @@
     private boolean suppressErrorsForCallingMissingFunctions;
     // flag to suppress errors when an unset execution mode value is fetched from a node
     private boolean suppressErrorsForAccessingUnsetExecutionModes;
+    // flag to suppress errors when an unset execution mode value is fetched from a node
```
CanBerker, line 11: Copy pasted comment. Please remove or update

src/main/java/org/rumbledb/compiler/VisitorHelpers.java
CanBerker: Very nice catch!

src/main/java/org/rumbledb/compiler/VisitorConfig.java
```diff
@@ -91,12 +108,14 @@ public boolean suppressErrorsForAccessingUnsetExecutionModes() {
         .withsuppressErrorsForFunctionSignatureCollision(true)
         .withSuppressErrorsForCallingMissingFunctions(false)
         .withSuppressErrorsForAccessingUnsetExecutionModes(false)
+        .withSetUnsetToLocal(true)
```
CanBerker, line 111: Previously, we would set all unset function identifiers to local(without using a visitor) to solve deadlocks and then make a final visitor pass.

I wanted to clarify the current state after seeing this new flag `setUnsetToLocal`:
It seems that this flag is only used to set unset variable references to local. If this is the case I would recommend updating the name to indicate this. 

Furthermore, (I feel my brain is a bit bloated with information so I may be missing some important details while asking the following two questions):
- Do we still set the functions from Unset to Local "prior" to the final pass? Or do we do it during the final pass?
- In what scenerio do variables require a forced execution more from unset to local?

src/main/java/org/rumbledb/compiler/VisitorHelpers.java
ghislainfourny: Thanks!

src/main/java/org/rumbledb/compiler/VisitorHelpers.java
ghislainfourny: Good catch, done.

src/main/java/org/rumbledb/compiler/StaticContextVisitor.java
ghislainfourny: Good idea, this is now separated.

src/main/java/org/rumbledb/compiler/VisitorConfig.java
ghislainfourny: done

src/main/java/org/rumbledb/compiler/VisitorConfig.java
ghislainfourny: > I wanted to clarify the current state after seeing this new flag `setUnsetToLocal`:
> It seems that this flag is only used to set unset variable references to local. If this is the case I would recommend updating the name to indicate this.

Yes, that is the case and this is renamed.

ghislainfourny: > * Do we still set the functions from Unset to Local "prior" to the final pass? Or do we do it during the final pass?

Yes, we still do. This code is still here.

ghislainfourny:
Thanks a lot for looking, Can.

There are still many many things that I want to improve in here, so this is a small step that fits in the bigger picture.

CanBerker:
The recent commits look great. I am approving the PR. When you make the tests succeed, feel free to merge.

PR #686, opened 2020-05-28
Function or variable name

ghislainfourny:


src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -396,15 +398,16 @@ public Void visitTypeSwitchExpression(TypeSwitchExpression expression, Void argu
 
     @Override
     public Void visitNamedFunctionRef(NamedFunctionReferenceExpression expression, Void argument) {
-        addInputVariableDependency(expression, "Function " + expression.getIdentifier());
+        // TODO add arity
+        addInputVariableDependency(expression, expression.getIdentifier().getName());
```
ghislainfourny, line 402: I should remember to add arity here

ghislainfourny:
@CanBerker it would be awesome if you could check if I didn't overlook anything. This is in preparation for library module support: function names and variable names actually have namespaces (called QNames in XQuery). This PR only updates all variable and function names to use the new Name class, but there is otherwise no material change. It also makes us compliant with the W3C standard.

For now in all our examples, variable and function names are unprefixed (this is why until know we used String). Unprefixed variable names are in no namespace. Unprefixed function names are in the Rumble namespace.

In the future, users will be able to define namespaces and library modules.

src/main/java/org/rumbledb/context/Name.java
```diff
@@ -0,0 +1,214 @@
...
+    public Name addArityToFunctionName(int arity) {
+        return new Name(this.namespace, this.prefix, this.localName + "#" + arity);
+    }
+
+    /**
+     * Returns the namespace.
+     * 
+     * @return the namespace, or null.
+     */
+    public String getNamespace() {
```
CanBerker, line 113: I think these getter comments can be left out.

src/main/java/org/rumbledb/context/Name.java
```diff
@@ -0,0 +1,214 @@
...
+     * @return the expanded name
+     */
+    public static Name createVariableInNoNamespace(String localName) {
+        return new Name(null, null, localName);
+    }
+
+    /**
+     * Creates an expanded name that has the Rumble namespace. By default, in Rumble, unprefixed
+     * function names live in the Rumble namespace. This namespace is for convenience and includes
+     * all functions in XQuery's fn namespace, JSONiq's jn and jnlib namespaces, as well as any
```
CanBerker, line 84: If jn and jnlib are abbreviations, I don't think I am familiar with them. It could be helpful to give the full name in that case.

src/main/java/org/rumbledb/compiler/DynamicContextVisitor.java
```diff
@@ -71,8 +72,8 @@ public DynamicContext visit(Node node, DynamicContext argument) {
     @Override
     public DynamicContext visitFunctionDeclaration(FunctionDeclaration declaration, DynamicContext argument) {
         InlineFunctionExpression expression = (InlineFunctionExpression) declaration.getExpression();
-        Map<String, SequenceType> paramNameToSequenceTypes = new LinkedHashMap<>();
-        for (Map.Entry<String, SequenceType> paramEntry : expression.getParams().entrySet()) {
+        Map<Name, SequenceType> paramNameToSequenceTypes = new LinkedHashMap<>();
+        for (Map.Entry<Name, SequenceType> paramEntry : expression.getParams().entrySet()) {
```
CanBerker, line 76: I want to ask a question with the hopes of learning from you on the topic of OOP & readability:

I found myself wondering about the potential benefits and drawbacks of subclassing the 'Name' class. I believe creating FunctionName, ParamName, and VariableName subclasses could benefit us with improved readability and type safety. (eg. we could simply call the LinkedHashMap (paramToSequenceTypes since type information would already indicate that the keys are FunctionNames). But I was wondering if you can point out some drawbacks with this extended hierarchy such as performance or maintenance.

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -158,14 +159,15 @@ public Node visitModuleImport(JsoniqParser.ModuleImportContext ctx) {
 
     @Override
     public Node visitFunctionDecl(JsoniqParser.FunctionDeclContext ctx) {
-        String fnName = ctx.fn_name.getText();
-        Map<String, SequenceType> fnParams = new LinkedHashMap<>();
+        // TODO namespace
```
CanBerker, line 162: TODO

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -859,10 +861,13 @@ public Node visitParenthesizedExpr(JsoniqParser.ParenthesizedExprContext ctx) {
...
     public Node visitVarRef(JsoniqParser.VarRefContext ctx) {
-        String name = ctx.name.getText();
+        String localName = ctx.name.getText();
+        String prefix = null;
         if (ctx.ns != null) {
-            name = name + ":" + ctx.ns.getText();
+            prefix = ctx.ns.getText();
         }
+        // TODO resolved namespace
+        Name name = Name.createVariableInNoNamespace(localName);
```
CanBerker, line 870: TODO & Prefix seems unused

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
```diff
@@ -338,14 +340,14 @@ public Void visitPredicateExpression(PredicateExpression expression, Void argume
             expression,
             getInputVariableDependencies(expression.getPredicateExpression())
         );
-        removeInputVariableDependency(expression, "$");
+        removeInputVariableDependency(expression, Name.CONTEXT_ITEM);
```
CanBerker, line 343: Very nice!

src/main/java/org/rumbledb/runtime/quantifiers/QuantifiedExpressionVarIterator.java
```diff
@@ -57,7 +58,7 @@ public QuantifiedExpressionVarIterator(
         // this.sequenceType = sequenceType;
```
CanBerker, line 58: Just noticed this commented out line. Is this on our TODO list ?

src/main/java/sparksoniq/spark/ml/GetEstimatorFunctionIterator.java
```diff
@@ -45,10 +46,14 @@
...
+    public static final List<Name> estimatorFunctionParameterNames = new ArrayList<>(
             Arrays.asList(
-                "estimator-input-f6c87df3-fcba-47c7-a5ff-a1a7553b1cab",
-                "estimator-paramobject-ded8adb9-df6f-42b2-b493-863a421a2754"
+                Name.createVariableInRumbleNamespace(
+                    "estimator-input-f6c87df3-fcba-47c7-a5ff-a1a7553b1cab"
+                ),
+                Name.createVariableInRumbleNamespace(
+                    "estimator-paramobject-ded8adb9-df6f-42b2-b493-863a421a2754"
+                )
```
CanBerker, line 56: I like your use of namespace here 👍

src/main/java/org/rumbledb/compiler/DynamicContextVisitor.java
ghislainfourny: That's an excellent point and good thinking.

There is a benefit of subclassing in terms of code quality. There is a benefit in readability of the consuming code.

There is a drawback in performance as class hierarchies are slower than flat classes.  There is drawback in readability of the producing code as developers need to open several files instead of just one.

In this case, the code in the Name class is executed billions of time, as it is massively used while accessing the dynamic context -> this is why performance considerations win.

In a case where the code is executed rarely (such as client-side or desktop application), the benefits would start to outweigh the drawbacks.

If you are interested in seeing the difference of performance, I have a sandbox here:

https://github.com/ghislainfourny/PerformanceMeasurements

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: This is intentional. I marked the places I need to modify next when I add namespace declaration support. Thanks for the check.

src/main/java/org/rumbledb/context/Name.java
ghislainfourny: Added

src/main/java/org/rumbledb/context/Name.java
ghislainfourny: Done

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
ghislainfourny: That's intentional, it will be used in a follow up PR when we have namespace declarations.

src/main/java/org/rumbledb/compiler/VariableDependenciesVisitor.java
ghislainfourny: :)

src/main/java/org/rumbledb/runtime/quantifiers/QuantifiedExpressionVarIterator.java
ghislainfourny: Yes. It's commented out because it's not used (yet) and so avoids a Java warning. But in the future, quantified expression type checking should be implemented.

ghislainfourny:
> This looks good to me. I just have minor comments
> 
> Edit:
> I have applied `String*.name` regex search to look for any missing name usages. I have found the following case:
> ApplyTransformerRuntimeIterator.java
> ` 66,18: for (String name : this.columnNamesOfGeneratedVectors)`

Thanks. I think this is fine, column names are always indeed strings.

PR #695, opened 2020-06-05
Trace

ghislainfourny:


src/main/java/org/rumbledb/runtime/functions/base/Functions.java
```diff
@@ -2470,5 +2475,17 @@ private static BuiltinFunction createBuiltinFunction(
             AnnotateFunctionIterator.class,
             BuiltinFunction.BuiltinFunctionExecutionMode.DATAFRAME
         );
+
+        /**
+         * function converts given RDD or local data to a DataFrame using a schema
+         */
```
CanBerker, line 2481: I wanted to leave a heads up with the non-updated comment here

ghislainfourny: Good catch, thanks!

PR #697, opened 2020-06-08
Namespace declaration

ghislainfourny:


src/main/java/org/rumbledb/context/StaticContext.java
```diff
@@ -151,4 +152,29 @@ public boolean hasVariable(Name variableName) {
...
+
+    public boolean bindNamespace(String prefix, String namespace) {
+        if (this.namespaceBindings == null) {
+            this.namespaceBindings = new HashMap<>();
+        }
+        if (!this.namespaceBindings.containsKey(prefix)) {
+            this.namespaceBindings.put(prefix, namespace);
+            return true;
+        }
+        return false;
```
ghislainfourny, line 164: TODO: throw error when binding a namespace twice
TODO: support qname in parameter names

PR #699, opened 2020-06-11
Base uri

ghislainfourny:
This branch introduces the static base URI to the static context, and makes the static context available to the runtime iterators.

The static base URI is used to resolve relative paths relative to the current file (it was incorrectly resolved relative to the working directory, which is problematic if running a query from another location on disk). It will also be used to import modules, also relative to the current file.

Most of the lines of code changed are for adapting the paths in all test queries that read from a resource file, to make these paths relative to the query files.

TODO before merge: double-check on Amazon EMR that it doesn't break anything with HDFS there.

@CanBerker since it's rather a significant change, feel free to look at it in case I overlooked anything.


ghislainfourny:
Status update: works wonderfully on Amazon EMR with error messages that are much easier to understand. Waiting for confirmation that it also works on Windows.

PR #706, opened 2020-06-24
Add test case that shows that importing a module repeatedly is broken.

ingomueller-net:
I added a test case which I think should fail currently. I don't know how to run the tests, so I don't know if it actually breaks. I think that this is due to a bug: if I include `moduleA` and `moduleB`, `moduleA` should be allowed to include `moduleB` without me knowing.


ghislainfourny:
Thanks, Ingo! Indeed, that's a limitation of Rumble that I need to address. The current implementation is assuming module dependencies are a tree, not a DAG, but you are absolutely correct that DAGs should be supported in the end.

PR #747, opened 2020-08-14
Empty sequence order

ghislainfourny:


ghislainfourny:
@ingomueller-net one less missing feature (see concretely the tests).

ingomueller-net:
Nice, bravo!

Issue #760, opened 2020-08-31
Allowing empty returns wrong results 


losmi83:
Several issues with allowing empty on the master:

1) Wrong result with local collections:

```
let $stores := [  { "store number" : 1, "state" : "MA" } ]
let $nations := [  { "name": "US", "state": "MA" }, { "name": "US", "state": "CA" } ]
for $nation in $nations[], $store allowing empty in $stores[]
    [ $$.state eq $nation.state ] 
return { "name": $nation.state, "stores": $store."store number" }
```
Output: 
```
{ "name" : "MA", "stores" : 1 }
{ "name" : "CA", "stores" : null }
{ "name" : "CA", "stores" : null }
{ "name" : "CA", "stores" : null }
...
{ "name" : "CA", "stores" : null }
Warning! The output sequence contains 9223372036854775807 items but its materialization was capped at 200 items. This value can be configured with the --result-size parameter at startup
```

2) Infinite loop with local collections:
```
let $stores :=[  { "store number" : 1, "state" : "MA" } ]
let $nations := [  { "name": "US", "state": "MA" }, { "name": "US", "state": "CA" } ]
let $join := 
    for $nation in $nations[], $store allowing empty in $stores[]
        [ $$.state eq $nation.state ] 
    return { "name": $nation.state, "stores": $store."store number" }
return [$join]
```

3) Wrong result with local + distributed collections:
```
let $stores := parallelize([  { "store number" : 1, "state" : "MA" } ])
let $nations := [  { "name": "US", "state": "MA" }, { "name": "US", "state": "CA" } ]
for $nation in $nations[], $store allowing empty in $stores[]
    [ $$.state eq $nation.state ] 
return { "name": $nation.state, "stores": $store."store number" }
```

Output:
```
{ "name" : "MA", "stores" : null }
{ "name" : "CA", "stores" : null }
```
The result is correct, however, when nations is distributed and stores is local.

4) Allowing empty with two distributed collections not supported

```
let $stores := parallelize([  { "store number" : 1, "state" : "MA" } ])
let $nations := parallelize([  { "name": "US", "state": "MA" }, { "name": "US", "state": "CA" } ])
for $nation in $nations[], $store allowing empty in $stores[]
    [ $$.state eq $nation.state ] 
return { "name": $nation.state, "stores": $store."store number" }
```

Output:

```
⚠️  ️There was an error.
A for clause expression cannot produce a big sequence of items for a big number of tuples, as this would lead to a data flow explosion.
[...]

PR #761, opened 2020-08-31
Fix allowing empty.

ghislainfourny:


ghislainfourny:
This fixes #760 points 1, 2 and 3.

PR #844, opened 2020-12-15
Json reader

ghislainfourny:
@CanBerker hopefully this should make the streaming issue go away as well as encoding problems we had. I managed to find an equivalent functionality in gson to directly produce our Items.


CanBerker:
@ghislainfourny Thanks for taking the time to look into this. I would like to briefly summarize my progress on the JsonIter side. I tried the dynamic generation (as indicated by the docs) and reflection solutions of JsonIter today but unfortunately it did not bear fruit. I started investigating the usage of underlying jsonassist library and dynamic code generation of JSONIter before stopping for the day. In any case, I am not sure if this issue can easily be fixed without digging deep into and potentially improving JsonIter's implementation

I like the clean feel of the gson API and it seems you've managed to reproduce behavior with minimal changes.

Since JsonIter boats leading performance, I wanted to check how the performance of GSON fares compares to other similar libraries and found the following resources. I would assume performance would be of importance to you and would highly recommend checking the first link below. For example [dsljson](https://github.com/ngs-doo/dsl-json) on the far right of the graph looks promising.
- [Compares a large number of libraries summarized in a nice graph](https://interviewbubble.com/performance-comparison-of-json-libraries-jackson-vs-gson-vs-fastjson-vs-json-simple-vs-jsonp/)
- [highly biased against gson](https://medium.com/@dannydamsky99/heres-why-you-probably-shouldn-t-be-using-the-gson-library-in-2018-4bed5698b78b)

Lastly, I would be happy to learn more if there is a specific reason why you've chosen the gson alternative. Let me know what you think.

ghislainfourny:
Thanks @CanBerker for the thorough check.

Indeed this deserves more investigation and dsljson looks like the fastest.

I tried to make use of DslJson in branch "DslJson" but didn't manage yet, because the API corresponding to our use case (streaming through JSON not knowing in advance what we'll find) doesn't seem to be well documented, and JsonReader is a lower level of abstraction (compared to gson and jsoniter) with missing functionality for detecting what's next. Maybe I am not looking at the right class -- It would be great if you could take a look if you have time and find an elegant solution.

CanBerker (approved):
LGTM 👍

ghislainfourny:
Thank you @CanBerker !

PR #845, opened 2020-12-16
Dsl json

ghislainfourny:


ghislainfourny:
@CanBerker this is as far as I managed to go, but it still requires more work.

ghislainfourny:
I think it's in a much better state now. I managed.

The only inefficiency left is parsing a top-level number, as it will trigger an IOException at EOF that gets caught, which would be slower.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -62,13 +63,45 @@
...
+                    token == '+'
+                        || token == '-'
+                        || token == '.'
+                        || (token >= '0' && token <= '9')
+                        || token == 'e'
+                        || token == 'E'
+                ) {
+                    sb.append((char) token);
+                    try {
+                        token = object.getNextToken();
```
CanBerker, line 91:
```suggestion
                        if (object.getCurrentIndex() = object.length()) {
                            break;
                        }
                        token = object.getNextToken();
```

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -62,13 +63,45 @@
...
+                    } catch (IOException e) {
+                        if (!isTopLevel) {
+                            RumbleException ex = new ParsingException(
+                                    "Parsing error! , or ] or } expected but found " + (char) token,
+                                    metadata
+                            );
+                            ex.initCause(e);
+                            throw ex;
+                        }
+                        // Here atm this will slow down if this is a top-level, lone number.
```
CanBerker, line 101: Using currentIndex and Length as suggested above, the efficiency problem with top level integers you've mentioned as the first point can be addressed.

I am curious what other conditions can trigger this IOException block here. If the out of bounds problems was the only issue, I believe this try/catch block can be eliminated as well.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -62,13 +63,45 @@
...
+    public static Item getItemFromObject(JsonReader<Object> object, ExceptionMetadata metadata, boolean isTopLevel) {
         try {
-            if (object.whatIsNext().equals(ValueType.STRING)) {
-                return ItemFactory.getInstance().createStringItem(object.readString());
+            if (isTopLevel) {
+                object.getNextToken();
             }
-            if (object.whatIsNext().equals(ValueType.NUMBER)) {
-                String number = object.readNumberAsString();
+            byte token = object.last();
```
CanBerker, line 71: The API offered by dslJson is an "interesting" one. While getNextToken() and last() do the job, they are not very descriptive and can even be misleading for new developers (ie. last() -> gets the mostRecentByte or the last byte in the stream?)

I would be highly in favor of defining our own interface to dslJson which abstract away these details and provide a cleaner API. It would also make our code much more resilient and easy to fix in case something changes in dslJson's API in the future. Or even more, in case we decide to change the supporting library dslJson with something else.

What do you think?

Edit: On a second thought, we are parsing the JSON input character by character with various checks at each turn. I wonder if we are losing the speed benefit against GSON in this process. Not to mention readability. I would be very curious to see the performance comparison of these two implementations.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
ghislainfourny: This is a great finding, Can. Many thanks. I solves the issue and IOException is no longer needed.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
ghislainfourny: Many thanks for your thorough comments.

Regarding the dslJson API: I agree -- the low-level is actually a plus though, because we can support more parsing options (lenient, duplicates, etc) that we couldn't with gson or jsonIter.

Regarding adding an API: we could consider adding our own internal API -- since this is a critical portion of the code that gets executed billions of times though, we need to be careful if we add more interfaces and method calls than there is already, and carefully check the performance impact if we do so.

src/main/java/org/rumbledb/runtime/functions/input/UnparsedTextLinesFunctionIterator.java
```diff
@@ -65,7 +65,7 @@ public UnparsedTextLinesFunctionIterator(
         if (this.children.size() == 1) {
             strings = SparkSessionManager.getInstance()
                 .getJavaSparkContext()
-                .textFile(uri.toString());
+                .textFile(uri.toString(), 10);
```
CanBerker, line 68: Can you extract this minPartitions number to a Constant

src/main/java/org/rumbledb/runtime/functions/io/JsonDocFunctionIterator.java
```diff
@@ -74,10 +79,19 @@ public Item next() {
...
             } catch (IteratorFlowException e) {
-                throw new IteratorFlowException(e.getJSONiqErrorMessage(), getMetadata());
+                RumbleException ex = new IteratorFlowException(e.getJSONiqErrorMessage(), getMetadata());
+                ex.initCause(e);
+            } catch (IOException e) {
+                RumbleException ex = new IteratorFlowException(
+                        "I/O Exception while binding the input stream",
+                        getMetadata()
+                );
+                ex.initCause(e);
```
CanBerker, line 94: The built exceptions here are not thrown

src/main/java/org/rumbledb/items/parsing/ItemParser.java
CanBerker: >Regarding the dslJson API: I agree -- the low-level is actually a plus though, because we can support more parsing options (lenient, duplicates, etc) that we couldn't with gson or jsonIter.

This is beautiful, I had not considered this.

> Regarding adding an API: we could consider adding our own internal API -- since this is a critical portion of the code that gets executed billions of times though, we need to be careful if we add more interfaces and method calls than there is already, and carefully check the performance impact if we do so.

Fair point, I may try to refactor this part in a separate branch after we get the initial implementation working. Performance reports can be generated thereafter

CanBerker (commented):
@ghislainfourny What is the timeline you have in mind for this PR? Are you planning to merge today (with/without address points 2 and 3) or leave remaining points for another PR after the holiday season ?

src/main/java/org/rumbledb/runtime/functions/input/UnparsedTextLinesFunctionIterator.java
ghislainfourny: Thank you for taking care of it, Can!

src/main/java/org/rumbledb/runtime/functions/io/JsonDocFunctionIterator.java
ghislainfourny: Also thank you!

ghislainfourny:
> @ghislainfourny What is the timeline you have in mind for this PR? Are you planning to merge today (with/without address points 2 and 3) or leave remaining points for another PR after the holiday season ?

I have opened an issue here: 

https://github.com/RumbleDB/rumble/issues/848

Did I miss anything?

I think the only prerequisite for merging as it is would be a quick confirmation that q1/q4/q5 are not slower than the current master with JsonIter. Then, we could merge and address the other issues as a follow up.

It is fine if we only merge in January though, no need to do it in a hurry.

CanBerker:
For the points discussed in this PR the issue seems complete. The following can be added as mentioned in the followup PR https://github.com/RumbleDB/rumble/pull/847
- Modifying  DslJson according to Rumble's needs and publishing it to maven
- Consider introducing an internal method that calls ItemParser when JsonInput is received from standard input

PR #847, opened 2020-12-18
feature: DslJSON Implement single top level item assertion for ItemParser

CanBerker:
Dear @ghislainfourny 

I have been investigating the following task you've set and here are my findings:
> finding a way to check whether it's the end (otherwise no error is thrown if there is more after the successfully parsed JSON)

Firstly my understanding is that there can only be one top level Item. Is this correct ?

Road forward with DslJson:
- getNextToken() implementation of DslJson throws if there are whitespace characters at the end of the file with no further characters.
- read() method can be used instead to detect trailing whiteSpaces byte by byte
- DslJson has a private method called wasWhiteSpace() that covers a large number of checks against various whitespace characters (~30 switch statements). I would suggest we modify the library and give this method higher accessibility. The licence of the project seems to be BSD-3 which permit modification (Please also check for yourself in case I'm missing something)
  - **(Actionable)** Verify DslJSON can be modified & published
  - **(Actionable)** Clarify how publishing DslJSON to a (shared)/remote maven repo for Rumble should be handled  

Testing Related: 
- I have added some tests which currently detect deformities in JSON which follow a valid item.
- Unfortunately these tests can't detect whitespaces such as newlines at the moment. I'm guessing this is why ParseJson.jq test fails. This test should hopefully be fixed once whitespace detection is improved.
- Writing tests requires creating a multitude of well-formed and ill-formed JSON files. I was wondering if we should define (unless we already have) a (public or internal) method that reads JSON from standard input and then calls our ItemParser rather or in rather than parsing it to Items via Visitors. This would simplify testing.
  - **(Actionable)** Consider a new method that reads JSON from standard input

PS: I have defined the assertSingleItemAtTopLevel() to ease reviewing at the moment. Once implementation has matured this method can be inlined for performance


src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -202,6 +210,18 @@ public static Item getItemFromObject(JsonReader<Object> object, ExceptionMetadat
         }
     }
 
+    private static void assertSingleItemAtTopLevel(JsonReader<Object> object, boolean isTopLevel)
+            throws java.io.IOException {
+        if (isTopLevel) {
+            while(object.getCurrentIndex() != object.length()) {
+                object.read();
+                if (object.last() != 32) {  // TODO: replace this with wasWhiteSpace() of DslJson::JsonReader
+                    throw new RuntimeException("More than one top level item detected in JSON");
```
ghislainfourny, line 219: This should be a parsing exception. Actually, we should create a new exception JsonParsingException with error code FOJS0001 and use it everywhere else instead of ParsingException, because XPST0003 is normally only for query parsing exceptions and is thrown statically, while FOJS0001 is thrown dynamically.

(see https://www.w3.org/TR/xpath-functions-31/#func-parse-json )

ghislainfourny (changes requested):
Thank you for improving further, Can! Just a comment on error codes.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
ghislainfourny: Also, the code for duplicate keys should be FOJS0003, caught by the ItemParser (not sure if this is what is thrown at the moment, as it is only caught later in the objectitem constructor I believe).

CanBerker: Thanks for catching this as these should definitely be on our to-do list.

The RuntimeException is caught and converted to a ParsingException in the ItemParser at the moment. All ParsingExceptions thrown return XPST0003 I think.

PR #920, opened 2021-02-16
Static typing merge

ghislainfourny:


src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -20,6 +20,20 @@
```
ghislainfourny, line 20: Main update: pass null instead of item* when type unspecified.

PR #942, opened 2021-03-26
Fix: Exception when grouping with keys of different types

elwin:
This PR addresses the comparison issue in group by when items of different types are compared.

I opted to create an additional function `boolean equalFlworKey()` that doesn't throw an exception upon type mismatch and instead returns false.

There were other approaches I considered:
- Have `compareWithFlworKey` not throw exception upon type mismatch and return instead -2: Let callers deal with this case
  - 🚫 Caller looses information about why the comparison failed, can't throw error message with detailed error message anymore (without further refactoring)
- Add a boolean flag `compareWithFlworKey` that determines whether exception should be thrown: If true, throw exception upon type mismatch, otherwise return -2
  - 🚫 Changes signature of public method
  - 🚫 Different return values based on function arguments

Please let me know if you'd prefer a different approach.


elwin:
Merged with #943

PR #1017, opened 2021-05-18
Implementation-independent tests

elwin:
This PR modifies some tests such that they are (a bit more) implementation independent. This issue arose when we used the DecimalGamma encoding for the item key which changed the (undefined) ordering of the results. With this change, the ordering in some test-cases should be implementation independent through the use of ordering.

Only test-cases that failed with the alternative grouping implementation were changed - it is very likely that there are further test-cases that still do have undefined ordering.


src/test/resources/test_files/runtime-spark/DataFrames/GroupbyClause10.jq
```diff
@@ -22,6 +22,7 @@ for $i in parallelize((
 ))
 let $j := flatten($i.a)
 group by $j
+order by serialize($j)
 return {
-  "key" : $j, "count": sum($i.b)
+  "key" : serialize($j), "count": sum($i.b)
```
elwin, line 27: The `serialize` is not strictly necessary. However, without it the output contains two groups with the same key `null` that were ordered differently, which I found confusing at first.

Result without the `serialize`:
<pre>
(
<b>  { "key" : "null", "count" : 4 },</b>
  { "key" : "1", "count" : 2 },
  { "key" : "2", "count" : 1 },
  { "key" : "bar", "count" : 1 },
  { "key" : "false", "count" : 2 },
  { "key" : "foo", "count" : 3 },
<b>  { "key" : "null", "count" : 2 },</b>
  { "key" : "true", "count" : 4 }
)
</pre>

ghislainfourny: This is because one group is for the empty sequence, the other for null.

In order to distinguish them properly in the output, it is common to use an array instead, which preserves the values in the output while making the empty sequence distinguishable from a null:

    { "key" : [ $j ], "count" : sum($i.b) }

ghislainfourny (changes requested):
Looks great! Just a comment on using [] rather than serialize for the group by test.

src/test/resources/test_files/runtime-spark/DataFrames/GroupbyClause10.jq
elwin: Ah this is indeed a nice solution!

Just updated it! :)

PR #1168, opened 2022-02-04
Initial lets

ghislainfourny:


ghislainfourny:
TODO:
check that either there are no group by clauses, or none of the let variables is used after a group by clause.

PR #1201, opened 2022-10-19
Fix StringConcatIterator to throw XPTY0004 for non-singleton operands

thomastzhou:
**On current master:**
Input: `"A" || ("B", "C")`
Expected: Error code XPTY0004 due to non-singleton operands
Actual: `"AB"`

**Changes**
- StringConcatIterator correctly checks for sequences of more than 1 item and throws XPTY0004


mschoeb:
Thanks, this has since been fixed on master so I will close this for now

PR #1247, opened 2023-12-21
Make function static

ghislainfourny:


ghislainfourny:
@mschoeb Now that the master is officially on Spark 3.5 scala 2.13, this branch is no longer needed, but there are two small fixes that should go through.

PR #1249, opened 2024-02-22
Make jQuery includes protocol-aware

darioackermann:
The current version of RumbleDB includes jQuery on the public pages using the http-protocol.
This works under the premise that the public page itself is also browsed HTTP.

However, when using a proxy in front of Rumble which serves content using HTTPS (i.e., every modern nginx proxy app), all modern browsers will refuse to load the script due to MixedContent policy

image

This change makes the inclusion of jQuery protocol aware; it will load jQuery over http for sites served over http, and over https for sites served over https


ghislainfourny:
Thank you very much for this, @darioackermann ! Also for the other two CLs.

PR #1251, opened 2024-02-28
Big performance merge

ghislainfourny:
Copying comments from @mschoeb 

This PR contains some steps towards merging the work of dominik. It includes:

disable OrderByClauseSparkIterator.createOrderExpression(). many errors happen due to
OrderByClauseSparkIterator.createOrderExpression() trying to access the resultingtype which is not set. The NativeClauseContext is created in OrderByClauseSparkIterator.tryNativeQuery without a resultingType. I disabled this new functionality and now it just falls back to not using a native query
In flowr clauses, the last LET clause was often deleted. This happened due to the new ProjectionPushdownVisitor not correctly visiting all let clauses. Now it visits them all
I left in the original codes as references. These two changes decrease the amount of failing tests from 580 to 64 on gitlab


ghislainfourny:
@mschoeb I found the cause why the let clause was deleted. The visitor for the return clause defaulted to the default action, which caused the let clause to be visited twice. This should now be fixed with the deletion of unused let clauses re-activated.

ghislainfourny:
All tests now pass. For the remaining issue with the order by that is currently deactivated (so, not triggering native execution at this point), I created a separated branch.

PR #1252, opened 2024-06-05
Adding scripting functionality to RumbleDB

DavidBuzatu-Marian:
The PR introduces scripting capabilities within the RumbleDB engine. Scripting enables support for side-effects, both in query bodies and functions.
Changes are backwards compatible with previous RumbleDB versions.
Scripting introduces the construct of statements, which are execution units that do not return anything and execute locally. Local execution is enforced as side-effects may affect parallel computation where cross-dependencies exist between parallel jobs.
With scripting, we also enabled `while` loops, as well as interrupts such as `break loop`, `continue loop`, and `exit returning`.


ghislainfourny (approved):
This is amazing work!

PR #1269, opened 2024-10-24
Update install.md

ghislainfourny:


MSthe00:
We can close this right? There is a different commit but with the identical changes on master already

PR #1272, opened 2024-10-28
Spark 3.5 (Scala 2.12)

ghislainfourny:


ghislainfourny:
We no longer support scala 2.12 from now on.

PR #1304, opened 2025-03-05
Add issue templates

Byczax:
Hello,
I created templates for the future reported issues, so we will have constant way of reporting, so it's much easier to find needed information and compare issues for possible duplicates.


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,78 @@
...
+        # paste output here
+
+        # On Windows:
+        C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
+        # paste output here
+        ```
+
+        </details>
+    validations:
+      required: true
```
mschoeb, line 56:
```suggestion
      required: false
```
I still think we can make this optional, since its not related to many issues and makes issue creation easier. Its good to have for things like terminal behaviour, but not needed for things like spec edgecase not correctly handled (which is evaluated on gitlab cicd where we dont even know this info)

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,78 @@
...
+        # paste output here
+        $ ant -version
+        # paste output here
+        $ spark-submit --version
+        # paste output here
+        ```
+
+        </details>
+    validations:
+      required: true
```
mschoeb, line 78:
```suggestion
      required: false
```
same

.github/ISSUE_TEMPLATE/enchantment.yaml
mschoeb, line 1: the file is called enchantment, shouldnt it be enhancement?

mschoeb (commented):
Everything else LGTM 🚀

PR #1394, opened 2025-11-03
Simplify code.

ghislainfourny:


EPMatt (approved):
LGTM, with a small question: is commenting output lines like this intentional?

`// System.err.println(AnnotationsTestsBase.counter++ + " : " + this.testFile);`

If that is intentional, we could remove the lines all together, and also drop the `counter` property from RumbleDBTestCommons, as it looks like it's not used / updated anywhere else.

PR #1419, opened 2026-03-18
Update versions in pom.xml.

ghislainfourny:


EPMatt (approved):
LGTM. Tested with `mvn clean compile assembly:single`

ghislainfourny:
> LGTM. Tested with `mvn clean compile assembly:single`

Thank you!

PR #1430, opened 2026-04-17
Add full support for format-number and decimal-format declarations from prolog through static context

ghislainfourny:
Add full support for format-number and decimal-format declarations from prolog through static context


ghislainfourny:
@EPMatt would it be possible to check if there are any regressions in the qt3 tests against the master branch? There were a few merge conflicts that I solved and I would like to make sure I didn't break anything. Thanks!

@HenrikPaetzold could you please check if in the qt3 function tests the decimal format prologs were added successfully based on passing tests? Otherwise it means that test suite workflow does not take yet into account your latest change.

HenrikPaetzold:
When checking out the rumble repository at this pull request (latest commit `aa7c8eb`) and using the latest master commit of the rumble-test-suite to run `Fn1Test` locally, only `numberformat121` and `numberformat122` fail due to a stack overflow. This is expected — I believe I already mentioned this in Thursday’s meeting.

However, in the pipeline, all `format-number` tests that define a decimal format in the test environment are failing. Probably to the prolog not being added to the query. Additionally, if I checked correctly, all `codepoints-to-string` tests that require a different XML version are also failing. This would probably mean, that the pipeline is not yet using the latest version of the rumble-test-suite, right? 
@ghislainfourny

ghislainfourny:
@HenrikPaetzold thank you for checking, this is great. Yes that's probably it. Merging into rumble-test-suite is currently not enough to get changes to work. Let us wait for @EPMatt 's input on this.

EPMatt:
@ghislainfourny thanks for pinging me. We had some unmerged features in my fork, that allow to run serialization tests in the test suite. Please see https://github.com/RumbleDB/rumble-test-suite/pull/5

PR #1432, opened 2026-04-19
feat: master's thesis final improvements

EPMatt:
This PR includes a few final improvements related to the Master's thesis work, as previously discussed.

## Changelog
- feat(item): store node names as Name instances, not QNameItem instances. This optimizes performance for large number of items
- fix(runtime): separate JSONiq and XQuery postfix and unary lookup runtime iterators, so that semantics match what is defined for the two languages. In JSONiq, accessing an array position with a wrong index returns the empty sequence, whereas in XQuery 3.1 raises a runtime exception.


ghislainfourny (approved):
Great work!

PR #1456, opened 2026-05-05
Support for json-doc#2

ghislainfourny:
This PR adds support for fn:json-doc option maps, including handling for liberal, duplicates, escape, and fallback.
It introduces a dedicated JSON parser to support stricter error handling, duplicate-key policies, liberal JSON extensions and fallback behavior for invalid XML characters.
Main changes

Added 2-argument variants of fn:parse-json and fn:json-doc

Implemented JSON parsing options:

liberal
duplicates
escape
fallback

Added JSON-specific exceptions and error codes
Added handling of JSON null for XQuery mode or JSONiq mode
Added XML codepoint validation utility
Updated default XML version handling to use XMLUtils


src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -38,11 +38,9 @@
 import java.time.ZoneId;
 import java.time.OffsetDateTime;
 import org.rumbledb.api.Item;
-import org.rumbledb.exceptions.ExceptionMetadata;
-import org.rumbledb.exceptions.OurBadException;
-import org.rumbledb.exceptions.ParsingException;
-import org.rumbledb.exceptions.RumbleException;
+import org.rumbledb.exceptions.*;
```
ghislainfourny: This needs to be expanded to an explicit list of packages.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -64,10 +62,7 @@
 import java.sql.Date;
 import java.sql.Timestamp;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
ghislainfourny: This needs to be expanded to an explicit list of packages.

src/main/java/org/rumbledb/runtime/functions/io/JsonDocFunctionIterator.java
```diff
@@ -49,38 +30,134 @@ public JsonDocFunctionIterator(
...
-                );
+                )
+            ) {
                 JsonReader object = new JsonReader(new InputStreamReader(is));
-                return ItemParser.getItemFromObject(object, getMetadata());
-            } catch (IteratorFlowException e) {
-                throw new IteratorFlowException(e.getJSONiqErrorMessage(), getMetadata());
+                return ItemParser.getItemFromObject(object, isJSONiq, getMetadata());
+            } catch (CannotRetrieveResourceException e) {
+                throw new UnavailableResourceException(e.getMessage(), getMetadata());
```
ghislainfourny: We need to chain exceptions to make debugging easier for users. After creating the new exception, you can call initCause() to pass the old one. This applies to all other places.

src/main/java/org/rumbledb/runtime/functions/json/JSONParsingOptions.java
```diff
@@ -0,0 +1,286 @@
...
+                }
+            }
+
+            Item argument = ItemFactory.getInstance().createStringItem(escapedSequence);
+            variableValues.addVariableValue(
+                parameterNames.get(0),
+                Collections.singletonList(argument)
+            );
+
+            List<Item> result = functionItem.getBodyIterator().materialize(functionContext);
```
ghislainfourny: there should be a more adequate materialize function called materializeAtMostOneItemOrNull that also checks there is at most one item.

All you need to do is catch MoreThanOneItemException.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
```diff
@@ -84,63 +79,156 @@ public class ItemParser implements Serializable {
...
             if (object.peek() == JsonToken.NUMBER) {
                 String number = object.nextString();
-                if (number.contains("E") || number.contains("e")) {
-                    return ItemFactory.getInstance().createDoubleItem(Double.parseDouble(number));
-                }
-                if (number.contains(".")) {
-                    return ItemFactory.getInstance().createDecimalItem(new BigDecimal(number));
-                }
-                return ItemFactory.getInstance().createIntegerItem(number);
+                return ItemFactory.getInstance().createDoubleItem(Double.parseDouble(number));
```
ghislainfourny: Please do not throw the code away; there is a new option "number-format" in XQuery 4 that allows to choose and we can make "adaptive" the default for JSONiq 1.0 to be backwards compatible.
https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-parse-json

Thus, you can here make the behavior depend on the number-format option, and make the default value of number-format depend on whether it is JSONiq 1.0 or not.

No need to worry about differentiating XQuery 3.1 and XQuery 4.0, we will simply publicly support the latest version.

The third possible behavior for number-format is to parse all numbers to a decimal item , which should be easy to support too.

src/main/java/org/rumbledb/items/parsing/ItemParser.java
HenrikPaetzold: This was done automatically, either by spotless or IntelliJ. Has happened a few times now...

src/main/java/org/rumbledb/items/parsing/ItemParser.java
HenrikPaetzold: done

src/main/java/org/rumbledb/runtime/functions/json/JSONParsingOptions.java
HenrikPaetzold: done

src/main/java/org/rumbledb/runtime/functions/io/JsonDocFunctionIterator.java
HenrikPaetzold: done

src/main/java/org/rumbledb/items/parsing/ItemParser.java
HenrikPaetzold: It was the default IntelliJ IDEA option that replaces imports with wildcards once the number of imports from the same package exceeds five. Spotless was not the issue.

PR #1468, opened 2026-05-10
Refactor HTML pages to remove jQuery external dependency

Paul-Gy:
While using RumbleDB locally for simple data processing, I noticed the web UI stops working entirely when offline, clicking the buttons does nothing. The issue is caused because the HTML pages silently depend on jQuery from an external CDN.

To avoid this issue, replaced jQuery's `$.post()` with native `fetch()` (which is now [fully supported](https://caniuse.com/fetch)). Also included a few small robustness improvements along the way: network errors are now shown in the results area instead of silently ignored; the submit button is disabled during a request to prevent duplicate submissions.

The design is basically the same as before with minor alignement corrections, but I’m open to a redesign, either in this PR or in a follow-up one, if there’s interest.

I also noticed a few HTML schema issues reported by the official [World Wide Web Consortium (W3C) validator](https://validator.w3.org/nu/), which are now fixed :)

Tested with RumbleDB 2.1.0 latest release and is working fine!


ghislainfourny (approved):
Thanks a lot for the fix, Paul!
The button labels should be bumped to 2.1.0.
Otherwise, this is good to go. If you want to further improve the cosmetics of the pages, do not hesitate to make a proposal.

Paul-Gy:
Thank you for the review, I've updated the version :)

PR #1490, opened 2026-05-20
Add `$err:` variable support in try-catch blocks

CaiJimmy:


src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -3025,11 +2969,15 @@ public Node visitQuantifiedExpr(JsoniqParser.QuantifiedExprContext ctx) {
 
     @Override
     public Node visitTryCatchExpr(JsoniqParser.TryCatchExprContext ctx) {
-        Expression tryExpression = (Expression) this.visitExpr(ctx.try_expression);
+        Expression tryExpression = ctx.try_expression == null
```
ghislainfourny, line 3028: Good catch!

ghislainfourny (approved):

PR #1492, opened 2026-05-20
Make map:entry(), map:put() and map:remove() more efficient. Running test queue now takes 10 minutes instead of 22.

ghislainfourny:


ghislainfourny:
@CaiJimmy I think I managed to fix the performance issue with the XQuery test suite. Since there are a lot of changes, it would be nice if you could review the code. It is also a good opportunity to make yourself familiar with the implementation of RumbleDB :)

src/main/java/org/rumbledb/runtime/functions/maps/MapRemoveFunctionIterator.java
```diff
@@ -104,6 +104,11 @@ private void initializeResult(DynamicContext context) {
             this.resultItem = mapItem;
             return;
         }
+        if (mapItem.getMutabilityLevel() == -1) {
+            // fast path: update the mutable map in place
```
CaiJimmy: I'm not sure if I understood this comment. Shouldn't -1 mean that the map is immutable?

src/main/java/org/rumbledb/runtime/functions/maps/MapPutFunctionIterator.java
```diff
@@ -87,6 +87,10 @@ public Item materializeFirstItemOrNull(DynamicContext context) {
         List<Item> valueSequence = new ArrayList<>();
         this.valueIterator.materialize(context, valueSequence);
 
+        if (mapItem.getMutabilityLevel() == -1) {
+            // fast path: update the mutable map in place
```
CaiJimmy: Same here, I think it should be `immutable map`?

src/main/java/org/rumbledb/items/MapEntryItem.java
```diff
@@ -0,0 +1,373 @@
...
+    /**
+     * This is an optimization version of maps when there is exactly one key-value pair.
+     */
+    private Item key;
+    private List<Item> value;
+
+    public MapEntryItem() {
+        this.key = null;
+        this.value = Collections.emptyList();
+    }
```
CaiJimmy, line 53: Do we need this constructor? I think it's not used anywhere for now.

src/main/java/org/rumbledb/items/MapEntryItem.java
```diff
@@ -0,0 +1,373 @@
...
+
+    public MapEntryItem(Item key, List<Item> value) {
+        this.key = key;
```
CaiJimmy, line 56: Maybe we should check that key is not null here. Because in other methods like `isObject()`, we call methods of key without checking whether it's null pointer or not.

src/main/java/org/rumbledb/items/ArrayItem.java
```diff
@@ -242,13 +243,7 @@ public ItemType getDynamicType() {
...
             return ItemTypeFactory.createEmptyArrayType();
         }
-
-        ItemType result = this.arrayItems.get(0).getDynamicType();
-        for (int i = 1; i < this.arrayItems.size(); i++) {
-            result = result.findLeastCommonSuperTypeLax(this.arrayItems.get(i).getDynamicType());
-        }
-
-        return ItemTypeFactory.createAnonymousArrayType(result);
+        return BuiltinTypesCatalogue.arrayItem;
```
CaiJimmy, line 246: So if I understand correctly, we are moving the computation of dynamic type of object and array from class method to `createItemTypeFromItem` of `ItemFactory`?

src/main/java/org/rumbledb/items/MapEntryItem.java
ghislainfourny: This is for Kryo serialization.

PR #1493, opened 2026-05-20
fix: handle wildcard expressions in try-catch expressions and statements correctly

CaiJimmy:
We are currently comparing error code in catch declaration without considering namespace resolution, this pull request fixes that.

Also, previously we haven't considered that there are four different types of wildcard in catch declaration, it should also be fixed now.

`LinkedHashMap` is used instead of `HashMap` so when we iterate through catch blocks, the insertion order is preserved.


ghislainfourny (approved):
Looks great, and very nice first use of the regression reports!

PR #1496, opened 2026-05-22
Add regression report to QT3 pipeline

CaiJimmy:


CaiJimmy:
There's probably something non-deterministic going on with tests nodeexpression22 and nodeexpression35, which sometimes fails and sometimes does not. Need to be investigated.

ghislainfourny (approved):

PR #1511, opened 2026-06-05
Support partial builtin function call

### LINKED ISSUE #1485: partial function application of builtin functions
### Feature Description

Partial function application does not yet work with builtin functions.

### Proposed Solution

_No response_

### Additional Context

_No response_

CaiJimmy:
closes https://github.com/RumbleDB/rumble/issues/1485

Idea: transform partial builtin function call into a new inline function

### DIFF
--- a/src/main/java/org/rumbledb/compiler/BuiltinPartialApplicationRewriteVisitor.java
+++ b/src/main/java/org/rumbledb/compiler/BuiltinPartialApplicationRewriteVisitor.java
@@ -27,7 +27,10 @@ public class BuiltinPartialApplicationRewriteVisitor extends CloneVisitor {
         BuiltinFunction builtin = BuiltinFunctionCatalogue.getBuiltinFunction(expression.getFunctionIdentifier());
 
         if (!expression.isPartialApplication() || builtin == null) {
-            return expression;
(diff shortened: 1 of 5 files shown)


ghislainfourny (approved):
Very nice!
Quick question: do we also now support the following query, or does it only work when it is known at compile time that the function is builtin?
For example:
let $f := math:pow#2 return $f(?,2)

CaiJimmy:
@ghislainfourny 
No, it does not work yet:

```
org.rumbledb.exceptions.UnsupportedFeatureException: There was an error on line 1 in file:/Users/jimmy/Documents/BDFE/exam_docker_lab/notebooks/test.jq:

let $f := math:pow#2 return $f(?,2)
                            ^

Code: [RBST0002]
Message: Partial application of builtin named function references is not supported yet.
```

I can try to fix it. Should I merge this one first or wait to have it fixed?

ghislainfourny:
I am fine either way. This may actually be quite complex and cannot be solved with a compile-time visitor. It probably needs some clockwork implementation similar to what was already done for partial applications of user-defined-functions. Thus, I am fine with merging this, since it nicely solves static partial builtin function calls.

PR #1514, opened 2026-06-05
Create PartialFunctionCallIterator to centralize the logic of partial functions

CaiJimmy:


ghislainfourny (approved):
Very nicely architectured!

A minor hint: perhaps rename FunctionCallIterator to StaticFunctionCallIterator and PartialFunctionCallIterator to StaticPartialFunctionCallIterator to make clear that partial dynamic function calls are (already) covered by the dynamic function call iterator?

PR #1515, opened 2026-06-16
Fix mutable map copy values

CaiJimmy:
Mutable copy should not start with `values = this.getItemValues();`, but a new initialized ArrayList


ghislainfourny (approved):
Looks great, there is just MapEntryItem to fix to use an import instead of hardwiring the package in the code.

PR #1528, opened 2026-06-18
feat: add method to retrieve unmodifiable map of builtin functions

CaiJimmy:
This is useful for LSP, which needs to retrieve the list of builtin functions. Currently it's done via reflection


ghislainfourny (approved):
It makes sense!

PR #1532, opened 2026-06-18
fix: JSONiq XML tag attribute not parsed correctly

### LINKED ISSUE #1519: XML attributes are not parsed correctly in JSONiq Parser
### What did you expect to happen?

```
<html class="1"></html>
```

### How can we reproduce it (as minimally and precisely as possible)?

Run it

### Anything else we need to know?

_No response_

### RumbleDB version

<details>
after running the RumbleDB, it will show version next to the logo.
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Related Software versions

<details>

```console
$ java -version
# paste output here
$ mvn --version
# paste output here
$ ant -version
# paste output here
$ spark-submit --version
# paste output here
```

</details>

CaiJimmy:
closes #1519

### DIFF
--- a/.github/workflows/run-qt3.yml
+++ b/.github/workflows/run-qt3.yml
@@ -104,7 +104,7 @@ jobs:
 
   qt3-jsoniq:
     needs: [resolve-baseline, build]
-    uses: RumbleDB/rumble-test-suite/.github/workflows/qt3-suite.yml@master
+    uses: CaiJimmy/rumble-test-suite/.github/workflows/qt3-suite.yml@fix-xml-string-convert
     with:
       tested_parser: jsoniq
       rumble_artifact_name: rumble-build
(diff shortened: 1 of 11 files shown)


ghislainfourny:
A quick hint on this one:
In JSONiq:
1. String literals only follow backslash escaping and & is a character like any other (so that JSON content can be copy-pasted without modifications). We never mix & and \ escaping in the same string literal, it is either or. It is thus expected that some test suite tests will fail in the JSONiq syntax.
2. Attribute values in direct element constructors (with the attr="..." XML syntax inside the element opening tag, which are technically not string literals) do follow & escaping and \ is a normal character, so that XML can be copy-pasted.
3. (unquoted) Content within a direct element constructor also follows & escaping with \ being a normal character.

CaiJimmy:
Moved to https://github.com/RumbleDB/rumble/pull/1659

Part of this PR has fixes for `fn:translate`, that will be dealt in a separate PR.

PR #1538, opened 2026-06-25
Language improvements in formatting

ghislainfourny:


HenrikPaetzold:
@ghislainfourny I think this pipeline still doesn't use the newest master-henrik branch of the rumble-test-suite, right? There is still a lot of tests that should not be skipped anymore.

pom.xml
```diff
@@ -439,6 +439,11 @@
             <artifactId>mongo-spark-connector_2.13</artifactId>
             <version>10.5.0</version>
         </dependency>
+        <dependency>
```
ghislainfourny, line 487: From online documentation it seems ICU4J might be shipped together with Spark. Thus, we can declare this dependency as provided (see how this is done with other dependencies) to not make the RumbleDB jar larger.
It is probably a good idea to align the ICU4J version with the one that Spark 4.0.3 depends on (there is a maven command that should allow you to recursively list all dependencies of Spark).

ghislainfourny:
> @ghislainfourny I think this pipeline still doesn't use the newest master-henrik branch of the rumble-test-suite, right? There is still a lot of tests that should not be skipped anymore.

This is correct. It took a while to merge the rumble-test-suite because of a chicken and egg issue but it is now solved. The rumble-test-suite contains your latest changes.

PR #1546, opened 2026-06-25
Fix compiler warnings

CaiJimmy:


src/main/java/org/rumbledb/runtime/functions/util/formatting/pictures/FormatNumber/NumberPictureFormatter.java
```diff
@@ -355,79 +353,4 @@ private static final class MantissaExponentPair {
         }
     }
 
-    // DEBUGGING METHODS
```
ghislainfourny, line 358: These are intentionnally here, unused but here in case any debugging is needed.

ghislainfourny (changes requested):

src/main/java/org/rumbledb/runtime/functions/util/formatting/pictures/FormatNumber/NumberPictureFormatter.java
CaiJimmy: > These are intentionnally here, unused but here in case any debugging is needed.

I see, they have been recovered

PR #1549, opened 2026-06-25
Refactor to use `instanceof` pattern matching

CaiJimmy:
In the codebase we have many usage of `instanceof` followed by a manual casting. Java 14 introduced pattern matching for instanceof that simplified this (https://www.baeldung.com/java-pattern-matching-instanceof)

For example:

```java
if (node instanceof InlineFunctionExpression) {
      ((InlineFunctionExpression) node).setExpressionClassification(expressionClassification);
      return expressionClassification;
}
```

Becomes:

```java
if (node instanceof InlineFunctionExpression inlineFunctionExpression) {
    inlineFunctionExpression.setExpressionClassification(expressionClassification);
    return expressionClassification;
}
```


ghislainfourny (approved):
I wanted to do this for a long time. Thank you for taking care of it!

PR #1553, opened 2026-06-25
Refactor more classes to use the instanceof pattern-matching operator

CaiJimmy:


ghislainfourny (approved):
Thank you for taking care of this!

PR #1547, opened 2026-06-25
Remove deprecated method calls and implementations

CaiJimmy:
Merge after #1546

PR #1555, opened 2026-06-25
Drop deprecated method definitions

CaiJimmy:
Merge after #1547


ghislainfourny (commented):
This we should do shortly before the 3.0 release. We can only remove deprecated methods in a major release. We can keep this PR open until then.
(We can absolutely make the next release 3.0 to do so, btw)

PR #1556, opened 2026-06-25
Replace StringBuffer with StringBuilder

CaiJimmy:
StringBuffer is thread-safe. However, we only ever use local StringBuffers, so we never have a synchronisation problem. And StringBuilder is faster compared to StringBuffer.

Reference: https://www.geeksforgeeks.org/java/stringbuffer-vs-stringbuilder/


ghislainfourny (approved):
Great idea!

PR #1568, opened 2026-06-28
Upgrade Spotless plugin to 3.7.0

CaiJimmy:


pom.xml
```diff
@@ -218,11 +218,6 @@
                             <!-- include all java files in "java" folders under "src" -->
                             <include>src/**/java/**/*.java</include>
                         </includes>
-
```
ghislainfourny, line 221: What is the rationale for removing these?

ghislainfourny (commented):
Thank you Jimmy! I just have one question.

ghislainfourny (approved):

pom.xml
CaiJimmy: With this pull request (https://github.com/RumbleDB/rumble/pull/1510), Java class files generated by ANTLR are now stored in the `generated-sources` directory rather than the `src` directory. Therefore, the exclusion rule will never apply as these files are not included.

PR #1569, opened 2026-06-28
Add Lombok as Maven dependency

CaiJimmy:
Merge after https://github.com/RumbleDB/rumble/pull/1568

This will be needed later for #1567, but because that pull request has a lot of changes, I'm splitting it into smaller pull requests.

Lombok offers many useful annotations like `@Getter` and `@Setter`, which helps avoid writing trivial getters and setters in the code. Maybe we can do a migration in the future removing trivial getters and setters to clean up the codebase.

Because we are using Eclipse compiler, a special `.jar` file has to be added to support Lombok, reference: https://projectlombok.org/setup/ecj


pom.xml
```diff
@@ -203,26 +252,13 @@
             <plugin>
                 <groupId>com.diffplug.spotless</groupId>
                 <artifactId>spotless-maven-plugin</artifactId>
-                <version>3.3.0</version>
```
ghislainfourny: I do not understand these changes, what are they for?

pom.xml
```diff
@@ -203,26 +252,13 @@
...
-                    </execution>
-                </executions>
+                <version>3.7.0</version>
                 <configuration>
                     <java>
                         <includes>
                             <!-- include all java files in "java" folders under "src" -->
                             <include>src/**/java/**/*.java</include>
                         </includes>
-
```
ghislainfourny: I do not understand these changes, what are they for?

pom.xml
```diff
@@ -61,16 +62,59 @@
...
                 <configuration>
-                    <compilerId>eclipse</compilerId>
                     <source>17</source>
                     <target>17</target>
-                    <compilerArgs>
-                        <arg>-properties</arg>
-                        <arg>${project.basedir}/org.eclipse.jdt.core.prefs</arg>
-                    </compilerArgs>
                 </configuration>
+                <executions>
```
ghislainfourny, line 69: This seems unrelated. What is the goal here?

ghislainfourny (approved):
Perfect, this is good to go!

PR #1574, opened 2026-06-29
Fix XQuery string literals with braces, correct atomization of elements and documents to untypedAtomic

ghislainfourny:


ghislainfourny:
Comment to self: the regressions are XQuery 1.0 tests that were incorrectly successful. We will have to update the test suite to make sure to execute these in XQuery 1.0 mode.

PR #1583, opened 2026-06-30
Add new configuration model classes for Rumble

CaiJimmy:
## Summary

This pull request introduces `RumbleConfiguration` as a new typed configuration model for RumbleDB. It is intended to replace the monolithic `RumbleRuntimeConfiguration` with a structured, immutable configuration aggregate composed of smaller domain-specific config objects.

The classification is almost identical to the one used in new CLI parser, so we can perform a 1 to 1 conversion (#1581)

**This pull request only adds the new classes. It does not yet replace the old `RumbleRuntimeConfiguration`.**

## Compared to `RumbleRuntimeConfiguration`

`RumbleRuntimeConfiguration` combines CLI argument parsing, defaulting, validation, mutable runtime state, and configuration access in a single class. `RumbleConfiguration` separates those concerns by grouping configuration into typed sections such as `input`, `output`, `runtime`, `bindings`, `semantics`, and `server`.

This makes configuration easier to understand, easier to evolve, and safer to use. Instead of mutating one large object through setters, callers build **immutable** configuration values and access them through a clear nested API.

## Main Improvements

- Replaces a flat mutable config object with a typed aggregate.
- Separates configuration by responsibility into dedicated model classes.
- Uses immutable value objects and builder-based construction.
- Localizes defaults and validation within the relevant config models. (Instead of a huge `init()` method)
- Introduces explicit `ExecutionMode` enum class instead of inferring mode from raw CLI state.

## Usage

Getter-style access:

```java
RumbleConfiguration configuration = RumbleConfiguration.builder().build();

String queryPath = configuration.input().queryPath();
int resultSizeCap = configuration.runtime().resultsSizeCap();
boolean showErrorInfo = configuration.debug().showErrorInfo();
String xmlVersion = configuration.semantics().xmlVersion();
```

Builder-style construction:

```java
RumbleConfiguration configuration = RumbleConfiguration.builder()
    .executionMode(ExecutionMode.RUN)
    .input(InputConfig.builder().queryPath("queries/main.jq").build())
    .runtime(RuntimeConfig.builder().resultsSizeCap(100).build())
    .debug(DebugConfig.builder().showErrorInfo(true).build())
    .build();
```

Builder-style update:

```java
RumbleConfiguration updated = configuration.toBuilder()
    .runtime(configuration.runtime().toBuilder().resultsSizeCap(500).build())
    .build();
```

## New public configuration API (`org.rumbledb.api.RumbleConfiguration`)

We have now a public facing configuration API (previously, we only had the internal one)

Configuration can now be created through a builder:

```java
RumbleConfiguration configuration = RumbleConfiguration.builder()
    .with("runtime.resultsSizeCap", 100)
    .with("debug.showErrorInfo", true)
    .with("semantics.xmlVersion", "1.1")
    .build();
```

Existing configurations can be derived without mutation:

```java
RumbleConfiguration updated = configuration.toBuilder()
    .with("runtime.resultsSizeCap", 200)
    .build();
```

Values can be retrieved as plain Java values or converted to a requested type:

```java
int cap = configuration.get("runtime.resultsSizeCap", Integer.class);
Object runtime = configuration.get("runtime");
```

The PR also introduces `ExternalBindings`, separating query inputs from engine configuration:

```java
ExternalBindings bindings = new ExternalBindings();

bindings.bindItem("item", item);
bindings.bindItems("items", items);
bindings.bindDataFrame("rows", dataFrame);
bindings.unbind("item");
```


src/main/java/org/rumbledb/bindings/Binding.java
```diff
@@ -0,0 +1,5 @@
+package org.rumbledb.bindings;
+
+public sealed interface Binding
+        permits DataFrameBinding, FileBinding, ItemSequenceBinding, LexicalBinding, StandardInputBinding {
+}
```
CaiJimmy, line 5: Instead of defining methods like `isDataFrameBinding()` + `getDataFrameBinding`, I believe it's better to use sealed interface (Java 17+) to indicate that these classes are **disjoint types**, and we can benefit from exhaustive switch feature in the future (Java 21+):

```java
// Java 21+ — both in the same compilation unit
sealed interface PaymentResult permits Success, Declined, Pending {}
 
record Success(String txId)   implements PaymentResult {}
record Declined(String reason) implements PaymentResult {}
record Pending(String ref)    implements PaymentResult {}
 
// This compiles — all three variants covered
String describe(PaymentResult r) {
    return switch (r) {
        case Success s  -> "Paid: "    + s.txId();
        case Declined d -> "Refused: " + d.reason();
        case Pending p  -> "Waiting: " + p.ref();
        // no default needed — compiler sees all three
    };
}
```

src/main/java/org/rumbledb/api/RumbleConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+        return new RumbleConfigurationBuilder(this.configuration.toBuilder());
+    }
+
+    /**
+     * Returns a configuration value as a plain Java value.
+     *
+     * @param path dot-separated configuration path
+     * @return the value at the requested path
+     */
+    public Object get(String path) {
```
ghislainfourny, line 39: I think this signature should be compatible with Python. We will test it once this is all in place.

src/main/java/org/rumbledb/api/RumbleConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns a configuration value converted to the requested Java type.
+     *
+     * @param path dot-separated configuration path
+     * @param valueType requested Java type
+     * @param <T> requested Java type
+     * @return the value at the requested path
+     */
+    public <T> T get(String path, Class<T> valueType) {
```
ghislainfourny: How many different types are there? Is it only string, integer and boolean?

If so these three additional methods (calling the templated getter) would be convenient to the end user:
getAsString()
getAsBoolean()
getAsInt()

src/main/java/org/rumbledb/bindings/Binding.java
ghislainfourny: This is a sensible design choice.

src/main/java/org/rumbledb/config/model/ExecutionMode.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ *
+ */
+
+package org.rumbledb.config.model;
+
+/**
+ * Application execution mode selected by the launcher.
+ */
+public enum ExecutionMode {
```
ghislainfourny: The terminology 'Execution Mode' is already used to describe lists vs. rdd vs. dataframes. Perhaps just RumbleMode or something like this?

ghislainfourny (commented):
It looks great! I left a few comments.

src/main/java/org/rumbledb/api/RumbleConfiguration.java
CaiJimmy: For now, we have:


- `List<String>`
- `boolean`
- `int`
- `string`
- `ExecutionMode(Enum)`

ghislainfourny (approved):
Great job!

PR #1604, opened 2026-07-03
Migrate to JUnit 6

CaiJimmy:
The current JUnit 4 is deprecated.

Also fixes a problem that XQueryTest was not loading `.xq` files properly (it was looking for `.jq` files, so no test was running)


ghislainfourny (approved):
Great initiative!

PR #1605, opened 2026-07-03
Cleanup repeated test class file codes by moving them to shared methods

CaiJimmy:
- Annotation test execution logic is moved out from AnnotationTestBase to a separate AnnotationTestExecutor class
- New SparkAnnotationsTestBase and UpdateRuntimeTestsBase to keep shared logic of these type of tests

Stop duplicating the same RumbleRuntimeConfiguration inside each test case


ghislainfourny (approved):
Great work as always!

PR #1607, opened 2026-07-06
Migrate to new RumbleConfiguration

CaiJimmy:
This pull request migrates the codebase to use `RumbleConfiguration` instead of `RumbleRuntimeConfiguration`. It removes all the references to `RumbleRuntimeConfiguration`, and modifies the signature of internal and public methods.

It also implements Serializable and KryoSerializable interface for `RumbleConfiguration` so it can work in Spark correctly.

`RumbleRuntimeConfiguration` will be deleted in the upcoming pull request.


src/test/java/iq/NativeFLWORRuntimeTestsParallelismDeactivated.java
```diff
@@ -20,28 +20,19 @@
...
 
 public class NativeFLWORRuntimeTestsParallelismDeactivated extends RuntimeTests {
 
     @Override
-    public RumbleRuntimeConfiguration getConfiguration() {
-        return new RumbleRuntimeConfiguration(
-                new String[] {
-                    "--variable:externalUnparsedString",
-                    "unparsed string",
-                    "--escape-backticks",
```
CaiJimmy, line 35: `--escape-backticks` was an invalid flag (not found in RumbleRuntimeConfiguration)

src/test/java/iq/DeltaUpdateRuntimeTests.java
```diff
@@ -33,31 +32,6 @@ public class DeltaUpdateRuntimeTests extends UpdateRuntimeTestsBase {
...
-                "--print-iterator-tree",
-                "yes",
-                "--output-format",
-                "delta",
-                "--show-error-info",
-                "yes",
-                "--apply-updates",
-                "yes",
-            }
-    );
```
CaiJimmy, line 60: For some reason, these config settings were used in UpdatesForRumbleBenchmarkTest, so I moved them there.

src/test/java/iq/base/TestConfigurations.java
```diff
@@ -28,28 +28,35 @@ public final class TestConfigurations {
...
-                    ItemFactory.getInstance().createIntItem(2),
-                    ItemFactory.getInstance().createIntItem(3),
-                    ItemFactory.getInstance().createIntItem(4),
-                    ItemFactory.getInstance().createIntItem(5)
-                )
-            );
+    public static RumbleConfiguration.RumbleConfigurationBuilder defaultConfigurationBuilder() {
+        return RumbleConfiguration.builder()
+            .configureRuntime(r -> r.materializationCap(200));
+    }
```
CaiJimmy, line 34: The default configuration is not yet used everywhere. We should create a separate pull request (to not create too much change in this) and make use of it.

src/main/java/org/rumbledb/api/SequenceOfItems.java
```diff
@@ -55,7 +55,7 @@ public class SequenceOfItems {
     public SequenceOfItems(
             RuntimeIterator iterator,
             DynamicContext dynamicContext,
-            RumbleRuntimeConfiguration configuration
+            RumbleConfiguration configuration
     ) {
```
CaiJimmy, line 59: Because this constructor should not be used directly (by user), I have changed it to receive the internal configuration object

src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -280,79 +286,69 @@ public Node visitMainModule(JsoniqParser.MainModuleContext ctx) {
...
-            prolog.addDeclaration(
-                new VariableDeclaration(
-                        externalVariable,
-                        true,
-                        SequenceType.createSequenceType("item*"),
-                        null,
-                        null,
-                        createMetadataFromContext(ctx)
-                )
-            );
```
CaiJimmy, line 348: I cleaned up the code here. Previously, we stored lexical, item, and data frame bindings in independent maps, which caused potential collisions. Now, we only keep one map, so only one for loop is needed, eliminating this problem.

ghislainfourny (approved):

PR #1618, opened 2026-07-08
implement more functions

HenrikPaetzold:


src/main/java/org/rumbledb/runtime/functions/xml/HasChildrenFunctionIterator.java
```diff
@@ -21,6 +23,22 @@ public HasChildrenFunctionIterator(
 
     @Override
     public Item materializeFirstItemOrNull(DynamicContext context) {
-        throw new UnimplementedFunctionException("fn:has-children", getMetadata());
+        Item node = getContextNode(context);
+        if (node == null) {
+            return ItemFactory.getInstance().createBooleanItem(false);
+        }
+        if (!node.isNode()) {
```
ghislainfourny, line 30: Normally many of these checks are caught by the function signature already and are redundant. You can check the entry in the builtin function catalogue.

ghislainfourny: In this case I think the check is only needed for the 0-arity function, which gets the node from the context item.

ghislainfourny (approved):

src/main/java/org/rumbledb/runtime/functions/xml/HasChildrenFunctionIterator.java
HenrikPaetzold: Yes. I adapted the entries in the `BuiltinFunctionCatalogue` yesterday, so that the signature takes `node()` instead of `item`, but I was not sure if that would also apply to the context item.

PR #1625, opened 2026-07-08
Display failed test case file name in Java Maven test

CaiJimmy:
Due to PR #1604, the name of the exact file that caused the test failure is no longer showing in the console report. This is an issue between the JUnit 5+ and Surefire plugins. I configured Surefire to output the file name in the XML report. However, this configuration cannot be applied to the console output. Hence, this commit adds an auxiliary method that adds the file name to the assertion error message.

I also discovered a GitHub action that reads the Surefire XML and displays the results as an action annotation: https://github.com/ScaCap/action-surefire-report

It does not seem very popular, but it works.

image

So we no longer need to look for the exact fail case in the huge log.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,7 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     continue-on-error: true
     permissions:
-      checks: write
       contents: read
     steps:
       - name: Checkout the repository
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -14,8 +14,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     continue-on-error: true
-    permissions:
-      contents: read
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v5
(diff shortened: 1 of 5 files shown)


ghislainfourny:
Great work @CaiJimmy ! You can merge after reverting the purposefully failed test.

ghislainfourny (approved):

.github/workflows/maven.yml
```diff
@@ -101,3 +104,11 @@ jobs:
           path: target
       - name: Run Test ${{ matrix.test_class }}
         run: mvn -Dtest=${{ matrix.test_class }} test
+      - name: Publish Surefire Report ${{ matrix.test_class }}
+        if: always()
+        uses: ScalableCapital/action-surefire-report@v2
+        with:
+          report_paths: target/surefire-reports/TEST-*.xml
+          check_name: Test Report - ${{ matrix.test_class }}
+          fail_on_test_failures: false
```
ghislainfourny, line 110: How come the test report still marks as failed on test failures in spite of this being false?

ghislainfourny:
Is there a way to "merge" the tests and Test Report checks, to avoid increasing the list?

PR #1634, opened 2026-07-09
Fix regression.

ghislainfourny:


ghislainfourny:
It seems like the regression was not really a regression, not sure why it was flagged as such. Perhaps because the file was not found, which is the same error?

HenrikPaetzold (approved):

PR #1640, opened 2026-07-10
Revert qt3tests to use master.

ghislainfourny:


ghislainfourny:
@HenrikPaetzold @CaiJimmy these 18 regressions are expected during the transition period right after merging this test suite change.

This is because we are now skipping tests not applicable to version 3.1.

PR #1653, opened 2026-07-10
Fix atomic item handling in max / min function

CaiJimmy:


src/main/java/org/rumbledb/runtime/functions/sequences/aggregate/MaxFunctionIterator.java
```diff
@@ -81,7 +82,13 @@ public MaxFunctionIterator(
...
     ) {
         super(arguments, staticContext);
-        this.iterator = this.children.get(0);
+        // fn:max() applies atomization as part of function conversion. Keep this
+        // explicit here because the aggregate iterator consumes its argument
+        // directly rather than through a typed function-call wrapper.
+        this.iterator = new DataFunctionIterator(
+                List.of(this.children.get(0)),
+                staticContext
+        );
```
ghislainfourny: This is an unusual pattern. The design contract of function iterators is that the input is guaranteed to match the signature. Normally this iterator should be added automatically upon calling based on the anyAtomicType* signature. Is there a calling path where this is not done?

src/main/java/org/rumbledb/runtime/functions/sequences/aggregate/MinFunctionIterator.java
```diff
@@ -79,7 +80,13 @@ public MinFunctionIterator(
...
     ) {
         super(arguments, staticContext);
-        this.iterator = this.children.get(0);
+        // fn:min() applies atomization as part of function conversion. Keep this
+        // explicit here because the aggregate iterator consumes its argument
+        // directly rather than through a typed function-call wrapper.
+        this.iterator = new DataFunctionIterator(
+                List.of(this.children.get(0)),
+                staticContext
+        );
```
ghislainfourny: This is an unusual pattern. The design contract of function iterators is that the input is guaranteed to match the signature. Normally this iterator should be added automatically upon calling based on the anyAtomicType* signature. Is there a calling path where this is not done?

ghislainfourny (changes requested):

src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
```diff
@@ -1056,14 +1056,14 @@ private static BuiltinFunction createBuiltinFunctionWithSequenceTypes(
      */
     static final BuiltinFunction min1 = createBuiltinFunction(
         new Name(Name.FN_NS, "fn", "min"),
-        List.of("item*"),
+        List.of("anyAtomicType*"),
```
ghislainfourny, line 1059: Nice catch! This is the official way to ensure atomization upon calling the function.

src/main/java/org/rumbledb/runtime/functions/sequences/aggregate/MaxFunctionIterator.java
CaiJimmy: You are right! I have removed this extra code.

ghislainfourny (approved):
Great job!

PR #1654, opened 2026-07-10
Fix element constructor and missing space between atomic elements

CaiJimmy:
A space was missing between atomic elements inside a element constructor.

And check is added to prevent prefix collision in element constructor

### DIFF
--- a/src/main/java/org/rumbledb/expressions/xml/ComputedNamespaceConstructorExpression.java
+++ b/src/main/java/org/rumbledb/expressions/xml/ComputedNamespaceConstructorExpression.java
@@ -101,7 +101,14 @@ public class ComputedNamespaceConstructorExpression extends Expression {
 
     @Override
     public List<Node> getChildren() {
-        return new ArrayList<>();
+        List<Node> result = new ArrayList<>();
+        if (this.prefixExpression != null) {
+            result.add(this.prefixExpression);
+        }
+        if (this.uriExpression != null) {
+            result.add(this.uriExpression);
+        }
+        return result;
     }
 
     @Override
@@ -120,4 +127,3 @@ public class ComputedNamespaceConstructorExpression extends Expression {
         sb.append(" }\n");
     }
 }
-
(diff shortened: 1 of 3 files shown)


src/main/java/org/rumbledb/runtime/xml/DirElemConstructorRuntimeIterator.java
```diff
@@ -124,9 +126,13 @@ public Item materializeFirstItemOrNull(DynamicContext dynamicContext) {
                             textAccumulator = new StringBuilder();
                         }
 
+                        if (item.isAtomic() && previousItemWasAtomic) {
+                            textAccumulator.append(' ');
+                        }
                         // accumulate the text content
```
CaiJimmy, line 132: This was cause of many test failures in XML

src/main/java/org/rumbledb/expressions/xml/ComputedNamespaceConstructorExpression.java
```diff
@@ -101,7 +101,14 @@ public <T> T accept(AbstractNodeVisitor<T> visitor, T argument) {
 
     @Override
     public List<Node> getChildren() {
-        return new ArrayList<>();
+        List<Node> result = new ArrayList<>();
```
ghislainfourny, line 104: Good catch, indeed this one was not implemented correctly

src/main/java/org/rumbledb/items/xml/ElementItem.java
```diff
@@ -392,7 +392,37 @@ public void addOrReplaceNamespace(Item namespaceItem) {
...
             this.namespaces = new HashMap<>();
         }
-        this.namespaces.put(namespace.getPrefix(), namespace.getUri());
+        String prefix = namespace.getPrefix();
+        String uri = namespace.getUri();
+
+        // Namespace declarations can conflict with the prefix retained in a
+        // constructed element/attribute QName. Preserve the expanded name by
+        // assigning a fresh prefix and declaring it for the original URI.
+        if (hasConflictingPrefix(this.dmNodeName, prefix, uri)) {
```
ghislainfourny, line 401: Good catch

src/main/java/org/rumbledb/runtime/xml/DirElemConstructorRuntimeIterator.java
ghislainfourny: Also a nice fix! This is explicitly in the specification
https://www.w3.org/TR/xquery-31/#id-content
under bullet point 1.e.iii

ghislainfourny (approved):
Congrats!

PR #1656, opened 2026-07-12
Fix unparsed-text tests

HenrikPaetzold:
Makes `fn:unparsed-text`, `fn:unparsed-text-lines`, and `fn:unparsed-text-available` more compliant with F&O 3.1 (§14.6.5–14.6.7). Also introduces a dedicated `unparsed-text-lines` class instead of reusing the shared `jn:text-file` class (which crashed on string encoding). Encoding detection/decoding is now shared between `fn:json-doc` and the `unparsed-text*` family via `TextResourceUtil` instead of being duplicated. Additionally includes a trailing-slash fix in `VisitorHelpers` for base-URI resolution of relative hrefs.

### DIFF
--- a/src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
+++ b/src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
@@ -789,14 +789,14 @@ public class BuiltinFunctionCatalogue {
         new Name(Name.FN_NS, "fn", "unparsed-text-lines"),
         List.of("string?"),
         "string*",
-        org.rumbledb.runtime.functions.io.UnparsedTextLinesFunctionIterator.class,
+        UnparsedTextLinesFunctionIterator.class,
         BuiltinFunction.BuiltinFunctionExecutionMode.RDD
     );
     static final BuiltinFunction unparsed_text_lines2 = createBuiltinFunction(
         new Name(Name.FN_NS, "fn", "unparsed-text-lines"),
         List.of("string?", "string"),
         "string*",
-        org.rumbledb.runtime.functions.io.UnparsedTextLinesFunctionIterator.class,
+        UnparsedTextLinesFunctionIterator.class,
         BuiltinFunction.BuiltinFunctionExecutionMode.RDD
     );
     static final BuiltinFunction text_file1 = createBuiltinFunction(
--- a/src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
+++ b/src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
@@ -789,14 +789,14 @@ public class BuiltinFunctionCatalogue {
         new Name(Name.FN_NS, "fn", "unparsed-text-lines"),
         List.of("string?"),
         "string*",
-        UnparsedTextLinesFunctionIterator.class,
+        org.rumbledb.runtime.functions.io.UnparsedTextLinesFunctionIterator.class,
         BuiltinFunction.BuiltinFunctionExecutionMode.RDD
     );
     static final BuiltinFunction unparsed_text_lines2 = createBuiltinFunction(
         new Name(Name.FN_NS, "fn", "unparsed-text-lines"),
         List.of("string?", "string"),
         "string*",
-        UnparsedTextLinesFunctionIterator.class,
+        org.rumbledb.runtime.functions.io.UnparsedTextLinesFunctionIterator.class,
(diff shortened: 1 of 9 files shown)


HenrikPaetzold:
@CaiJimmy, sorry for pinging. :( Would you happen to know, why the analysis generation for qt3-jsoniq keeps failing?

```
[ERROR] Cannot use 'in' operator to search for 'integrity' in undefined
pnpm: Cannot use 'in' operator to search for 'integrity' in undefined
```

CaiJimmy:
> @CaiJimmy, sorry for pinging. :( Would you happen to know, why the analysis generation for qt3-jsoniq keeps failing?
> 
> ```
> [ERROR] Cannot use 'in' operator to search for 'integrity' in undefined
> pnpm: Cannot use 'in' operator to search for 'integrity' in undefined
> ```

No worry no worry. It's an upstream bug. I have fixed the version of pnpm temporarily to prevent this (https://github.com/pnpm/action-setup/issues/276)

Should be fixed now (hopefully)

HenrikPaetzold:
That seems to have worked, thanks! :)

src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
```diff
@@ -789,14 +789,14 @@ private static BuiltinFunction createBuiltinFunctionWithSequenceTypes(
         new Name(Name.FN_NS, "fn", "unparsed-text-lines"),
         List.of("string?"),
         "string*",
-        UnparsedTextLinesFunctionIterator.class,
+        org.rumbledb.runtime.functions.io.UnparsedTextLinesFunctionIterator.class,
```
ghislainfourny: Is there a reason for adding the package prefix here?

src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
```diff
@@ -789,14 +789,14 @@ private static BuiltinFunction createBuiltinFunctionWithSequenceTypes(
...
         BuiltinFunction.BuiltinFunctionExecutionMode.RDD
     );
     static final BuiltinFunction unparsed_text_lines2 = createBuiltinFunction(
         new Name(Name.FN_NS, "fn", "unparsed-text-lines"),
         List.of("string?", "string"),
         "string*",
-        UnparsedTextLinesFunctionIterator.class,
+        org.rumbledb.runtime.functions.io.UnparsedTextLinesFunctionIterator.class,
```
ghislainfourny: Is there a reason for adding the package prefix here?

src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
```diff
@@ -4087,16 +4087,16 @@ private static BuiltinFunction createBuiltinFunctionWithSequenceTypes(
 
     static final BuiltinFunction innermost = createBuiltinFunction(
         new Name(Name.FN_NS, "fn", "innermost"),
-        List.of("item*"),
-        "item*",
+        List.of("node()*"),
+        "node()*",
```
ghislainfourny, line 4091: Good catch!

src/main/java/org/rumbledb/runtime/functions/io/UnparsedTextFunctionIterator.java
```diff
@@ -24,21 +24,15 @@
...
+import org.rumbledb.runtime.AtMostOneItemLocalRuntimeIterator;
 import org.rumbledb.runtime.RuntimeIterator;
-import org.rumbledb.runtime.functions.base.LocalFunctionCallIterator;
-import org.rumbledb.runtime.functions.input.FileSystemUtil;
 
-import java.net.URI;
 import java.util.List;
 
-public class UnparsedTextFunctionIterator extends LocalFunctionCallIterator {
+public class UnparsedTextFunctionIterator extends AtMostOneItemLocalRuntimeIterator {
```
ghislainfourny, line 33: Very nice!

ghislainfourny:
296 additional passing tests, huge congratulations!
Just a minor comment regarding the reason for a prefixed package then this is good to go.

src/main/java/org/rumbledb/context/BuiltinFunctionCatalogue.java
HenrikPaetzold: Fixed, thank you! Don't know how that got there.

ghislainfourny (approved):

PR #1658, opened 2026-07-13
Use different iterators for `jn:text-file` and `fn:unparsed-text-line` and fix URI parsing

CaiJimmy:
Previously both uses `UnparsedTextLinesFunctionIterator`. They should not share the same iterator because the second argument has a different meaning (encoding vs num of partitions).

Also, we should probably switch to Java URI implementation instead of Hadoop one, because there are tests failing because Hadoop URI does not support Windows style path.

### DIFF
--- a/src/main/java/org/rumbledb/runtime/functions/URIUtils.java
+++ b/src/main/java/org/rumbledb/runtime/functions/URIUtils.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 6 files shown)


src/main/java/org/rumbledb/runtime/functions/URIUtils.java
```diff
@@ -0,0 +1,59 @@
...
+                    "A relative URI reference requires an absolute base URI"
+            );
+        }
+
+        URI resolvedURI = baseURI.resolve(referenceURI);
+        if (!resolvedURI.isAbsolute()) {
+            throw new URISyntaxException(reference, "The URI reference could not be resolved to an absolute URI");
+        }
+        return resolvedURI;
+    }
```
CaiJimmy, line 58: A separate pull request will be created to make other functions use this util method. Currently we have tests failing because of wrong URI resolution:

Image

PR #1659, opened 2026-07-13
Fix XML attribute parsing in JSONiq parser

### LINKED ISSUE #1519: XML attributes are not parsed correctly in JSONiq Parser

CaiJimmy:
Because of difference in string handling between JSONiq and XQuery grammar, we can only do XML style decoding when the string literal is inside a XML attribute. In the rest of places, string decoding should follow JSONiq rules.

To solve this problem, each time before going into string rule, lexer has to figure out if it's part of a XML element attribute. To do so, a Java based state machine has been introduced:

```mermaid
stateDiagram-v2
    [*] --> Outside

    Outside --> AfterTagOpen: "&lt"
    AfterTagOpen --> AfterTagName: "element QName"
    AfterTagName --> AfterAttributeName: "attribute QName"
    AfterAttributeName --> BeforeAttributeValue: "="

    BeforeAttributeValue --> InsideAttribute: "quote"
    InsideAttribute --> AfterTagName: "matching closing quote"

    AfterTagOpen --> Outside: "not a QName"
    AfterTagName --> Outside: "&gt or /&gt"
    AfterAttributeName --> Outside: "not ="
```

Merge should happen after we update the test-suite to avoid escaping XML attribute strings: https://github.com/RumbleDB/rumble-test-suite/pull/44


src/test/resources/test_files/xml/DirectAttributeEscapeBoundary.jq
```diff
@@ -0,0 +1,4 @@
+(:JIQS: ShouldRun; Output="(true, true, true)" :)
+("&amp;&quot;" eq "&amp;&quot;",
```
ghislainfourny, line 2: In JSONiq, this is literally & a m p ; & q u o t ; that is a string with eleven characters. This equality is correct.
We could also add string-length("&amp;amp;&amp;quot;") eq 11 to check this.

src/test/resources/test_files/xml/DirectAttributeEscapeBoundary.jq
```diff
@@ -0,0 +1,4 @@
...
+ data(<e x="&amp;&quot;"/>/@x) eq "&\"",
```
ghislainfourny: Since this (x="&amp;amp;&amp;quot;") is an XML attribute value (not a string literal), the string value of this attribute has two characters & and " so that the equality with the string literal on the right, which has the same two characters, is indeed correct.

src/test/resources/test_files/parser/JsonStringXMLReferences.jq
```diff
@@ -0,0 +1,11 @@
+(:JIQS: ShouldParse :)
+(
+  {"quotes\"" : 1}."quotes\"",
+  {"braces" : "{$p}"}.braces,
+  "&amp;&quot;",
```
ghislainfourny, line 5: "&" should also parse, because in a string literal in the JSONiq syntax, & is a character like any other.

src/test/resources/test_files/xml/DirectAttributeEscapeBoundary.jq
```diff
@@ -0,0 +1,4 @@
...
+ data(<e x="\n"/>/@x) eq "\\n")
```
ghislainfourny: This is correct, too. In XML attribute values, \ is a character like any other.

ghislainfourny:
At first sight, it looks great and the behavior is as expected. I will take some more time to go through the changes.

ghislainfourny (approved):
Very impressive, complex yet clean. Very good test coverage, too.

PR #1660, opened 2026-07-14
Fix TranslateFunctionIterator to handle Unicode code points

CaiJimmy:
Previously we casted codepoints to Java chars (16 bits), but unicode code points can be more than 16 bits. This PR replaces char with integer, and uses Java's `codePoints()` method to do the conversion.

This was extracted from #1532


ghislainfourny:
@HenrikPaetzold do you want to review, since you worked a lot with codepoints?

HenrikPaetzold:
It looks good! :)

ghislainfourny (approved):
@CaiJimmy if you want to be the one crossing the 26,000 mile, now is your opportunity!

CaiJimmy:
Thanks @ghislainfourny and @HenrikPaetzold for the review!

PR #1662, opened 2026-07-14
Fix handling of apostrophes and quotes in element content.

ghislainfourny:
' and " are allowed in element content, but this is currently not parsing because the XQuery lexer always treats them as starting a string literal. This introduces a state machine to recognize when these characters appear in element content so that they are treated as regular characters.

### DIFF
--- a/src/main/java/org/rumbledb/parser/xquery/XQueryLexer.g4
+++ b/src/main/java/org/rumbledb/parser/xquery/XQueryLexer.g4
@@ -36,7 +36,6 @@ tokens {EscapeQuot, EscapeApos, DOUBLE_LBRACE, DOUBLE_RBRACE}
     // characters inside direct element content. We therefore keep a small
     // lexer-side state to know when we are in direct element content and
     // should not enter string-literal modes on ' or ".
-    private int lastDefaultTokenType = Token.INVALID_TYPE;
     private int directTagNameStartOffset = -1;
     private boolean pendingDirectTag = false;
     private boolean insideDirectTag = false;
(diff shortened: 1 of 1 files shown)


ghislainfourny:
@CaiJimmy it would be great if we could also make the same change in the JSONiq lexer (unless it does not have this issue), probably after merging your other fix for attribute values. The two state machines might interfere so we need to make sure both work.

src/main/java/org/rumbledb/parser/xquery/XQueryLexer.g4
```diff
@@ -29,6 +29,117 @@ tokens {EscapeQuot, EscapeApos, DOUBLE_LBRACE, DOUBLE_RBRACE}
...
     // for counting braces inside string literals
     private int bracesInside = 0;
+
+    // Direct element content follows the XQuery 3.1 grammar:
+    // ElementContentChar ::= Char - [{}<&]
+    // so both apostrophes and quotation marks are valid plain content
+    // characters inside direct element content. We therefore keep a small
+    // lexer-side state to know when we are in direct element content and
+    // should not enter string-literal modes on ' or ".
+    private int lastDefaultTokenType = Token.INVALID_TYPE;
```
CaiJimmy: Is this field used? Seems unused.

ghislainfourny: It was removed :) Thank you for the catch!

CaiJimmy:
This pull request can be closed as the bug has been fixed in #1666

PR #1665, opened 2026-07-14
Fix datetime related bugs

### LINKED ISSUE #831: Calling currentTime() multiple times in same query

### LINKED ISSUE #1287: xs:dayTimeDuration("P31D") should not be equal to xs:yearMonthDuration("P1M")
Right now rumble considers them equal but according to the test [op/duration-equal.xml] K-DurationEQ-32 they should be different

HenrikPaetzold:
Fixes #1287: xs:dayTimeDuration/xs:yearMonthDuration eq now compares months and day-time components independently instead of anchoring both to a fixed reference date. 
Fixes #831: DynamicContext now serializes currentDateTime via Kryo so current-dateTime() stays stable across executors.


ghislainfourny (commented):
Great fix, congratulations :)

ghislainfourny (approved):

PR #1666, opened 2026-07-14
Move string and attribute context handling from lexer to parser rules

CaiJimmy:
- Move JSONiq and XQuery string/attribute context handling from lexer state machines into parser rules. This made the lexer code reduce massively because it does not longer need to guess the meaning of quote and apos.
- This fixed problem with apostrophes and quotes in attributes and element content.
- Add shared utilities for string decoding, direct-constructor processing, hidden-token recovery, and adjacent text merging.
- Add regression coverage for quotes, apostrophes, escapes, comments, entities, nested constructors, and invalid tag/attribute syntax.

The role of lexer is much simpler now, it only emits token. And we have two modes: DEFAULT_MODE and STRING_MODE (for backtick template strings)


src/main/java/org/rumbledb/parser/xquery/XQueryLexer.g4
```diff
@@ -57,10 +45,15 @@ PredefinedEntityRef: '&' ('lt'|'gt'|'amp'|'quot'|'apos') ';' ;
...
+// a token lets the parser report the standard syntax error instead of
+// silently treating it as literal text.
+Ampersand: '&' ;
+
 // Escapes are handled as two Quot or two Apos tokens, to avoid maximal
 // munch lexer ambiguity.
-Quot        : '"' -> pushMode(QUOT_LITERAL_STRING);
-Apos        : '\'' -> pushMode(APOS_LITERAL_STRING);
+Quot        : '"';
+Apos        : '\'';
```
CaiJimmy, line 56: Lexer will emit Quot and Apos as it is, and parser will decide their meaning depending on the context

src/main/java/org/rumbledb/parser/xquery/XQueryParser.g4
```diff
@@ -431,44 +430,50 @@ argument: exprSingle | QUESTION ;
 
 nodeConstructor: directConstructor | computedConstructor ;
 
-directConstructor: dirElemConstructorOpenClose
-                 | dirElemConstructorSingleTag
-                 | (COMMENT | PI)
-                 ;
+// Keep the shared start-tag prefix outside the open/close and self-closing alternatives.
+directConstructor
+    : LANGLE open_tag_name=qname attributes=dirAttributeList
```
CaiJimmy, line 435: This part was shared between `dirElemConstructorOpenClose` and `dirElemConstructorSingleTag`, this change just makes them share the same prefix

src/main/java/org/rumbledb/parser/xquery/XQueryParser.g4
```diff
@@ -431,44 +430,50 @@ argument: exprSingle | QUESTION ;
...
+                     | dirAttributeValueApos
                      ;
 
-dirAttributeContentQuot : contentChar                     
-                        | DOUBLE_LBRACE | DOUBLE_RBRACE
-                        | dirAttributeValueApos
+dirAttributeContentQuot : LBRACE LBRACE | RBRACE RBRACE
                         | LBRACE expr? RBRACE
+                        | ~(Quot | LBRACE | RBRACE | Ampersand | PredefinedEntityRef | CharRef
+                            | LANGLE | COMMENT | XMLDECL | PI | CDATA)
```
CaiJimmy, line 466: Illegal tokens inside a attribute value

ghislainfourny (approved):
Great work. I think the approach is sensible.

PR #1670, opened 2026-07-16
Add ResourceResolver and ModuleImportLoader classes to centralize URI and module import logic

CaiJimmy:
ResourceResolver handles how URIs are resolved, and provides API to override. Rumble API has been modified to accept an instance of this class.

`CompilationConfiguration` class has been added to wrap RumbleRuntimeConfiguration + ResourceResolver class.

This is because in `next` branch, we will also have `ExternalBindings` class, and it will go there.

### DIFF
--- a/src/main/java/org/rumbledb/compiler/URILiteralUtils.java
+++ b/src/main/java/org/rumbledb/compiler/URILiteralUtils.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0.
+ */
+
+package org.rumbledb.compiler;
+
+import org.rumbledb.exceptions.CannotRetrieveResourceException;
+import org.rumbledb.exceptions.ExceptionMetadata;
+import org.rumbledb.exceptions.InvalidURILiteralException;
+import org.rumbledb.runtime.functions.input.FileSystemUtil;
+
+import java.net.URI;
+
+final class URILiteralUtils {
+
+    private URILiteralUtils() {
+    }
+
+    static URI resolve(URI baseURI, String literal, ExceptionMetadata metadata) {
+        try {
+            return FileSystemUtil.resolveURI(baseURI, literal, metadata);
+        } catch (CannotRetrieveResourceException exception) {
+            InvalidURILiteralException result = new InvalidURILiteralException(
+                    "Invalid URI literal: " + literal,
+                    metadata
+            );
+            result.initCause(exception);
+            throw result;
+        }
+    }
+}
--- a/src/main/java/org/rumbledb/runtime/functions/input/FileSystemUtil.java
+++ b/src/main/java/org/rumbledb/runtime/functions/input/FileSystemUtil.java
@@ -76,6 +76,12 @@ public class FileSystemUtil {
     }
 
     private static URI parseURIReference(String value) throws URISyntaxException {
+        try {
+            return new Path(value).toUri();
+        } catch (HadoopIllegalArgumentException e) {
+            // Fall back to Java URI parsing for URI references Hadoop Path cannot parse.
+            // For example, urn: is a valid URI reference but not a valid Hadoop Path.
+        }
         String escapedValue = value.replace(" ", "%20");
         URI uri = new URI(escapedValue);
         if (!value.equals(escapedValue) && uri.isOpaque()) {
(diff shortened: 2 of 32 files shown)


src/main/java/org/rumbledb/runtime/functions/input/FileSystemUtil.java
```diff
@@ -53,8 +54,7 @@ public static URI resolveURI(URI base, String url, ExceptionMetadata metadata) {
             );
         }
         try {
-            Path relativePath = new Path(url);
-            URI relativeURI = relativePath.toUri();
+            URI relativeURI = parseURIReference(url);
             URI resolvedURI = base.resolve(relativeURI);
             if (url.endsWith("/")) {
                 // preserve trailing slash if any for correct resolution against it as a directory in the future.
```
CaiJimmy, line 68: Hadoop path does not support opaque URI like: `urn:isbn:0451450523`, so we switched to Java URI implementation. The only catch is that Java URI does not support space in URI, so it has to be encoded (replace with `%20`)

src/main/java/org/rumbledb/compiler/URILiteralUtils.java
```diff
@@ -0,0 +1,34 @@
...
+package org.rumbledb.compiler;
+
+import org.rumbledb.exceptions.CannotRetrieveResourceException;
+import org.rumbledb.exceptions.ExceptionMetadata;
+import org.rumbledb.exceptions.InvalidURILiteralException;
+import org.rumbledb.runtime.functions.input.FileSystemUtil;
+
+import java.net.URI;
+
+final class URILiteralUtils {
```
CaiJimmy: A separate URI Literal Utils class for compiler, because the error code of compile time invalid URI is different than runtime (XQST0046 vs FODC0002)

CaiJimmy:
@ghislainfourny the two regressions are false negative. Those tests have a flag:

```
      <dependency type="feature" value="serialization" satisfied="false"/>
```

Since RumbleDB does support serialization, these tests should be skipped according to standard:

```
                  
                  <p>The attribute setting <code>satisfied="false"</code> indicates that the test should only
                     be run if the dependency is NOT satisfied.
                  </p>
                  
```

Previously we did not check this: https://github.com/RumbleDB/rumble-test-suite/pull/50/changes#diff-475625265e96f444fd2b33ba014d8be96709adbe243b9ed2a9c81c1d623b7021R326-R328

src/main/java/org/rumbledb/runtime/functions/input/FileSystemUtil.java
ghislainfourny: Thanks! We used to do it this way (fix %20 manually and use Java URIs) but it was very cumbersome and error prone especially with Windows, Linux, MacOS, slashes, backslashes, etc. Since the tests run on Linux, any change is quite risky and could come back to bite us later (ETH exams are on Windows, many jsoniq pip users use macOS). With Hadoop Path we finally managed to resolve the issues. 
It would be nice if we could keep the Hadoop Path by default and only fall back to Java URIs for URIs that Hadoop does not support.
Also, since this class is for fetching resources from the file system, does it make sense to have a urn-schemed URI passed to this class while expecting it to succeed?

ghislainfourny:
> @ghislainfourny the two regressions are false negative. Those tests have a flag:
> 
> ```
>       <dependency type="feature" value="serialization" satisfied="false"/>
> ```
> 
> Since RumbleDB does support serialization, these tests should be skipped according to standard:
> 
> ```
>                   
>                   <p>The attribute setting <code>satisfied="false"</code> indicates that the test should only
>                      be run if the dependency is NOT satisfied.
>                   </p>
>                   
> ```
> 
> Previously we did not check this: https://github.com/RumbleDB/rumble-test-suite/pull/50/changes#diff-475625265e96f444fd2b33ba014d8be96709adbe243b9ed2a9c81c1d623b7021R326-R328

Good catch!

ghislainfourny:
Do the qt3tests heavily test location hints?

CaiJimmy:
> Do the qt3tests heavily test location hints?

No, they do have some queries with location hints but in general they don't use that feature.

```
  <p>Generally, <code>import module</code> declarations within a query are written without
    any location hints, except where the purpose of the test is to ensure that location hints
    can be processed. If an implementation provides no way of resolving an <code>import module</code>
    declaration without a location hint, then the test driver may modify the query to add a location
    hint before processing.</p>
```

src/main/java/org/rumbledb/runtime/functions/input/FileSystemUtil.java
CaiJimmy: I think the problem is that we are combining two methods. During compilation, declarations such as 'base-uri' use the same method and can cause a parse error, as this can be 'urn:'. During runtime, methods such as 'json-file' also use the same method and expect a valid file system path.

I have therefore separated them into two methods: 'resolveURI' and 'resolveFileSystemURI'.

ghislainfourny (approved):
Great work! I think you may be the one hitting 27,000 this week!

PR #1674, opened 2026-07-17
Fix assert-serialization-error in test suite

ghislainfourny:


ghislainfourny:
@CaiJimmy these tests are intended for XPath, not XQuery and are now skipped. XQuery is generally a superset of XPath, but there are a few subtle differences.

PR #1681, opened 2026-07-18
Add `legacy` option to `fn:parse-json`, `fn:json-doc`

HenrikPaetzold:
Adds a RumbleDB-specific `legacy` option to `fn:json-doc` and `fn:parse-json`'s options map. It requires every other option to be left unset or at its default; when set, parsing is routed through the legacy Gson-based parser instead of RumbleDB's own spec-conformant parser. This can produce unexpected results for duplicate object keys and invalid codepoints, a warning is therefore emitted whenever `legacy` is used.

Users that already relied on the Gson-based parser's behavior before this rewrite can opt back into it via the legacy option.


ghislainfourny (approved):
It looks great! Thank you

PR #1696, opened 2026-07-20
Function deepcopy optimization

CaiJimmy:
The bottleneck is the RuntimeIterator.deepCopy operation, which uses Java serialization and deserialization to create independent copies of the function.

Many of our tests involve recursive or large function calls, which cause this operation to slow everything down by becoming the bottleneck. 

The PR addresses this problem by:

1. Reusing the same iterator if possible
2. Serializing the body only once at the beginning and deserializing it later.

With this PR, we reduced the **QT3 runtime in CI from 13 minutes to 5 minutes**, as well as the Maven test time from 14 minutes to 7 minutes.

This is not the final solution. The problem is that RuntimeIterator mixes mutable and immutable states within the same object. 

In the future, we want to make the iterator immutable and keep a "cursor" object for the local volcano iterator. With that change, serialization for copying will no longer be necessary. The iterator itself will become immutable, and the copy operation will create a new instance of the cursor (which should be small and only keep states).


ghislainfourny (approved):
I am unsure whether this could cause regressions with functions that are ML estimators and transformers, but since all ML tests pass I would go for it. If we discover bugs in ML later, we will fix them. Thanks for taking the initiative!

PR #1705, opened 2026-07-21
Remove Serializable and KryoSerializable implementations from compiler-only classes

CaiJimmy:
Similar to StaticContext, these classes are not part of serialization graph, so implementing Serializable does not make any sense (adding unused codes)


ghislainfourny:
Good catch. Are we positive they are nowhere in the transitive closure of any serializable runtime iterator for example?

ghislainfourny (approved):

PR #1710, opened 2026-07-22
Implement `fn:innermost`, `fn:outermost`, `fn:lang`

HenrikPaetzold:
Adds implementation for `fn:innermost`, `fn:outermost`, `fn:lang`.

These nodes require access to their parent nodes, which is achieved by a compile-time scan (ParentPointerAnalysisVisitor) for calls to these functions, disabling the parent-pointer-removal optimization when one is found.

Also re-commits a fix for `fn:unparsed-text-lines` that seems to have gone under.


ghislainfourny (approved):
Great work!

PR #1713, opened 2026-07-22
Drop reset method in iterator interface

CaiJimmy:
This method was never used, and equivalent to `close + open`


ghislainfourny (approved):
936 deleted lines and only one added, this is the best PR I have seen in a very long time!

PR #1715, opened 2026-07-22
Add missing `@Override` annotation

CaiJimmy:
Annotating methods with @java.lang.Override improves code readability since it shows the intent. In addition, the compiler **emits an error when a signature of the overridden method doesn't match the superclass method.**

https://docs.oracle.com/javase/8/docs/api/java/lang/Override.html


ghislainfourny (approved):
Very good catch.

ghislainfourny:
Is there a way to configure the linter to fail if an Override is missing?

CaiJimmy:
> Is there a way to configure the linter to fail if an Override is missing?

Yes, I previously found this tool: https://errorprone.info/

We can add Google Java Format linter and this once next branch is merged to master, so we don't need to deal with big merge conflicts

PR #1716, opened 2026-07-22
Cleanup RuntimeStaticContext and switch to builder pattern

CaiJimmy:
This pull request refactors RuntimeStaticContext class to make it use Lombok `@Builder` annotation, because the number of fields is growing larger and adding more arguments to the constructor is not very elegant.

The class is now immutable. To create a new instance, use `.toBuilder().build()`.

A `fromStaticContext` static method centralizes the logic of conversion from StaticContext to RuntimeStaticContext. 

`withoutDecimalFormats`, `withStaticType`, `withMetadata` and `withExecutionMode` methods are just trivial wrappers around builder method. Will be removed in the following PR.

This PR prepares for moving isUpdating and isSequential fields from RuntimeIterator to here.


ghislainfourny (commented):
Clean and neat!

ghislainfourny (approved):

PR #1729, opened 2026-07-23
Make `children` field in RuntimeIterator private

CaiJimmy:
Previously, this field was protected, so all subclasses can read and write it freely. However, this allowed patterns like this in subclass constructors:

```java
super(null, staticContext);    /// The first argument is List<RuntimeIterators> children
this.children.add(tryExpression);
for (RuntimeIterator value : catchExpressions.values())
    this.children.add(value);
```

To fix the pattern, we no longer allow subclasses to read `children` field directly, and force them to use the superclass' constructor so we get a better encapsulation.

`getChild(int)` and `getChildren()` are added to RuntimeIterator, both marked as `protected final`

### DIFF
--- a/src/main/java/org/rumbledb/runtime/RuntimeIterator.java
+++ b/src/main/java/org/rumbledb/runtime/RuntimeIterator.java
@@ -23,10 +23,7 @@ package org.rumbledb.runtime;
 import java.io.*;
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
+import java.util.*;
 
 import lombok.NonNull;
 import org.apache.spark.api.java.JavaRDD;
@@ -82,7 +79,7 @@ public abstract class RuntimeIterator implements RuntimeIteratorInterface<Item>,
         this.isOpen = false;
         this.isUpdating = false;
         this.isSequential = false;
-        this.children = children;
+        this.children = List.copyOf(Objects.requireNonNullElse(children, Collections.emptyList()));
     }
 
     /**
--- a/src/main/java/org/rumbledb/runtime/RuntimeIterator.java
+++ b/src/main/java/org/rumbledb/runtime/RuntimeIterator.java
@@ -67,7 +67,7 @@ public abstract class RuntimeIterator implements RuntimeIteratorInterface<Item>,
     protected transient boolean isOpen;
     protected boolean isUpdating;
     protected transient boolean isSequential;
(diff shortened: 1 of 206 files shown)


src/main/java/org/rumbledb/runtime/control/IfRuntimeIterator.java
```diff
@@ -46,10 +47,10 @@ public IfRuntimeIterator(
             boolean isUpdating,
             RuntimeStaticContext staticContext
     ) {
-        super(null, staticContext);
-        this.children.add(condition);
-        this.children.add(branch);
-        this.children.add(elseBranch);
+        super(
+            List.of(condition, branch, elseBranch),
```
CaiJimmy, line 51: Using `List.of` creates an immutable copy. In the RuntimeIterator constructor, `List.copy` is used again, but it won't copy it again because it can detect that the argument is already unmodifiable.

src/main/java/org/rumbledb/runtime/control/TypeswitchRuntimeIterator.java
```diff
@@ -34,12 +36,14 @@ public TypeswitchRuntimeIterator(
...
-        }
-        this.children.add(defaultCase.getReturnIterator());
+        super(
+            Stream.of(
+                Stream.of(test),
+                cases.stream().map(TypeswitchRuntimeIteratorCase::getReturnIterator),
+                Stream.of(defaultCase.getReturnIterator())
+            ).flatMap(Function.identity()).toList(),
+            staticContext
+        );
```
CaiJimmy, line 46: In case of multiple arguments, we use `Stream` API (JDK 8+). Calling `.toList()` also creates an immutable list.

src/main/java/org/rumbledb/runtime/RuntimeIterator.java
```diff
@@ -81,11 +79,7 @@ protected RuntimeIterator(List<RuntimeIterator> children, RuntimeStaticContext s
         this.isOpen = false;
         this.isUpdating = false;
         this.isSequential = false;
-
-        this.children = new ArrayList<>();
-        if (children != null && !children.isEmpty()) {
-            this.children.addAll(children);
-        }
+        this.children = List.copyOf(Objects.requireNonNullElse(children, Collections.emptyList()));
```
CaiJimmy, line 82: In most of cases (I think should be all cases but I might have missed), this will not perform a copy because we always pass an immutable list.

So some performance gain we might get compared to previous versions, which always performs a explicit copy.

ghislainfourny (approved):
LGTM, useful change.

PR #1730, opened 2026-07-24
Remove the KryoSerializable implementation

CaiJimmy:
By default, Kryo uses a FieldSerializer to automatically serialise all the fields of a class using Java reflection. Therefore, we don't need to implement the KryoSerializable interface unless there are very specific performance reasons, such as controlling the binary format (which is not the case here).

In many classes, we are doing exactly the same as the default behaviour. In some cases, fields are forgotten, which is difficult to detect at compile time and can introduce errors at runtime.

`Serializable` interface from Java is kept because we still use it for other things, and it's a fallback for Kryo serializer. 

Another advantage we get is that the behavior is consistent with Java Serializable: fields marked as `transient` are also not serialized in Kryo.

----

Also, some errors appeared during this refactor, meaning that the overridden Kryo serialisation behaviour was not consistent with the Java serialisation.

- https://github.com/RumbleDB/rumble/pull/1730/changes/28ab1d2df97e2110c1913c44c430b0f44e29dfcd : `allowedPrefixes` was not serialized in Kryo but in Java Serializable (well, in fact only `this.arguments` was serialized in Kryo while all other fields were considered for Java serialization)
-  https://github.com/RumbleDB/rumble/pull/1730/changes/0e4156081be88c3e2ed8d004f3cae7018edfbeec : FunctionSignature was part of serialization graph via DynamicContext, but it never implemented no-arg constructor because it was not included in the `write` method


ghislainfourny (approved):
This is a very interesting insight. Somehow we had worked under the assumption we needed to tune the behavior to our needs.

PR #1731, opened 2026-07-24
Replace trivial `hashCode` and `equal` methods with Lombok annotation

CaiJimmy:
Avoid a lot of boilerplate code, and we also guarantee that the behavior between equal and hashcode remains the same.

Remaining problem: we have many classes like CommentItem that implements custom `equals` but no `hashCode`. This might lead to strange bugs...: 

image


src/main/java/org/rumbledb/runtime/flwor/FlworDataFrameColumn.java
```diff
@@ -85,37 +88,6 @@ public FlworDataFrameColumn(String columnName, StructType inputSchema) {
...
-            return false;
-        }
-        if (!this.variableName.equals(other.variableName)) {
-            return false;
-        }
-        if (!this.columnFormat.equals(other.columnFormat)) {
-            return false;
-        }
-        return true;
-    }
```
CaiJimmy, line 109: Basically comparing all fields

src/main/java/org/rumbledb/types/SequenceType.java
```diff
@@ -259,20 +262,6 @@ public SequenceType refineArityIfSubtype(Arity otherArity) {
...
-        if (!(other instanceof SequenceType sequenceType)) {
-            return false;
-        }
-        if (isEmptySequence()) {
-            return sequenceType.isEmptySequence();
-        }
-        if (sequenceType.isEmptySequence()) {
-            return false;
-        }
-        return this.getItemType().equals(sequenceType.getItemType()) && this.getArity().equals(sequenceType.getArity());
```
CaiJimmy, line 273: These two ifs are not necessary

ghislainfourny (approved):

PR #1732, opened 2026-07-24
Consistent hashing and equal methods

CaiJimmy:
Follow up of #1731 

Refine item comparison and hashing to make same-key semantics consistent across atomic, structured, and sequence values. Done by creating multiple Abstract classes for these types that defines `hashCode` and `equal` methods in a centralized way.

This fixes collation-aware string comparison, Gregorian temporal keys, numeric equality/hash behavior, and edge cases in arrays and objects. And many duplicated codes are removed.

Also preserves concise decimal lexical values when values cross DataFrame boundaries, avoiding floating-point artifacts in output, and removes an unsupported DataFrame implementation of distinct-values so it consistently uses XDM-correct RDD semantics.

With consistent hashing code implementation, we can now use HashMap and HashSet instead of TreeMap and TreeSet and get a better performance (O(1) lookup)


src/test/resources/test_files/runtime/Collation6.jq
```diff
@@ -3,4 +3,5 @@ declare default collation "http://www.w3.org/2005/xpath-functions/collation/code
 
 for $i in ("foo", "bar")
 group by $j := $i collation "http://www.w3.org/2005/xpath-functions/collation/codepoint"
+order by $j descending
```
CaiJimmy, line 6: The output order was not deterministic because we did not specific this. And it triggered an error when I changed the hashCode implementation.

src/main/java/org/rumbledb/items/xml/ElementItem.java
```diff
@@ -147,14 +147,6 @@ public boolean isElementNode() {
         return true;
     }
 
-    @Override
-    public boolean equals(Object other) {
-        if (!(other instanceof ElementItem otherElementItem)) {
-            return false;
-        }
-        return this.getXmlDocumentPosition().equals(otherElementItem.getXmlDocumentPosition());
```
CaiJimmy, line 155: We did not check if `position` is `null` here.

src/main/java/org/rumbledb/items/ArrayItem.java
```diff
@@ -81,24 +81,6 @@ public Item copy(boolean mutable) {
...
-            return false;
-        }
-        for (int i = 0; i < getSize(); ++i) {
-            if (!getItemAt(i).equals(otherItem.getItemAt(i))) {
-                return false;
-            }
-        }
-        return true;
-    }
-
```
CaiJimmy, line 101: Moved to AbstractArrayItem

src/main/java/org/rumbledb/items/SequenceArrayItem.java
```diff
@@ -69,32 +69,6 @@ public Item copy(boolean mutable) {
...
-            }
-            for (int j = 0; j < thisMember.size(); j++) {
-                if (!thisMember.get(j).equals(otherMember.get(j))) {
-                    return false;
-                }
-            }
-        }
-        return true;
-    }
-
```
CaiJimmy, line 97: Moved to SequenceArrayItem.

src/main/java/org/rumbledb/items/SequenceArrayItem.java
```diff
@@ -69,32 +69,6 @@ public Item copy(boolean mutable) {
...
-            List<Item> thisMember = this.getSequenceAt(i);
-            List<Item> otherMember = otherItem.getSequenceAt(i);
-            if (thisMember.size() != otherMember.size()) {
-                return false;
-            }
-            for (int j = 0; j < thisMember.size(); j++) {
-                if (!thisMember.get(j).equals(otherMember.get(j))) {
-                    return false;
-                }
-            }
```
CaiJimmy, line 93: This is equivalent to `thisMember.equals(otherMember)` because `List<T>.equal` does that

src/main/java/org/rumbledb/items/LazyObjectItem.java
```diff
@@ -105,35 +105,6 @@ public Item copy(boolean mutable) {
...
-        }
-        for (String s : otherItem.getStringKeys()) {
-            Item v = getItemByKey(s);
-            if (v == null) {
-                return false;
-            }
-            if (!otherItem.getItemByKey(s).equals(v)) {
-                return false;
-            }
-        }
```
CaiJimmy, line 132: Moved to `AbstractMapItem.equals`. Also, two loop was not necessary, we only need to do it once + size check

src/main/java/org/rumbledb/items/MapItem.java
```diff
@@ -59,7 +58,7 @@ public class MapItem implements Item {
     public MapItem() {
         this.values = new ArrayList<>();
         this.keys = new ArrayList<>();
-        this.keyToIndex = new TreeMap<>(new ItemSameKeyComparator());
+        this.keyToIndex = new HashMap<>();
```
CaiJimmy, line 61: Now that we have a stable hashCode method, we can make use of HashMap, which is significantly faster than TreeMap

src/main/java/org/rumbledb/items/AtomicItemEquivalence.java
```diff
@@ -0,0 +1,146 @@
...
+import org.rumbledb.api.Item;
+
+/**
+ * Defines the stable Java equality and hashing contract for atomic items.
+ *
+ * Query value comparisons are intentionally not used here: they can be context-dependent, can
+ * raise dynamic errors, and do not form an equivalence relation. The XDM {@code op:same-key}
+ * relation is stable and is already the equality required by maps.
+ */
+public final class AtomicItemEquivalence {
```
CaiJimmy, line 31: Logic extracted from src/main/java/org/rumbledb/items/ItemSameKeyComparator.java

src/main/java/org/rumbledb/items/MapWithRemovedEntryItem.java
```diff
@@ -54,7 +54,7 @@ public MapWithRemovedEntryItem() {
 
     public MapWithRemovedEntryItem(Item original, List<Item> removedKeys) {
         this.original = original;
-        this.removedKeys = new TreeSet<>(new ItemSameKeyComparator());
+        this.removedKeys = new HashSet<>();
```
CaiJimmy, line 57: Same here, now we can use HashSet instead of TreeSet for better performance

src/main/java/org/rumbledb/types/AbstractItemType.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public final boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+        if (!(other instanceof AbstractItemType otherType)) {
+            return false;
+        }
+        return this.equalityKey().equals(otherType.equalityKey());
+    }
```
CaiJimmy, line 46: If equalityKey is not overridden, it will compare by name (old behavior)

src/main/java/org/rumbledb/items/MapItem.java
ghislainfourny: It sounds good. Can you confirm it does not change the order of display of the keys? While not semantically relevant, it is important for user-experience.

CaiJimmy: > It sounds good. Can you confirm it does not change the order of display of the keys? While not semantically relevant, it is important for user-experience.

It can change the order of display of keys by default. In fact, some test cases has been updated to include a `order by` clause so the output is deterministic. For example `src/test/resources/test_files/runtime-spark/LocalClauses/GroupbyClauseLocal2.jq`.

ghislainfourny (approved):

PR #1733, opened 2026-07-25
Fix space between XML elements

CaiJimmy:
Test suite PR: https://github.com/RumbleDB/rumble-test-suite/pull/60

By default, the `declare boundary-space` setting should be `strip` (i.e. do not preserve boundary space). This PR fixes the spacing issue between elements.

Also, the test suite was not checking the XML assertion correctly (if the `file` attribute is provided, we should compare the file, which was not being done previously). Regressions are expected because we were previously performing the comparison incorrectly. Specifically, these **three regressions are cases that failed previously, but have now appeared because we have fixed the test suite check.**

- app/Demos.xml:currency-svg
- app/XMark.xml: XMark-All
- prod/VarDecl.external.xml:extvardeclwithtype-23


ghislainfourny:
Do we need to merge a branch in the rumble-test-suite to get rid of the three regressions?

ghislainfourny (approved):

PR #1746, opened 2026-07-27
Add `final` modifier on instance fields

CaiJimmy:
More code safety. This is just a quick refactor done by IntelliJ. It does not cover all fields, because there might be setter methods that are actually unused.


ghislainfourny (approved):
LGTM!

PR #1748, opened 2026-07-28
Reject invalid regex syntax in fn:matches (FORX0002)

HenrikPaetzold:
fn:matches accepted multiple regex constructs that are valid in Java but illegal per the XPath F&O grammar, so no error was raised where the QT3 suite expects one.


HenrikPaetzold:
@ghislainfourny Out of curiosity: since when has AppTest no longer taken 9 minutes to run? I believe there was a sudoku test that took forever. What was done to mitigate this?

ghislainfourny:
> @ghislainfourny Out of curiosity: since when has AppTest no longer taken 9 minutes to run? I believe there was a sudoku test that took forever. What was done to mitigate this?

Indeed! Jimmy optimized with elegant findings on deep function copy and Kryo serialization 😄

HenrikPaetzold:
Very nice!

ghislainfourny (approved):
Great job!

PR #1750, opened 2026-07-28
Replace trivial getters and setters with Lombok annotations and remove unused methods and classes

CaiJimmy:
Remove unused classes and methods, and replace no-arg constructor with Lombok `@NoArgsConstructor` and trivial getters and setters with `@Getter` and `@Setter`.


src/main/java/org/rumbledb/api/SequenceWriter.java
```diff
@@ -40,7 +40,7 @@
  * </ul>
  *
  * The serialization method (json, tyson, adaptive, xml-json-hybrid, yaml, delta, ...) is always taken from
- * {@link SerializationParameters#getMethod()}, which is the single source of truth for the output
+ * {@code SerializationParameters#getMethod()}, which is the single source of truth for the output
```
CaiJimmy, line 43: Javadoc cannot find this method because we have replaced it with Lombok @Getter

ghislainfourny (approved):

CaiJimmy:
I'll hold this PR for a bit. Because once replaced getter with annotation, Intellij cannot longer find unused getter declaration easily. Maybe we should do a unused declaration cleanup before this.

PR #1767, opened 2026-07-31
Henrik: Final Report related performance and correctness fixes

HenrikPaetzold:
Below are numerous changes made to reduce the per-item computation time of `format-integer` and `codepoints-to-string`.

### Performance fixes:
- **`applyGrouping`** reversed the digit string twice, allocated the separator on every
  call, and appended character by character. Now one pass, sized up front.
- **`IntegerPictureFormatter`** resolved the ICU locale before dispatching on the picture,
  although only the word, ordinal and fallback paths ever read it. Now resolved at the
  point of use.
- **`VariableValues.getLocalVariableValue`** did four hash lookups for the same key on the
  success path. Now one.
- **`NamedFunctions`** wrapped every builtin argument in conversion/promotion iterators that
check isSubtypeOf per item, even when the static type already matched exactly. Now
skipped when the argument is already atomic, not untypedAtomic, and a subtype of the
parameter type.
- **`CodepointsToStringFunctionIterator`** went through getIntegerValue().intValueExact() for
every codepoint, allocating a BigInteger even when the value already fit an int. Now
isInt()/getIntValue() short-circuit that.

### Correctness Fixes
- **`StringLengthFunctionIterator`** used String.length(), which counts UTF-16 code units, not
codepoints


src/main/java/org/rumbledb/runtime/functions/FunctionCallArgumentConversion.java
```diff
@@ -128,16 +128,19 @@ public static RuntimeIterator wrapForFunctionConversion(
             RuntimeStaticContext runtimeStaticContext
     ) {
         ItemType targetItemType = sequenceType.getItemType();
+        // The function conversion rules apply to a generalized atomic type, that is an atomic type or a pure union
+        // type such as xs:numeric (XQuery 3.1, 3.1.5.2 and F&O 3.1, 1.6.3).
+        boolean targetIsGeneralizedAtomic = targetItemType.isAtomicItemType() || targetItemType.isUnionType();
```
ghislainfourny: A pure union type is more restrictive than any union type:
https://www.w3.org/TR/xquery-31/#dt-pure-union-type

I do not think we support list types at this point though.

ghislainfourny (changes requested):
Great work! Just a slight thing to adapt related to the definition of a pure union type.

src/main/java/org/rumbledb/runtime/functions/FunctionCallArgumentConversion.java
HenrikPaetzold: I did not see that. Thank you!

ghislainfourny (approved):

PR #1772, opened 2026-08-02
Add runtime plan and cursor definitions

CaiJimmy:
This pull request adds type definitions for cursors and plans, replacing the current iterator model that mixes mutable (cursor) and immutable (fields like children) states.

With this new model, the `deepCopy` method, which copies an iterator, is no longer necessary. This was previously a bottleneck in the item function, resulting in a long execution time in QT3 tests (#1696).

Another advantage is that we no longer need to worry about converting between local cursors, RDDs, and dataframes. Conversion will now be done transparently inside RuntimePlan, and each iterator can freely implement its capabilities.

## Cursors

| Class                        | Usage                                                     |
| ---------------------------- | --------------------------------------------------------- |
| Cursor                       | Core evaluation-state API                                 |
| AbstractLocalCursor          | Single-use lifecycle and failure cleanup                  |
| AtMostOneLocalCursor         | When the context supplies one input                       |
| ConcatLocalCursor            | Streaming comma-expression concatenation                  |
| ContextOrArgumentLocalCursor | Still supports multi-result context/argument functions    |
| FlatMappingLocalCursor       | Used by array, object and navigation streaming operations |
| IteratorLocalCursor          | General Java iterator and RDD-materialization adapter     |
| MappingLocalCursor           | Streaming type validation and casting                     |

## Plans

| Class                        | Usage                                                                                                                                                              |
| ---------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| RuntimePlan<T>               | Represents a runtime plan for executing a specific type T. Contains methods `getCursor()`, `getRDD()`, and `getDataframe()` and does the conversion transparently. |
| ItemRuntimePlan              | Specialized runtime plan for Item types, providing additional methods for item-specific operations like `isSparkJobNeeded()` and `getVariableDependencies()`.      |
| LocalRuntimePlan<T>          | Interface that indicates the capability to return a native cursor for type T                                                                                       |
| RDDRuntimePlan<T>            | Interface that indicates the capability to return an RDD for type T                                                                                                |
| DataFrameRuntimePlan<T>      | Interface that indicates the capability to return a DataFrame for type T                                                                                           |
| AtMostOneLocalRuntimePlan<T> | Interface that indicates the plan can return at most one local runtime plan for type T                                                                             |
| NativeQueryRuntimePlan       | Plan can return a native query for execution                                                                                                                       |
| UpdatingRuntimePlan          | Plan can return an updating runtime plan                                                                                                                           |


ghislainfourny (changes requested):
It looks great! This goes 99% of the way.

I have performance improvements to suggest to avoid too many method calls in the chain. We can discuss in today's call.

src/main/java/org/rumbledb/runtime/plan/RuntimePlan.java
```diff
@@ -0,0 +1,403 @@
...
+     * Evaluates this plan locally and returns its only result, or {@code null} for an empty sequence.
+     *
+     * @param context the dynamic context for the evaluation
+     * @return the only result, or {@code null}
+     * @throws MoreThanOneItemException if more than one result is produced
+     */
+    public final T materializeAtMostOne(@NonNull DynamicContext context) throws MoreThanOneItemException {
+        if (this.canEvaluateAtMostOneDirectly()) {
+            return this.atMostOneCapability.evaluateAtMostOne(context);
+        }
```
CaiJimmy, line 297: @ghislainfourny We no longer have `instanceof` per materializeAtMostOne method call.

`canEvaluateAtMostOneDirectly` is a wrapper of `this.atMostOneCapability != null`, and should be inlinable by the compiler because it's private

ghislainfourny (approved):

PR #1784, opened 2026-08-12
Implement schema import and validation expression support in the compiler

CaiJimmy:
This is the first step in bringing full XML schema support to Rumble. 

XQuery parser grammar has been updated to sync the definition of `validateExpr` with the XQuery 3.1. Along with that, a new `ValidateExpression` node type in the intermediate AST.

Runtime support is not yet added, and error is thrown when visiting ValidateExpression in RuntimeIteratorVisitor. This will be done in the following pull requests.

All those regressions are expected, because they use some form of `validate` expression, which until now, they are parsed but not executed. They will be fixed in the future.

Associated test suite update: https://github.com/RumbleDB/rumble-test-suite/pull/67, which adds schema to the list of supported features and rewrites the query correctly so Rumble can find the schema import location.


src/main/java/org/rumbledb/compiler/RuntimeIteratorVisitor.java
```diff
@@ -1480,6 +1482,12 @@ public ItemRuntimePlan visitValidateTypeExpression(ValidateTypeExpression expres
         return resultIterator;
     }
 
+    @Override
+    public ItemRuntimePlan visitValidateExpression(ValidateExpression expression, ItemRuntimePlan argument) {
+        throw new UnsupportedFeatureException(
+                "XML Schema validate expressions are not executable yet.", expression.getMetadata());
+    }
+
     @Override
```
CaiJimmy, line 1491: This will be implemented in the upcoming PRs.

ghislainfourny (approved):
Great work. I am fine with the regressions as this is the not-implemented-yet error.

PR #1800, opened 2026-08-17
Drop old `catch_var` from the parser grammar file

CaiJimmy:
This is not part of W3C XQuery standard, and we actually don't use the variable in the translation visitor, so it's dead.

Now the grammar has been aligned to the official standard:

image


.github/workflows/run-qt3.yml
```diff
@@ -57,6 +57,10 @@ jobs:
               );
             }
 
+            runs.sort(
+              (a, b) => new Date(b.created_at) - new Date(a.created_at)
+            );
+
```
CaiJimmy: It was not picking the latest result from the master branch as baseline for comparison

CaiJimmy:
@ghislainfourny This one is ready to be merged

ghislainfourny (approved):
LGTM

PR #1801, opened 2026-08-20
Extract common rules from JSONiq and XQuery grammar lexer into CommonLexer file

CaiJimmy:
Now that the JSONiq and XQuery grammar are so similar, I propose avoid repeating codes in grammar definitions, 
and use `import` syntax from ANTLR4 to avoid code duplication. 

This PR does it at lexer level, and the next PR will do it at parser level.

This does not affect Java translation level. Code reuse in TranslationVisitor and XQueryTranslationVisitor is more difficult and experimented in #1804, but did not work as I expected.


ghislainfourny (approved):
Great initiative.

PR #1802, opened 2026-08-20
Extract common rules from JSONiq and XQuery grammar parser into CommonParser file

CaiJimmy:
As following up of #1801, this PR does the same duplicated logic removal but in parser files.


ghislainfourny (approved):
LGTM

PR #1804, opened 2026-08-20
Create a single lexer and parser grammar to prepare for Java level code share

CaiJimmy:


CaiJimmy:
The type system in Java is not as flexible as TypeScript and cannot achieve similar code reuse easily. Maybe I will come back to this in the future, for now, a code reuse at ANTLR4 level seems fine.

PR #1805, opened 2026-08-26
Add a public API in VisitorHelpers for parsing library modules

CaiJimmy:
`parseLibraryModuleFromQueryWithStaticContextAndInference`

This is used currently in the language server to parse library modules correctly and obtain the static type. The current `parseLibraryModule` does not infer the type, and is `private`.


ghislainfourny (approved):
Maybe add in the method comment that it's for the language server.

PR #1808, opened 2026-09-02
Implement schema type handling for XML nodes and add NodeTypedValue class

CaiJimmy:
Add new `NodeTypedValue` class which is an immutable representation of an element or attribute node's XDM typed-value state +  validated atomic values.

The typed-value has 3 different states:

- `UNTYPED`: ordinary XML, which will be atomized as xs:untypedAtomic
- `AVAILABLE`: validation suceeded, and we can return the atomized items
- `UNAVAILABLE`: the node has a schema type but cannot expose a typed value; atomization raises FOTY0012, see `atomizedValue` method in `ElementItem`.

## Usage `NodeTypedValue.UNAVAILABLE`

Because we can have a node that does that has type, but does not have a typed value. For example:

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           targetNamespace="urn:people"
           elementFormDefault="qualified">

  <xs:element name="person">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="name" type="xs:string"/>
        <xs:element name="age" type="xs:integer"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

```xquery
xquery version "3.1";

import schema namespace p = "urn:people" at "people.xsd";

let $untyped :=
    <p:person><p:name>Alice</p:name><p:age>42</p:age></p:person>

let $validated := validate strict { $untyped }
return data($validated) (: This will give an FOTY0012 error, because argument to fn:data() contains a node that does not have a typed value. :)
```

> Raised by [fn:data](https://www.w3.org/TR/xpath-functions-31/#func-data), or by implicit atomization, if applied to a node with no typed value, the main example being an element validated against a complex type that defines it to have element-only content.

https://www.w3.org/TR/xpath-functions-31/#ERRFOTY0012

**This pull request does not use this enum value yet.** It will be used later when we introduce Xerces and PSVI support, so we will have cases that a element does not have type even after validation.

## Extra field in `setSchemaType` method in Item API

The signature of `setSchemaType` method in Item API has been modified to add an extra parameter `List<Item> typedValue` so casted values can be passed. Because when a node is validated, we already have the list of type values, no need to repeat the work in the future.

Before:

```
  <value>42</value>
  schema type = xs:integer
  data($node) re-casts "42" every time
```

After:

```
  <value>42</value>
  schema type = xs:integer
  typed value = xs:integer(42)
  data($node) returns the stored value
```


CaiJimmy:
@ghislainfourny should be now ready to be merged. I have updated the PR message to indicate the usage of `NodeTypedValue.unavailable`

ghislainfourny (approved):

ghislainfourny:
Looks good!
Do you have any tests checking that FOTY0012 is throws when applicable?

CaiJimmy:
> Looks good! Do you have any tests checking that FOTY0012 is throws when applicable?

Not in the current PR as it does not have runtime support yet. In QT3 we have some, like:

```xml
   <test-case name="FOTY0012">
      <description> Tests generating FOTY0012 </description>
      <created by="Tim Mills" on="2008-05-16"/>
      <environment ref="foty0012"/>
      <dependency type="spec" value="XQ10+"/>
      <dependency type="feature" value="schemaImport"/>
      <test><![CDATA[import schema namespace foo = "http://www.example.org/2007/schema/FOTY0012"; data(validate strict { <foo:element /> } )]]></test>
      <result>
         <error code="FOTY0012"/>
      </result>
   </test-case>
```

PR #1812, opened 2026-09-05
Implement XML Schema validation support in runtime

CaiJimmy:
This PR connects the XML Schema validation infrastructure to XQuery runtime execution.

It adds runtime support for:
- validate strict
- validate lax
- validate type with imported simple and complex schema types

Built-in atomic validate type expressions continue using the existing `BuiltinTypeValidator`.

The compiler now resolves validate type names against both built-in XML Schema types (from Xerces) and the imported schema catalog. During runtime-plan construction, the validation mode, target type, and schema catalog are passed to `XQueryValidateIterator`, which dispatches to the appropriate validation implementation in runtime.

For strict and lax validation without schema imports, an empty Xerces schema environment is created. This provides built-in definitions such as xs:anyType, allowing lax validation of undeclared elements while preserving normal strict-validation errors.

The schema loader was refactored so imported schemas and the built-in schema environment share the same grammar-pool-to-XSModel conversion logic.


ghislainfourny:
It looks good, can be merged after the two previous PRs is fixed and merged.

ghislainfourny (approved):

ghislainfourny:
Congratulations on passing 29k!

PR #1814, opened 2026-09-07
Support for `xs:IDREFS`,`xs:NMTOKENS` and `xs:ENTITIES`

CaiJimmy:
We support now castings like:

```xquery
let $values := " a  b " cast as xs:IDREFS
return deep-equal($values, ("a", "b")) and $values instance of xs:IDREF+
```


ghislainfourny:
It looks good, can be merged after the four previous PRs is fixed and merged.

ghislainfourny (approved):

PR #1815, opened 2026-09-07
Support built-in list and imported schema constructors

CaiJimmy:
Adds constructor functions for `xs:IDREFS`, `xs:NMTOKENS`, `xs:ENTITIES`, and imported XML Schema simple types. List constructors return typed sequences, while atomic constructors return optional single values.

Example:

```xquery
let $value := t:Count("42")
return $value eq 42 and $value instance of t:Count,
```

The logic of resolving constructor functions has been moved from `BuiltinFunctionCatalogue` to `ConstructorFunctionResolver`. And we pass now the `StaticContext` to `exist` method so we get access to schema catalog (previously, only `queryLanguage` string is passed).


ghislainfourny:
It looks good, can be merged after the five previous PRs is fixed and merged.

ghislainfourny (approved):

PR #1816, opened 2026-09-07
Support schema element and attribute tests

CaiJimmy:
Adds support for schema type annotations in `element(name, type)` and `attribute(name, type)`, including wildcards, nillable elements, and typed element tests inside `document-node()`.


ghislainfourny:
It looks good, can be merged after the six previous PRs is fixed and merged.

ghislainfourny (approved):
I like how you separated the name from the hierarchy, it's clean.

PR #1819, opened 2026-09-08
Support XML schema node tests

CaiJimmy:
Adds `schema-element()`, `schema-attribute()`, and `document-node(schema-element(...))` support in paths and sequence types. 

Some relevant changes:

## `SchemaElementNodeItemType`

This new class represents `schema-element(name)`, which can match several element names through a schema’s substitution group.

For example:

```xml
<xs:element name="animal" type="xs:string" abstract="true"/>
<xs:element name="dog" type="xs:string" substitutionGroup="t:animal"/>
<xs:element name="cat" type="xs:string" substitutionGroup="t:animal"/>
```

Then `$root/schema-element(t:animal)` can select validated `<t:dog>` and `<t:cat>` elements (not `t:animal` because it's abstract).

## `AttributeNodeItemType`

A new field `schemaTypeAlternatives` has been added to the class instance. This is used to represent multiple acceptable schema type names, when a schema uses a pure union.

For example:

```xml
<xs:simpleType name="Number">
  <xs:union memberTypes="xs:integer xs:float"/>
</xs:simpleType>
<xs:attribute name="code" type="t:Number"/>
```

If we have now an attribute `t:code` which has been validated against a local declaration of type xs:integer, then this check should return true, because `xs:integer` is a member of `t:Number`:

```
$element/@t:code instance of schema-attribute(t:code)
```

Previously,`AttributeNodeItemType` checked only whether the attribute’s type hierarchy contained `t:Number`. An integer’s hierarchy does not contain that union, so it would fail.

## ElementNodeItemType

Similarly, `schemaTypeAlternatives` has been added, for the same purpose.


ghislainfourny (approved):
Awesome!

PR #1840, opened 2026-09-21
Remove dead checkForUnsupportedStatement

CaiJimmy:


CaiJimmy:
Make sure that there are tests that covers the counter-example of this.

ghislainfourny (approved):

PR #1841, opened 2026-09-21
Extract scripting extension visitor logic

CaiJimmy:


src/main/java/org/rumbledb/compiler/TranslationVisitor.java
```diff
@@ -1997,25 +1995,13 @@ public ExitStatement visitExitStatement(JsoniqParser.ExitStatementContext ctx) {
...
-                new ReturnStatementClause(returnStatement, returnStatement.getMetadata());
-        lastFlowrClause.chainWith(returnStatementClause);
-        returnStatementClause = returnStatementClause.detachInitialLetClausesForStatements();
-        return new FlowrStatement(returnStatementClause, createMetadataFromContext(ctx));
+        return LoopStatementTranslation.flworStatement(
+                FlworStatementContext.from(ctx),
+                this.translationContext,
+                this::visitForClause,
+                this::visitLetClause,
+                child -> (Clause) this.visit(child),
```
CaiJimmy: I cannot avoid this casting as there's no `visitClause` method. Maybe we can add a `clause` rule that works as a disjunction of all possible clauses.

ghislainfourny (approved):

